PR #26, opened 2018-08-21
Taks/cli stream check

kali:
Hey @liautaud, if you're curious, this is what it took me to get simo's net to check :) Not in a suitable form for a review, but you may want to have a look anyway. It's been quite a ride :)


liautaud:
Holy cow, it looks like you've been having fun this last couple of weeks! I've just skimmed through the commits for now, I'll read the whole diff in a few days once I'm done with the internship report.

PR #65, opened 2019-02-19
Onnxlstm

kali:


onnx/tests/onnx/mod.rs
```diff
@@ -121,6 +119,8 @@ pub fn run_one<P: AsRef<path::Path>>(root: P, test: &str, optim: bool) {
                 );
             }
             for (ix, (a, b)) in computed.iter().zip(expected.iter()).enumerate() {
+//                println!("computed: {:?}", computed[ix].dump(true));
```
aldanor: Temporary comments?

core/src/analyser/rules/proxies.rs
```diff
@@ -129,18 +80,18 @@ impl Index<usize> for SharedTensorsProxy {
 /// solver.equals(input.rank, 2)
 /// solver.equals(input.shape[1], output.value[0][1])
 /// ```
-pub struct SharedTensorProxy {
+pub struct TensorProxy {
```
aldanor: So, `SharedTensorProxy` is now being replaced with tvec/slice being used directly?

onnx/tests/onnx/mod.rs
kali: Well this is test code, they come in handy when something does not work. I know it looks a bit dirty, but I haven't find a much better way.

core/src/analyser/rules/proxies.rs
kali: There are two things here. 
First a bad name, coming from an over-enthusiast regexp. SharedTensorProxy should be TensorProxy.
Second, the SharedTensorsProxy is gone (with an S at tensors). The rationale being that the input and output arity is always known before we reach the inference stage, so we can work with a regular slice there. It simplifies a bit processing operators with big lists of arguments (like the classifier or the lstm, so i though it was time to do this).

kali:
And thanks for having a look :)

kali:
@aldanor I've split the TensorProxy changes in a different branch to be merged ASAP as it should save you work implementing the rules.

aldanor:
Great, thanks! I would definitely prefer to rebase my branch early and in batches if possible :) (also it should make this PR much more readable after you merge that branch and rebase this one)

kali:
Fully agree. I worked alone for a long time on this project, so I've been a bit sloppy, please bear with me.

aldanor:
No worries :) (I have one such “project in Rust with huge scope, single developer” kind of project on top of me as well, so it’s fully understandable...)

kali:
This branch was merged on the deepspeech branch.

Issue #64, opened 2019-02-18
Support TreeEnsembleClassifier op


aldanor:
(I'm aware that it overlaps somewhat with #56, but it's a bit more specific, hence opening it as a separate issue)

Given that it's now officially possible to [convert LightGBM](https://github.com/onnx/onnxmltools/tree/master/onnxmltools/convert/lightgbm) ([and xgboost](https://github.com/onnx/onnxmltools/tree/master/onnxmltools/convert/xgboost)) tree ensemble classifiers into ONNX, how realistic would it be to expect tract to support [`TreeEnsembleClassifier`](https://github.com/onnx/onnx/blob/master/docs/Operators-ml.md#ai.onnx.ml.TreeEnsembleClassifier) op in the foreseeable future? This would potentially be a huge feature, instantly unlocking whole universe of tree ensemble classifiers (and potentially regressors as well).

// I'd be glad to help if there was some guidance on what to do and where, if needed; not quite sure how much of work it is to implement this since I'm not very familiar with the internals of tract.

Thanks!

kali:
Hi,

I had not had a look at the -ML operators in ONNX yet. I just skimmed over the TreeEnsembleClassifier, and I don't foresee any structural problems.

I will not implement any of the -ML short term. My focus for now is on some optimization bits and recurring networks: basically, completing ONNX core. But I'd be glad to guide you through it.

From 30 000 feet, you will first need to find some kind of test suite: as far as I can tell, ONNX does not provide any for its -ML operators. Then figure out a way to implement the classifier (as in algorithms and structures).

Implementing an operator in tract is relatively straight forward, and we now have dozens of examples, but it's not documented.

Most implementations are in "core/src/ops" while onnx and tensorflow convert their op definitions to core, or implement them on their own if I decided they were too idiosyncratic to go in core. To be honest, I am not too sure what to do with -ML. We could put them in onnx directly, or have a tract-onnx-ml new crate to hold them and thus make them optional. That would definitely be my favourite option if implementation requires to pull in more dependencies. But for starters, I guess we can just shove them somewhere like onnx/src/mlops.

You can have a look at core/src/ops/nn/lrn.rs for an instance of a straightforward operator. As far as I can tell, the classifier is stateless, so its general structure should look like this one. The weirdest part is the rules part, which must produce as much information about input and outputs as it can for the framework to figure out tensors types and shapes. You can safely start with no rules, don't pull your hair on that right away, just return Ok(()).

[...]

PR #70, opened 2019-02-24
Tree ensemble ONNX ML ops [WIP]

aldanor:
@kali Opening this draft PR so as to collect some potential early feedback.

The core generic tree ensemble engine seems to be working (what I've managed to test so far, I was actually super surprised that the tests vs lightgbm passed from the first try after I got it to compile, lol), now need to pin it to protobuf config, set up the rules, run a few basic benchmarks, wrap it in classifier/regressor types, etc. All features (including score post transforms except probit, different various comparison ops, inputnan handling etc) that can be provided in ONNX protobuf config are already supported here.

It's definitely not implemented in the most efficient way now, but I think it shouldn't be too bad (although before there's benchmarks, I have no idea how much worse it would be than the existing lightgbm/xgboost c++ engines).


kali:
#64

onnx/src/ops/ml/tree.rs
```diff
@@ -0,0 +1,664 @@
...
+                scores.map_inplace(|a| {
+                    *a = (*a - max).exp();
+                    norm += *a;
+                });
+                scores.map_inplace(|a| *a = *a / norm);
+            }
+            PostTransform::Logistic => {
+                scores.map_inplace(|a| {
+                    let v = 1. / (1. + f32::exp(-f32::abs(*a)));
+                    *a = if a.is_sign_negative() { 1. - v } else { v };
```
kali: I have a faster implementation somewhere in the guts of the Sigmoid operator (as a rational function, accurate to f32). I think I will move it to linalg (which I will rename tract-compute), so that we can use it anywhere it makes sense.

onnx/src/ops/ml/tree.rs
```diff
@@ -0,0 +1,664 @@
...
+                let max: f32 = scores
+                    .iter()
+                    .cloned()
+                    .filter(|&a| a != 0.)
+                    .max_by(|a, b| a.partial_cmp(&b).unwrap_or(Ordering::Equal))
+                    .unwrap_or(0.0);
+                let mut norm = 0.;
+                scores.map_inplace(|a| {
+                    if *a != 0. {
+                        *a = (*a - max).exp();
```
kali: :+1:

onnx/src/ops/ml/tree.rs
kali: It's probably less critical here than in neural networks, so no rush.

onnx/src/ops/ml/tree.rs
```diff
@@ -0,0 +1,664 @@
...
+            }
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct Tree {
+    n_classes: usize,
+    nodes: Vec<Node>, // TODO: can this be a slice/view into ensemble's contiguous storage?
+    leaves: Vec<Leaf>, // TODO: store as a collection of direct slices instead of indices?
```
kali: I think indices in vectors are a good compromise. Making trees work with slices make the code very difficult, if even possible, as it leads to self borrowing structures.

 If we have to worry about performance here, we can use unchecked vector index (it's unsafe, but we can live with it as the structure is fixed).

onnx/src/ops/ml/tree.rs
```diff
@@ -0,0 +1,664 @@
...
+    ensure!(cond, "Invalid {} #{}: {} = {} {} {} = {} ({:?})", t, index, an, a, cmp, bn, b, obj);
+    Ok(())
+}
+
+impl Tree {
+    pub fn build(n_classes: usize, nodes: &[Node], leaves: &[Leaf]) -> TractResult<Self> {
+        use self::Cmp::{Equal, Less, LessEqual};
+
+        let n_nodes = nodes.len();
+        ensure!(n_nodes != 0, "Invalid tree: no nodes");
```
kali: nice

onnx/src/ops/ml/tree.rs
```diff
@@ -0,0 +1,664 @@
...
+        self.branches().map(|b| b.feature_id).max().unwrap_or(0)
+    }
+
+    unsafe fn get_leaves_unchecked<T>(&self, input: &ArrayView1<T>) -> &[Leaf]
+    where
+        T: AsPrimitive<f32>,
+    {
+        let mut node_id = 0;
+        loop {
+            let node = self.nodes.get_unchecked(node_id);
```
kali: ho, and we're already there. perfect :)

rustfmt.toml
```diff
@@ -0,0 +1,7 @@
+use_small_heuristics = "Max"
+use_field_init_shorthand = true
+use_try_shorthand = true
+empty_item_single_line = true
+edition = "2018"
+unstable_features = true
+fn_args_density = "Compressed"
```
kali: do you have a rationale somewhere for moving away from the defaults ?

kali:
It looks quite good so far. Seems to me that the remaining is mostly plumbing it in tract, and I can help you there.

onnx/src/ops/ml/tree.rs
aldanor: Oh, I didn't mean self-borrowing. More like - the ensemble owns a contiguous collection of nodes / leaves, and trees just receive slices into that. However then it wouldn't be possible to just simply create a new tree without first creating an ensemble and then indexing into it, also there would be lifetimes all over the place... things would get a bit weird, so not sure it would make things a lot better.

For a generic solution, I think this is ok. I had ideas to add specializations later for near-balanced binary trees of the most common depths (e.g. up to 16; I've rarely seen more), which would be lightning fast barely with any lookups etc.. we'll handle it later, that's the fun part ;)

rustfmt.toml
aldanor: If I remember correctly, mostly because of `fn_args_density`; without it, rustfmt will often format this:

```rust
func(x1, x2, x3, x4, x5, x6, x7, x8, x9)
```

as 

```rust
func(
    x1,
    x2,
    x3,
    x4,
    x5,
    x6,
    x7,
    x8,
    x9,
)
```

even though its still within line length limit.

aldanor:
I already have some sketches for parsing from protobuf config, I think that shouldn't be too complicated.

I may need some help with inference rules later, in which case I'll let you know :)

onnx/src/ops/ml/tree.rs
aldanor: Makes sense. I flipped around the codebase but didn't find it - I guess I've missed it somehow.

Maybe there should be a toolbox of common operations, some of them simple scalar functions, some of them dealing with vectors (low-level, without any tract glue, without any validation, without bounds-checking etc), implemented as fast as possible so they can be reused in different parts of the library - like softmax, logistic, etc.

kali: Yeah, that was the idea of tract-linalg. One of the key features is its ability to compile fast, in order to iterate quickly over tests and benches of implementation. But I got the name wrong. This goes beyond linear algebra.

rustfmt.toml
kali: all right, I'll give it a try on the rest of the codebase and see if i like it :)

aldanor:
Ok, parsing config is done now, I think (haven't checked yet, but it compiles at least).

Now eval/inference/tests + regressor ensemble, hopefully most boilerplate is done already.

aldanor:
(Parsing the tree ensemble from `NodeProto` turned out to be quiiiiite a tedious adventure...)

onnx/src/ops/ml/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+                Ok(())
+            },
+        )?;
+
+        s.given(&inputs[0].datum_type, move |_, dt| {
+            Ok(match dt {
+                DatumType::F32 | DatumType::F64 | DatumType::I64 | DatumType::I32 => (),
+                _ => bail!("invalid input type for tree ensemble classifier: {:?}", dt),
+            })
+        })?;
```
kali: This is actually better than most rules() I wrote. I think we need a "one_of" rule and a "castable_to" rule for datum types.

onnx/src/ops/ml/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+                Node::new_leaf(start, end)
+            } else {
+                node.bail(&format!("leaf not found: tree_id={}, node_id={}", tree_id, node_id))?
+            }
+        };
+        nodes[tree_id].push(tree_node);
+
+        prev_tree_id = tree_id;
+        prev_node_id = node_id;
+    }
```
kali: about this part, which is indeed a bit tedious, have you consider using the zip macro from itertools to walk all the vectors in sync once your done with the sanity checks ?

onnx/src/ops/ml/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+        s.equals(&inputs[0].rank, 2)?;
+        s.equals(&outputs[0].rank, 1)?;
+        s.equals(&outputs[0].rank, 1)?;
+
+        s.given_3(
+            &inputs[0].shape,
+            &outputs[0].shape,
+            &outputs[1].shape,
+            move |s, s_in0, s_out0, s_out1| {
+                s.equals(s_in0[0], s_out0[0])?;
```
kali: unless I'm missing something this two shape constraint don't need to be in the given block. The benefit of being outside of the given block is that it will allow to make partial deduction on shape individual values, while the given() blocks are only executed if and when all the arguments are determined.

```
s.equals(&inputs[0].shape[0], &outputs[0].shape[0])?;
```

should do the trick, right after the rank checks (order does not matters).

onnx/src/ops/ml/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+                s.equals(s_out0[0], s_out1[0])?;
+                if let Ok(k) = s_in0[1].to_integer() {
+                    self.ensemble.check_n_features(k as _)?;
+                }
```
kali: These two checks are range checks, so going for "given" is the right option. In an ideal world, we would have rules in the form of "a gt b" and "a lte b", but implementing reasoning on range feels overkill.

onnx/src/ops/ml/mod.rs
aldanor: Yea, I was looking for `one_of()` rule and couldn't find it. Would be sure nice to have it in the solver.

As for `castable_to()`, it probably would be less useful if we want to stick strictly to onnx spec, since sometimes it's more like "what onnx thinks it's castable to" than "what it really is castable to".

onnx/src/ops/ml/mod.rs
aldanor: I think this part is actually overly tedious, the worst one is constructing the leaf map above... and in all cases I kind of need to know the node_id/tree_id on the previous step (well, sort of... it could be done in a different way, but it would be much harder to read). I think with itertools it would end up much harder to read anyway if you were to zip it between each other and also with previous steps.

onnx/src/ops/ml/mod.rs
aldanor: Makes sense, thanks! I'll change it to simple equalities. I looked at the existing code but wasn't sure when exactly you're supposed to use given().

onnx/src/ops/ml/mod.rs
aldanor: Ha, I was almost about to ask, why not have `s.lt()` and `s.le()` (maybe also `s.not()` then) :)

onnx/src/ops/ml/mod.rs
kali: Well, on top of that, there are cases where i need to diverge from the onnx spec in terms in typing: in the case of voice inputs, we often need to reason with a "time" dimension which is either a variable of the input, or an actual (infinite) stream. You may have noticed the TDim stuff... So in these situation, I need onnx (or tf) default integer types (i64 for onnx, i32 for tf) to be upcasted to a TDim that can compute expression like "S/2+4" and the like...

aldanor:
@kali Another question I had, re: dimensionality, onnx spec says tree ensembles always take 2D input and produce 1D output (in case of classifier, two outputs, 1D and 2D). Can we “broadcast” it down as well as an extension? Otherwise it’s pretty cumbersome to make single observation predictions. Basically, if you pass a 1D input, you get back 0D (or 0D + 1D for classifier).

kali:
I have no strong objection... What is the supplementary dimension ? Some kind of batch thing as in CNNs ? So far I have religiously respected the useless N dimension in images and voice tensors, but maybe it's time to depart from that.

aldanor:
Yea, it's "`N`", so that you can e.g. compute it on a million rows at once and get all of your outputs. I guess this way you only do all your inference and checks and input validation all once, so the relative cost of it becomes almost zero. But having only the "batch" version seems a bit weird, since most of the time this would be used for single observations (for batch versions, people just use Python and do their learning there).

kali:
Yeah, ok, so same thing. I think the situation is a bit different with the convolutions because they are hidden in the guts of wider networks, where as the tree classifier could be used as a single op network. So ok, go for it, but I will probably not do it on the convos just for the sake of consistency :)

kali:
Hey! You've been silent for a while :) You looked so close to the finishing line... Anything I can do to help ?

aldanor:
@kali Been a bit busy last week, was planning to return to this PR today or tomorrow - don't worry, I'm not abandoning this, because I'll need it for my own work :)

Looks like I'll have to rebase this on the recent op-register stuff in the master, but it looks like it would be reasonably easy.

One minor question: in the inference rules, what's the best way to write (I'm still slightly confused by given() and what is lazy and what is not):

- Either input0 is 2-dim, in which case output0 and output1 are 1-dim; in this case shape[0] for all three must match
- Or, input0 is 1-dim, in which case output0 and output1 are 0-dim

aldanor:
(Updated the comment above)

kali:
Good ! Glad to hear you have an actual paying use case. I'm actually curious about any bit of information you may be able to share about it, publicly or not.

Yeah, the op-register rebase should not easy to deal with. It  allows us to consider a separate tract-onnx-ml crate, with a dependency on tract-onnx. It would just have to implements a onnx_ml() method: that one would call `tract-onnx::onnx()` and register the ml operators on top of the onnx one. 

To be clear, I'm just explaining the purpose of the op-register/framework refactoring, *not* asking you to do this. We will do it at a later stage when we have a bit more -ml ops, or if I get some back-pressure on the binary size (snips doing embedded after all).

I'm thinking about another round of refactoring following that one, to have a cleaner separation of "logic ops" and "physical ops". Only the logic, high level ops would have to implement inference rules, not the dumb and strongly typed ops I'm starting to need in many places to improve performance. The idea being that once the logic ops have done their work, normalized the network and inferred all types and shapes, I want the framework to be responsible for keeping track of these types and shapes to simplify writing physical ops as much as possible. I'm getting bored of implementing rules() method that I know will never be called. In any case, I'll make sure the rebasing is ok. As you noticed, I keep having several open fires all the time, so I'm painfully aware when I do stuff that is hard to rebase.

For your question about inference rules: only the check_*_arity are eager. All the solver stuff is lazy. rules are appended in the solver by the op rules() code, then the framework will try to make sense of them and infer as many facts as it can.

The given_ closures are also lazy: during the actual inference process (when the framework runs the op solver, after the rules() method have returned), if the top-level rules are strong enough for the solver to determine the first argument of the given rule, then the closure is executed with the actual value in order to add more rules to the mix.

In your case, I think you could achieve it with something like:

```rust
s.given(inputs[0].rank, |s, input_rank| {
 if input_rank == 2 {
 ...
 } else {
 ...
 }
})?;
```

Or be smarted to avoid the `if` inside the `given` by using input_rank to pick the right axis when writing rules. (If you feel like doing a function to deal with two `if` branches, checks for examples, as the lifetimes are not exactly trivial).

But I'm pretty sure the given(inputs[0].rank, ...) will be there.

aldanor:
Thanks for explanation! Will get back to you when I rebase it and rework the inference rules. The eval itself should be fairly simple because the current evaluator already supports both 1-D and 2-D cases.

(Btw I also noticed that LightGBM exports ZipMap nodes in the onnx output graph, it's a part of onnx.ml but is super-trivial, so I'll add that one as well)

kali:
Hey, just a quick notice, I've been lead to merge a lot of things faster than I was originally planning. I think the rebase is still kind of ok. If it's not, just tell me and and I'll do it for you.

kali:
Hey, are you still interested in pursuing this ? Do you need help ? Should I take over ?

aldanor:
@kali Hey - my apologies, been busier than I expected lately, and it requires a contiguous chunk of time to concentrate on this and do some actual work 😢

Thanks for rebasing the branch on the latest master!

If you could help out with the inference rules, that would be fantastic, since you'll do it much faster anyway - I think we've covered most on how it should be done in the comments above? (re: 0d/1d/2d cases etc)

Btw Microsoft guys apparently fixed issues in onnx-runtime that I noticed when I was writing this (https://github.com/microsoft/onnxruntime/pull/1022) - so now we could steal some tests from them as well.

(I'll keep track of this PR in the meanwhile in case you commit anything and will comment from  the phone)

kali:
Hi, happy to know you're still onboard.

For context, onnx-ml came a bit back in my radar because i'm looking at the BiDaf network which exercise a lot of onnx section that are not greatly covered by tests (bidi lstm, scan op, and more). It also uses the onnx-ml CategoryMapper op, so I implemented it. 

I think I have written the right inference rules but I really have no way to try and test them. It would be great if we could use tests from onnxruntime indeed...

onnx/src/ops/ml/mod.rs
```diff
@@ -288,26 +288,24 @@ impl InferenceRulesOp for TreeEnsembleClassifier {
...
+
+        s.given(&inputs[0].rank, move |s, rank| {
+            if rank == 1 && rank != 2 {
+                bail!("First input rank must be 1 or 2");
+            }
+            if rank == 2 {
+                s.equals(&inputs[0].shape[0], &outputs[0].shape[0])?;
+                s.equals(&inputs[0].shape[0], &outputs[1].shape[0])?;
+            }
+            // FIXME: do we somehow know Nfeatures ?
```
aldanor: There's `TreeEnsemble.check_n_features()`?

There's no exact equals bound, but there's a >= bound which can be inferred from the featureids used in all trees.

bminixhofer:
Hi, I was just discussing running tree-based models (exported from LightGBM or XGBoost) in the browser. It looks like this PR is the closest thing we have to that. Is there any interest in taking this up again @aldanor @kali (as it seems from the commit history that it's already working to some degree? ) ?

Happy to help with test cases again if needed.

kali:
IIRC, it was very close to working. The tract side of thing may have aged pretty badly, but nothing I can not fix in a few minutes.. Also, I now have a way to make this an actual optional module instead of bloating the onnx components. So yeah, if you feel like providing some test cases, I'm happy to fix the plumbing and see where we stand.

bminixhofer:
Awesome! I'll add a couple of test cases for models exported from LightGBM / XGBoost / sklearn then.

kali:
Closing this PR, even if tree regressors are not implemented for now.

PR #72, opened 2019-02-27
Task/various fixes

alaa-saade:
- Allow concat and reduce operations to work with negative axis
- Fix gathering on scalar index (i.e. regular indexing of a tensor)
- Add i64 to TDim conversion, and relax condition on type of input for constant of shape
- Allow unspecified dimension in onnx model
- Typos and logger bug fixes


core/src/ops/array/concat.rs
```diff
@@ -4,15 +4,39 @@ use crate::ops::prelude::*;
...
 impl Concat {
+    fn resolve_axis(&self, rank: i64) -> TractResult<usize> {
+        let axis = {
+            let axis_res: TractResult<i64> = {
+                if 0 <= self.axis && self.axis <= rank - 1 {
+                    Ok(self.axis)
+                } else if -rank <= self.axis && self.axis < 0 {
+                    Ok(self.axis + rank)
+                } else {
+                    bail!("Illegal combination of values for rank and axis")
```
kali: Can we give the values in the error message (I would actually have been happy for a simpler check impl, but...)

core/src/ops/nn/reduce.rs
```diff
@@ -206,18 +207,45 @@ where
...
+                } else if -rank <= axis && axis < 0 {
+                    Ok(axis + rank)
+                } else {
+                    bail!("Illegal combination of values for rank and axis")
+                }
+            };
+            axis_res? as usize
+        };
+        Ok(axis)
+    }
```
kali: A bit convoluted, no ?

```
fn resolve_axis(axis: i64, rank: i64) -> TractResult<usize> {
                if 0 <= axis && axis <= rank - 1 {
                    Ok(axis as usize)
                } else if -rank <= axis && axis < 0 {
                    Ok((axis + rank) as usize)
                } else {
                    bail!("Illegal combination of values for rank and axis: {} and {}", rank, axis)
                }
    }
```

core/src/ops/array/concat.rs
kali: See the one in reduce...

onnx/src/ops/nn/mod.rs
```diff
@@ -10,7 +10,7 @@ macro_rules! reduce {
         |node| {
             let axes = node
                 .get_attr_opt_ints("axes")?
-                .map(|axes| axes.iter().map(|&i| i as usize).collect());
+                .map(|axes| axes.iter().map(|&i| i as i64).collect());
```
aldanor: Note: see #75

onnx/src/tensor.rs
```diff
@@ -45,12 +45,18 @@ impl Tractify<TypeProto_Tensor> for TensorFact {
         }
         if t.has_shape() {
             let shape = t.get_shape();
-            let shape: Vec<usize> = shape
+            let shape: TVec<DimFact> = shape
                 .get_dim()
                 .iter()
-                .map(|d| d.get_dim_value() as usize)
```
aldanor: Ref #75

PR #75, opened 2019-03-01
Refactor NodeProto ONNX pb helpers

aldanor:
As I've started writing pb config parser for tree-ml nodes, I figured it might be helpful to refactor protobuf getters a bit.

This brings an easier way to parse all kinds of ints, vecs/tvecs etc, with less boilerplate.

It's not just a refactor, a few functional changes as well, stricter validation:
- Trying to convert an int64 pb value into an integer type that can't contain it would now raise a descriptive error (especially with all the `usize` values); this also works with vecs/tvecs. In the current version, it would happily convert an `-1i64 as usize`; here, it would complain loudly. Ideally, all occurences of `as usize` in the codebase should also be double-checked.
- Booleans are now restricted to 0 and 1 (this was also inconsistent, in some places it was checking for 1, in some places - for non-zero, etc).

It's still a bit of a mess, with a whole bunch of random stuff in pb_helpers, but it's all internal so should be all good for now.

Upd: ouch, rebasing this on #72 and #73 would be a bit painful... :)

### DIFF
--- a/core/src/ops/array/pad.rs
+++ b/core/src/ops/array/pad.rs
@@ -1,6 +1,7 @@
 use crate::ops::prelude::*;
 use ndarray::*;
 use num_traits::AsPrimitive;
+use std::str::FromStr;
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum PadMode {
@@ -8,6 +9,19 @@ pub enum PadMode {
     Reflect,
     Edge,
 }
+
+impl FromStr for PadMode {
+    type Err = ();
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "reflect" => Ok(PadMode::Reflect),
+            "edge" => Ok(PadMode::Edge),
+            _ => Err(())
+        }
+    }
+}
+
 impl Default for PadMode {
     fn default() -> PadMode {
         PadMode::Constant(0.0)
--- a/core/src/ops/array/pad.rs
+++ b/core/src/ops/array/pad.rs
@@ -1,25 +1,12 @@
 use crate::ops::prelude::*;
 use ndarray::*;
 use num_traits::AsPrimitive;
-use std::str::FromStr;
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum PadMode {
     Constant(f32),
     Reflect,
     Edge,
-}
-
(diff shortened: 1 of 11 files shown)


core/src/ops/array/pad.rs
```diff
@@ -1,13 +1,27 @@
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "reflect" => Ok(PadMode::Reflect),
+            "edge" => Ok(PadMode::Edge),
+            _ => Err(())
+        }
+    }
+}
+
```
kali: I do not think the parser has its place in core, these strings are onnx specific.

core/src/ops/nn/padding.rs
```diff
@@ -8,6 +10,20 @@ pub enum PaddingSpec {
...
+        match s {
+            "NOTSET" => Ok(PaddingSpec::Valid),
+            "VALID" => Ok(PaddingSpec::Valid),
+            "SAME_UPPER" => Ok(PaddingSpec::SameUpper),
+            "SAME_LOWER" => Ok(PaddingSpec::SameLower),
+            _ => Err(()),
+        }
+    }
+}
+
```
kali: Same comment here (TF uses "SAME" and "VALID")

kali:
Thanks for attempting this. I like it a lot, I will probably it in TF side when I found the time. My only reservation is about the padding modes strings that I would prefer to keep out of core as they are not consistent between the two frameworks.

And yeah, the rebase will be a bit painful, but that's the price to pay for my "ho, I want to try that *now*" attitude.

kali:
Green light from the bench room, no perf impact (no big surprise here) and a tiny reduction in size.

aldanor:
@kali As for `FromStr` - totally agree, I haven't thought about tf/onnx. I will remove those both impls tonight and fix that (in the evening when I get back home).

Now the question is then, who rebases what on what :)

kali:
I've discussed it with @alaa-saade, we will merge yours first. The sooner the better, tonight is fine.

aldanor:
@kali I've removed FromStr impls and reverted parsing from strings for PadMode / PaddingSpec to be ONNX-specific.

kali:
Ha, great ! Thanks again

PR #79, opened 2019-03-08
isolate op-register

kali:


kali:
cc @aldanor, on the way to more flexible registration so that we can split up onnx-ml (or even onnx-nn-conv, onnx-nn-rec, onnx-rec-flow, who know ?)

aldanor:
So hmm there’s now multiple op registers depending on what kind of node you feed to them, ie the node definition type. What happens to shared registers? (Common stuff shared between all) Sorry, I’m a bit lost :)

kali:
Don't worry, I just wanted to mention that I started that, hopefully it will become clearer as I'm moving it forward :)

PR #86, opened 2019-04-03
Api cleanup

kali:


kali:
ping @alaa-saade, found the flush list bug at least.

PR #144, opened 2019-08-19
declutter lstm-nonlin

kali:


kali:
will be de facto merged as part of fused-conv branch.

PR #174, opened 2019-09-30
Disable onnx tests on Windows, until they can be fixed.

hrydgard:
Allows you to do a plain cargo build in the base directory.


kali:
Is this the only sub project in harness that blocks the build under windows ? The other are ok ?

hrydgard:
It appears so.

kali:
All right thanks, I will change the CI scripts accordingly.

PR #192, opened 2019-11-23
Checking os == ios instead of family == ios

pruthvikar:
Came across this bug trying to compile for iOS.

I wasn't able to link to _arm64simd_smmm_8x8 since ios is not a valid target_family. Changed to check against target_os instead. (assuming this is the intended check and not family == unix)
https://doc.rust-lang.org/reference/conditional-compilation.html#target_os


kali:
Thanks !

kali:
cut a 0.5.8 with your fix.

pruthvikar:
Great, thanks!

PR #200, opened 2019-12-06
add i8 variants in matmul codegen

kali:


core/src/ops/matmul/logic.rs
```diff
@@ -20,6 +20,16 @@ fn eval(
             return eval_t(a, b, a_trans, b_trans, c_trans, q_params, &|m, k, n| {
                 MMMWrapper::Quant((tract_linalg::ops().qmmm_i8_i32)(m, k, n))
             });
+        } else if (a.datum_type(), b.datum_type()) == (i8::datum_type(), i8::datum_type()) {
```
dkim, line 23: Something is wrong. This condition is the same as the previous condition at line 19.

kali: Thanks for spotting that.

PR #231, opened 2020-02-09
Added a short example of training a model in Python and then serving with Rust.  

mattalhonte:
Added a short example of training a model in Python and then serving with Rust.


examples/test_tract/src/main.rs
```diff
@@ -0,0 +1,27 @@
+use tract_core::ndarray;
+use tract_core::prelude::*;
+use rand::*;
+
+fn main() -> TractResult<()> {
+    // load the model
+    let model = tract_tensorflow::tensorflow().model_for_path("./my_model.pb")?;
+    println!("Loaded model");
+
```
kali: I'd prefer the example to optimise the model here (or after the input in created). The sequence should look like the one in https://github.com/snipsco/tract/tree/master/examples/tensorflow-mobilenet-v2#a-look-at-the-code, with the `set_input_fact` and into_optimized() call.

I just realized the tract-tensorflow crate example is not like that and will fix it promptly.

mattalhonte:
Cool, I'll work on all these!  tract (lowercase!) is rad, by the way.  This is my first Rust project that wasn't out of a book.

mattalhonte:
Heya.  So I tried to add an Optimization statement and it's not quite working.  Based off the snippet before, I put 

`model.set_input_fact(0, InferenceFact::dt_shape(f32::datum_type(), tvec!(1, 100)))?;`

Which should be right for processing 1 row with 100 columns?

And it gave me this:
```error[E0596]: cannot borrow `model` as mutable, as it is not declared as mutable
  --> src/main.rs:10:5
   |
7  |     let model = tract_tensorflow::tensorflow().model_for_path("./my_model.pb")?;
   |         ----- help: consider changing this to be mutable: `mut model`
...
10 |     model.set_input_fact(0, InferenceFact::dt_shape(f32::datum_type(), tvec!(1, 100)))?;
   |     ^^^^^ cannot borrow as mutable

error: aborting due to previous error```

**Comment by kali** [comment]
Ha, tract can only optimise shape-determined nets.

There is a trick ("look for Pulse in the Readme") to convert streaming networks (where one and only one dimension is streaming, with non-determined length) to a pulsing network that will operate on chunks of fixed size and produce output in streaming fashion.

But I think this goes beyond the scope of a "simple" example : the pulsing API is a bit harder to use, and the transformation is still a bit fragile so I consider that part experimental.

 Maybe we can start with this example demonstrating how to go from python to a regular tract deterministic graph (and thus choosing an arbitrary value for that dimension). Then we could consider pulse-transforming the network to demo the pulsing ? WDYT ?

**Comment by mattalhonte** [comment]
Oh, the docs said you could add an unknown dimension with `_`?  It has the example `shapefact![1, 2, _]`.  Couldn't quite figure out how to actually call `shapefact!` though.  Still learning how to navigate through the structs and methods and stuff of Rust crates.

**Comment by kali** [comment]
Mmm yeah, indeed, you can use shapefact! to define a partial shape. And that is very useful internally, but not so much for using the prediction. Which is usually fine.

As I said, tract can only really optimise network with known finite dimensions. This works out of the box for image-like input (that you typically get as one single buffer). And for time-driven data (like voice or sound processing), one can use tract and the pulse transformation to rewrite a network with one "magic" dimension (called "S" for Streaming) to a network operating on chunks.

So, what kind of data does your input represent, and what would the typical workflow be ?

**Comment by mattalhonte** [comment]
I was thinking of just having it represent Panel data, as a little "Hello World".  Though maybe a fixed-width Time Series would be better and play more nicely with tract?

**Comment by kali** [comment]
Both could be OK, I think, depending on what we want to demo as a prediction: the two possible scenario are
* input for prediction is of predetermined sized on all dimensions (for instance for panel data, prediction apply for an input of 1 person, 5 years, 12 questions, giving us a 5x12 shape input tensor). This is the simple case, probably the one we should target for a "simple" example that already demo how to get from keras to tract.
* input for prediction has one (and only one) undetermined dimension, or possibly inifinite. time series or real time voice work like that. This scenario is a bit more complicated in tract, requires manipulating the magic S shape and stateful APIs. For instance, keyword spotting (as in "alexa, blablabla") is the archetypal example, where you would feed repeatedly chunks of audio input and get a chunks of detection confidence. It could also help for time series of variable length, without the streaming/real-time constrait: predict the next day temperature from a day-to-day temperature capture of variable length (because weather stations where not established at consistent dates for instance).

**Comment by mattalhonte** [comment]
K, I made the changes, and cleaned some other stuff up.

File: examples/jupyter-keras-tract/src/main.rs [hunk 2]
```diff
@@ -0,0 +1,29 @@
...
+
+fn main() -> TractResult<()> {
+    // load the model
+    let mut model = tract_tensorflow::tensorflow().model_for_path("./my_model.pb")?;
+
+    // specify input type and shape
+    //model.set_input_fact(0, InferenceFact::dt_shape(f32::datum_type(), tvec!(_, 100)))?;
+
+    model.set_input_fact(0, InferenceFact::dt_shape(f32::datum_type(), tvec![10, 100]))?;
+    // Generate some input data for the model
```
kali: Well, one final thingy here, and that was the entire point: now that you have specified the input shape, you can call into_optimized() and get an even shorter running time :)

mattalhonte:
Ha, right!  Added the optimization call, and a short comparison in the readme.  Thanks so much for walking me through all this!

kali:
@mattalhonte thanks a lot for making this ! At your disposal if you want to play a bit more and venture into the streaming part of the API :)

PR #248, opened 2020-03-23
Upgrade liquid to 0.20

hrydgard:
The old version depended on an old syn version that we don't want in our deps.

By the way, was wondering why in the world tract would use a templating engine - a macro assembler was an unexpected but clever usecase!


kali:
Haha, thanks :) liquid felt a bit overkill at the beginning, but I'm glad I did it.

PR #250, opened 2020-03-24
Move env_logger to [dev-dependencies] wherever possible.

hrydgard:
And a tiny warning fix.

Prevents the env_logger dev dependency from leaking out to crates depending on tract.

### DIFF
--- a/harness/tf-moz-deepspeech/src/lib.rs
+++ b/harness/tf-moz-deepspeech/src/lib.rs
@@ -10,7 +10,6 @@ use std::str::FromStr;
 
 use tract_tensorflow::prelude::*;
 
-#[allow(dead_code)]
 #[cfg(test)]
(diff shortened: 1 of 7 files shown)


harness/tf-moz-deepspeech/src/lib.rs
```diff
@@ -11,6 +11,7 @@ use std::str::FromStr;
 use tract_tensorflow::prelude::*;
 
 #[allow(dead_code)]
```
linkmauve: You can probably remove that line, this function wouldn’t be dead if it was only compiled for tests.

hrydgard: Thanks, done.

kali:
merged on #249

kali:
0.6.1 !

PR #276, opened 2020-05-26
cast ints from/to bool

kali:


core/src/datum.rs
```diff
@@ -320,6 +320,27 @@ impl TryInto<f64> for bool {
     }
 }
 
+macro_rules! cast_to_and_from_bool {
+    ($t: ty) => {
+        impl TryInto<bool> for $t {
+            fn try_into(&self) -> TractResult<bool> {
+                Ok(*self == 0)
```
bio43, line 327: The int -> bool cast is not well defined, but according to https://github.com/onnx/onnx/issues/2285 , it seems that truncating the leading bits is the right approach. Suggestion:
```suggestion
                Ok((*self & 1) == 1)
```

kali: wow. I was more tired than I thought.

kali:
I'll stick to != 0 though. That's what MS onnxruntime does, and it it feels more natural to me than truncation. Probably matter of taste though.

bio43:
@kali ok, I tested this branch and can confirm it works for my use case.

PR #283, opened 2020-05-28
Add PyTorch example

teddykoker:
Excellent work with Tract! I added an example using PyTorch/ONNX that might make it easier for some to get started.


kali:
Wow, thanks ! That's great :)

examples/pytorch-resnet/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "pytorch-resnet"
```
kali, line 2: Can we use example- as a prefix here to be consistent with the other test ?

teddykoker: Yes!

kali:
all good ? :)

teddykoker:
Yup! I added standardization as well; this might need to be added to other examples.

kali:
Yep, but some networks have it built-in as the first stage... And well, the whole point of NN should be not to have to worry too much about a bit of color correction :)

kali:
Merged. Thanks a lot.

teddykoker:
Great! Thank you for your work on this, it is good to see other low level deployment options.

PR #285, opened 2020-05-31
Add with_ convenience methods, add into_runnable

bminixhofer:
- Added `with_xxx` convenience methods to allow building models in a more fluent style.
- Added `into_runnable` model which returns a `type RunnableModel<F, O, M> = SimplePlan<F, O, M>` to more fluently create runnable models.

I noticed an inconsistency in that `set_output_label` did not return a `TractResult<()>` like other `set_xxx` methods so I went ahead and fixed that.

I also converted the examples to use the new syntax. If you don't want to do that just yet I'm fine with undoing it.

### DIFF
--- a/examples/tensorflow-mobilenet-v2/README.md
+++ b/examples/tensorflow-mobilenet-v2/README.md
@@ -61,32 +61,36 @@ $ cat -n imagenet_slim_labels.txt | grep -C 3 654
 Everything happens in [src/main.rs](src/main.rs).
 
 ```rust
-let model = tract_tensorflow::tensorflow()
-    .model_for_path("mobilenet_v2_1.4_224_frozen.pb")? // load the model
-    .with_input_fact(0, InferenceFact::dt_shape(f32::datum_type(), tvec!(1, 224, 224, 3)))? // specify input type and shape
-    .into_optimized()? // optimize the model
-    .into_runnable()?; // make the model runnable and fix its inputs and outputs
+    let model = tract_tensorflow::tensorflow()
+        // load the model
+        .model_for_path("mobilenet_v2_1.4_224_frozen.pb")?
+        // specify input type and shape
+        .with_input_fact(0, InferenceFact::dt_shape(f32::datum_type(), tvec!(1, 224, 224, 3)))?
+        // optimize the model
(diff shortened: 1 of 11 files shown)


kali:
Well, I really like how the fluent API look like, I'm glad we did this, and thanks for doing it on the examples.

I noticed some reindentation here and there. Are you using rustfmt ? I'm not doing it systematically (basically just run it on a file when I'm editing it) so some files have not been reformatted in a while and rustfmt sometimes change a bit what it does... it could come from my side.

Additionaly, I was wondering if we should not put the comment on the line before instead of to the right in the examples, to prevent them from being hidden to the right (forcing reader to scroll horizontally).

bminixhofer:
I like the new syntax too!

I usually use rustfmt but I turned it off for this PR because I didn't want to change lines which were not subject of the PR. But I'm pretty sure none of the new code gets changed by rusfmt. Where specifically do you see new indentation?

You're right, the line with `with_input_fact` is definitely too long, but would you put the comments on the line before only for `with_input_fact` or for all lines of the builder?

examples/tensorflow-mobilenet-v2/README.md
```diff
@@ -61,36 +61,32 @@ $ cat -n imagenet_slim_labels.txt | grep -C 3 654
...
+let result = model.run(tvec!(image))?;
+
+// find and display the max value with its index
+let best = result[0]
+    .to_array_view::<f32>()?
+    .iter()
+    .cloned()
+    .zip(1..)
+    .max_by(|a, b| a.0.partial_cmp(&b.0).unwrap());
+println!("result: {:?}", best);
```
kali, line 89: Reindententation here and the other doc examples.

kali:
I think I would put them all on the line before for consistency (maybe even followed by a ":" to make it obvious they're describing the line below)...

bminixhofer:
Oh you were referring to the .md file. I changed it back and put all comments on the line before (but without a colon, I thought it looked weird and it's clear as it is IMO).

kali:
Well done :) Thanks !

bminixhofer:
Thanks!

PR #302, opened 2020-06-15
well, normalized model must have been helpful some day. lol.

kali:


kali:
@bminixhofer Thanks for the conversations we had, they pushed me to challenge a few things :)

PR #304, opened 2020-06-15
More model api cleanup

kali:


kali:
@bminixhofer I think tractjs should be unaffected as you're using InferenceModel::into_optimized()

bminixhofer:
Wow this and https://github.com/snipsco/tract/pull/302 are huge wins for the public API! Awesome work! And glad I helped indirectly ;)

PR #325, opened 2020-07-18
Onnx 1.6 1.7

kali:


kali:
discard in favour of #328 (no dot in branch name :/)

PR #339, opened 2020-08-03
Fix cross compiling to Windows with clang-cl

Jake-Shadle:
This fixes an issue where targeting `x86_64-pc-windows-msvc` from a non-windows host would fail due to explicit usage of `lib.exe` and `ml64.exe`. The code now assumes that if that target is used from a non-windows host, the compiler is capable assembling files with the GAS syntax instead.

### DIFF
--- a/linalg/build.rs
+++ b/linalg/build.rs
@@ -1,83 +1,125 @@
-use std::env::var;
-use std::ffi;
-use std::fs;
-use std::path;
+use std::{env, ffi, fs, path};
+
+fn var(k: &str) -> String {
+    env::var(k).unwrap()
+}
+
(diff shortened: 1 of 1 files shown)


kali:
Thanks for the patch, just checking the CI is happy with it. Should we put this configuration (targetting msvc from linux for instance) under CI ?

Jake-Shadle:
We could do that, the question would be if there are any other C/C++ dependencies that are compiled in, since unfortunately it's not convenient to acquire the MSVCRT or windows libraries on a non-windows platform.

kali:
All right. This sounds non-trivial enough. We'll consider we have enough coverage with the windows native build. Merging.

PR #340, opened 2020-08-03
Nnef dumper

kali:


kali:
Hey, @bminixhofer I'm curious to see if you can check the impact of the new nnef based approach (and at least if you manage to get it to work) with tractjs :) I see a ~25% code size reduction on my side.

I have plans for more code reduction, but I think this is an interesting milestone.

See if https://github.com/snipsco/tract/tree/main/examples/nnef-dump-mobilenet-v2 can help.

bminixhofer:
Exciting! I'll try it. You are referring to [this NNEF](https://www.khronos.org/nnef)? Never heard of it before.

Also, adding it to tractjs is not trivial (since it needs the equivalent of rust features, which is *of course* not that easy in the JS ecosystem) but I think I can get it to work via [TF.js style splitting in multiple packages](https://github.com/tensorflow/tfjs#about-this-repo).

kali:
yes that's the one :) consider tract_nnef to be alpha at this stage, i may need to change the api a bit, and the format extension are not 100% stable. but i'm curious to see how smaller a package it will give un with tract.js. It should help with startup time too as most of the decluttering is done beforehand. It does not solve the dynamic sizing issue, but this may be coming next.

kali:
@bminixhofer just so you know, stay in the image categorizer field for now. recurring are coming, probably next week.

kali:
@bminixhofer FYI, "your" model (as in onnx/test_cases/byte_sb_bidi_lstm) now works once dumped and reloaded.

PR #360, opened 2020-09-06
add tests for trees

bminixhofer:
This PR adds tests for tree ensembles (see https://github.com/snipsco/tract/pull/70).

There's four new tests:
- One for a LGBM GBDT regressor, exported via [hummingbird](https://github.com/microsoft/hummingbird). Hummingbird converts the trees to tensor ops, so it seems *really* close to working with tract as it is (just some problem with the OneHot operator I think).
- The same for a LGBM GBDT classifier.
- An XGBoost regressor, exported via [onnxmltools](https://github.com/onnx/onnxmltools). onnxmltools exports it as graph using the `TreeEnsemble*` op from ONNX-ML.
- The same for an XGBoost classifier.

I chose this combination of tree frameworks / exporting frameworks just because it works well. Also no scikit-learn trees in the tests, but I'm pretty sure those will work if the above work, and I think that's enough models to test against for initial support.

I commited the `*.onnx` files directly since they have ~ 10 KB each.


kali:
merged on #363

PR #363, opened 2020-09-12
Aldanor onnx ml

kali:


kali:
@bminixhofer I will need a bit more of help with the tree regressor: the tree classifier seems to be working, there was a relatively obvious entry point, but I have no idea what the tree regressor is supposed to do :)

@aldanor if you still have some interest in this, can you point me in the right direction ?

bminixhofer:
Sure, this is how I'd use the models with Python and onnxruntime:

```python
import onnxruntime

# this is just a sample from the sklearn wine dataset - target is to classify wines based on some features
# i trained the model as regression (although the task is classification)
data = [[12.08, 2.08, 1.7, 17.5, 97.0, 2.23, 2.17, 0.26, 1.4, 3.3, 1.27, 2.96, 710.0]]

# works with both, the `xgboost_regressor_tree` and `lgbm_regressor_tensor`
sess = onnxruntime.InferenceSession("model.onnx")
outputs = sess.run(None, {"input": data})[0]

# should print "[[0.97973335]]" if you're using the XGBoost model or "[[0.9594169]]" if you're using LGBM
print(outputs)
```

aldanor:
@kali hey, been out for a while :) I'll try to check the status of this on the weekend and reply then

aldanor:
@kali As I'm re-reading my own code and the recent commits, would you please comment on have recently updated, and what is left to be done? (I will figure it out on my own, but it would speed it up if you helped).

As for TreeRegressor, I remember I was trying to use Microsoft's runtime as reference, but they had bugs in it which I reported and they then [fixed](https://github.com/microsoft/onnxruntime/pull/1022), so we can use their impl as reference again and maybe borrow some tests too. They seemed to have refactored the whole thing heavily though, squashing both tree classifier/regressor into one [common](https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/core/providers/cpu/ml/tree_ensemble_common.h) engine.

Cheers :)

kali:
So... status update

* @bminixhofer was kind enough to generate test case for both regressor and classifier, using two frameworks. See #360. This branch is merged on this one.
* I had to do a few fixes on the tree loading code (mostly +1/-1 fixes).
* I fixed all the tract aspect of the classifier, and found relatively easily how to call the tree code to implement eval.
* I decided against implementing the operator with the missing N dimension. I think it is ok to implement "accidently" more than the spec, but writing complicated code to extend an already greedy specification is a different story. As a consequence, there might be a bit of dead code.
* At this point we have the classifier working (in "plain" mode, and "optim" mode) for the first generating strategy. If I remember correctly, the other strategy is blocked by the OneHot operator (which will be coming soon) so we'll not worry about it for now.

What we need to do next:

* if it is trivial (or simple), add support for the ONNX tree reggressor node: this is where we need your help, at least for the business side of thing, I can do the tract bits.
* finish/cleanup code (dead code from the 0D case)

And later (I will probably do that myself, nothing is baked at this stage)

* Split onnx-ml into its two separate crates, onnx-ml-opl and onnx-ml (with the label lookup operators) adding support for NNEF serialization and loading.
* This requires prior work at the OPL/NNEF level: we need an extension for String tensor serialization.

PR #366, opened 2020-09-18
Made a better version of the Keras->tract tutorial

mattalhonte:
Updated the Keras -> tract tutorial, taking advantage of the onnx support.  Now much cleaner, without a bunch of futzing with intermediate file formats!

One issue is that I wasn't sure how to do the production profiling.


kali:
This is great ! I'm wondering though if we don't want to keep the other "path" though TF1 as a separate "recipe" (maybe with a recommendation to move to ONNX instead and linking to this new one).

As for profiling... do you mean using tract command line to get a network profile ? tract command line is not documented, except by the --help flags. I will try to remedy this a bit. In the meantime you can try something like `tract my_model.onnx -O dump --cost --profile` or `tract my_model.onnx -O bench` or even `tract my_model.onnx -O criterion`.

mattalhonte:
Sure, I could put it as another recipe!
For profiling, I mean the equivalent of this part on the old version
> Compiled in release mode, the call to run() is clocked at 6 microseconds (0m0.000006s !) on one single core.

I didn't actually write that part, and I'm not sure how to check that the `run()` specifically does - seems different than just doing `time cargo run`, which is kinda the only thing I know at the moment.

kali:
I probably did write that (can't remember), then, and i'm pretty sure it was using `tract foo.onnx -O bench`.

mattalhonte:
Ah, cool!  I'll give it a go in the morning!

mattalhonte:
Sorry, not sure how the benchmarking was supposed to work - I tried `tract my_model.onnx -O bench` in the `tract/examples/jupyter-keras-tract` folder but just got `Command 'tract' not found`.  I'll try and put it in a separate folder for now, though!

kali:
Ho ! You need to install it :) `cargo install tract` will do. It's a bit long... but according to my ML colleagues it's worth it. It can help a lot with auditing and profiling networks.

mattalhonte:
Aha!

Hrm, so I tried 
`tract example.onnx -O bench`

And got
```
Error: Translating node #0 "main_input_input" Source ToTypedTranslator
    Caused by: Source node without a determined fact. Help: provide explicit input facts to your model.
```

I thought I did have input facts in my script, though?  This line:
`    .with_input_fact(0, InferenceFact::dt_shape(f32::datum_type(), tvec![10, 100]))?`

kali:
Yes ! But tract does not know about your code, so you need to give it the same hint: just add `-i 10x100xf32` before the bench.

You may want to try `tract example.onnx -O -i 10x100xf32 dump --profile --cost` to get an idea of what it can do for you.

mattalhonte:
Aha!  Worked!  Lovely output - particularly like the "Most time consuming operations" summary at the bottom.

Which number do I use as the final benchmark?

kali:
IIRC, the last number should the time per iteration for the entire net, that's the one you want.

kali:
Are we good for merge ? Or what can I do to help ?

mattalhonte:
Anything I can do to fix the failed run?  https://github.com/mattalhonte/tract/actions/runs/288631477

kali:
All right, gha did not let me see the ful logs on your run, but it seems fine right here. ok to merge ?

mattalhonte:
Should be good!

PR #371, opened 2020-10-07
Add 'ai.onnx' as default domain

acrrd:
Add 'ai.onnx' as a valid default domain name.

I did not found any official documentation that state that it is a possiile value for the default domain but I have a model that use it and I found it mentioned here as a default domain:
https://github.com/onnx/onnx/issues/688
https://github.com/onnx/onnx/blame/master/onnx/version_converter/convert.h


kali:
All right, looks reasonable enough :) thanks !

PR #381, opened 2020-10-21
Implement DynamicQuantizeLinear

### LINKED ISSUE #380: Implement DynamicQuantizeLinear
Implement the onnx operator DynamicQuantizeLinear.
https://github.com/onnx/onnx/blob/master/docs/Operators.md#DynamicQuantizeLinear.

acrrd:
This PR implement the onnx operator [DynamicQuantizeLinear](https://github.com/onnx/onnx/blob/master/docs/Operators.md#DynamicQuantizeLinear).
I could not reuse `quantize_linear_u8` because it does `x * scale.recip()` instead of `x / scale`, This change some number enough to be rounded to another integer and the onnx testsuite fails.

Is not currently possible to implement it using the operations already provided (like is done in the `dynamicquantizelinear_expanded` test) because not all needed operations are `TypedOp` and in particular `QuantizeLinear` expect the zero point and the scale to be constant tensor.

Close #380

### DIFF
--- a/core/src/ops/quant.rs
+++ b/core/src/ops/quant.rs
@@ -1,5 +1,6 @@
 use crate::internal::*;
 use crate::ops::element_wise::ElementWiseOp;
+use ndarray::ArrayViewD;
 use num_traits::AsPrimitive;
 use num_traits::Zero;
 use tract_linalg::lut::Lut;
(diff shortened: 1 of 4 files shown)


core/src/ops/quant.rs
```diff
@@ -146,6 +147,110 @@ fn info_quantize_linear_i8(q: &QuantizeLinearI8) -> TractResult<Vec<String>> {
...
+        Ok(tvec!(quantized_tensor, scale_tensor, zero_point_tensor))
+    }
+}
+
+impl TypedOp for DynamicQuantizeLinearU8 {
+    fn output_facts(&self, inputs: &[&TypedFact]) -> TractResult<TVec<TypedFact>> {
+        let mut quantized_fact = inputs[0].clone();
+        quantized_fact.datum_type = u8::datum_type();
+        let shape = ShapeFact::from_dims(&[])?;
+        let scale_fact = TypedFact::dt_shape(f32::datum_type(), shape.clone())?;
```
kali, line 246: I haven't tried it, but i think `TypedFact::dt_shape(f32::datum_type(), [].as_ref())?;` should work

kali (changes requested):
Kudos on going so far on your own ! First time an outsider gives me a fully baked op without a huge amount of help :)

core/src/ops/quant.rs
acrrd: The compiler cannot infer the type with just that. I ended up with:
```
let shape: &[usize] = [].as_ref();
let scale_fact = TypedFact::dt_shape(f32::datum_type(), shape)?;
```

kali: All right, thanks for trying 👍

kali (approved):

Issue #377, opened 2020-10-19
Very slow declutter in Debug mode


kali:
See #374 part 2.

kali:
hey @bminixhofer, do you think we've fixed it ?

bminixhofer:
No, still the same for me with `https://github.com/sonos/tract?branch=main#9cc80c35`:

`src/main.rs`
```rust
use tract_onnx::prelude::*;

fn main() -> TractResult<()> {
    let _model = tract_onnx::onnx()
        .model_for_path("model.onnx")?
        .with_input_fact(0, InferenceFact::dt_shape(u8::datum_type(), tvec!(1, 10)))?
        .into_optimized()?;

    Ok(())
}
```
Output:

```
(base) bminixhofer@pop-os:~/Documents/Experiments/tractdyn$ time cargo run --release
    Finished release [optimized] target(s) in 0.06s
     Running `target/release/tractdyn`

real    0m0.151s
user    0m0.099s
sys     0m0.055s
(base) bminixhofer@pop-os:~/Documents/Experiments/tractdyn$ time cargo run 
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
     Running `target/debug/tractdyn`

real    3m26.495s
user    3m25.117s
sys     0m0.156s
```

where `model.onnx` is again the model from https://github.com/bminixhofer/nnsplit/blob/master/models/fr/model.onnx.

kali:
All right, I think this one is the killer. I have feature gated a lot of assertions that are indeed too costly and irrelevant in day to day usage .

bminixhofer:
Yes! Back to normal for me with https://github.com/sonos/tract?branch=introduce-paranoid-assertions#457195a4:

```
(base) bminixhofer@pop-os:~/Documents/Experiments/tractdyn$ time cargo run --release
    Finished release [optimized] target(s) in 0.05s
     Running `target/release/tractdyn`

real    0m0.123s
user    0m0.111s
sys     0m0.016s
(base) bminixhofer@pop-os:~/Documents/Experiments/tractdyn$ time cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/tractdyn`

real    0m1.366s
user    0m1.335s
sys     0m0.036s
```

Thanks!

PR #384, opened 2020-10-22
less assertions on regular debug

kali:


kali:
Should help with #377

PR #385, opened 2020-10-24
Lut ub

kali:


kali:
FYI, @acrrd I introduced a regression in the lookup table (in the changed commited under the previously failling rebase), and I noticed you were using it.

PR #400, opened 2020-11-03
Add generic multiplication for i8 u8

acrrd:
This allow to multiply two matrices with different type (`i8` and `u8`).
This add only the generic multiplication and not the optimized one.

### DIFF
--- a/core/src/ops/matmul/mir.rs
+++ b/core/src/ops/matmul/mir.rs
@@ -38,6 +38,18 @@ fn eval(
                     MMMWrapper::Quant((tract_linalg::ops().qmmm_u8_u8)(m, k, n))
                 });
             }
+        } else if (a.datum_type(), b.datum_type(), q.c_datum_type)
+            == (i8::datum_type(), u8::datum_type(), i32::datum_type())
+        {
+            return eval_t(a, b, a_trans, b_trans, c_trans, q_params, &|m, k, n| {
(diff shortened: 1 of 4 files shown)


kali:
Thanks for this. I need to think calmly about whether or not we can implement this op (and all the i8/u8 combinations actually) using the i8xi8 assembly codes, or how much we need to alter them.

In any case, I don't plan on merging it as is actually, because I have an ungoing refactoring on the linalg/core interface that will conflict awfully. I plan on getting at a mergeable state on my refactoring branch later this week, or maybe next week, then rebase this branch (it should be pretty straightforward).

That is... unless you're in a rush and need it merged quickly.

acrrd:
This PR will make the model run, but without the optimizations is too slow anyway so we can wait. Let me know how can I assist for this.

kali:
Can I ask what is your target ? intel ? avx2 ?

acrrd:
Sure, we are interested in running it on mobile, so arm64/armv7.

kali:
@acrrd  Could you please check that #404 is working ?

acrrd:
Sorry for the delay. Is working thanks! Is this already using the optimized assembly code?

kali:
Thanks for checking.

No, it's basically just a rebase of what you did using the new linalg interface, so it's still relying on the generic rust fallback.

The good news is, I think I have a way to do the four combination without dealing with fours times the code, with very limited overhead. But it may take me a while to get there.

kali:
Will use #413 for optimised implementation progress.

PR #402, opened 2020-11-11
Fix segfault on iphone

acrrd:
We are experiencing a segmentation fault on iphone (at least iphone SE 2gen, X, 11) at [arm64simd_mmm_f32_8x8_core.tmpl#L466](https://github.com/sonos/tract/blob/fb397adfc61e7f054149ad757ee33411c16bbc59/linalg/arm64/arm64simd/arm64simd_mmm_f32_8x8_core.tmpl#L466).

The first commit where this happen is ['try to exploit double issue'](https://github.com/sonos/tract/commit/0be2578fca61a40d44835d0539439e8fad20432c). My guess is that at the end of the last iteration of `.packed_packed_loop_1` we are doing a `ld1` to prepare for the next iteration but this is the last one and then we are reading data that we are not suppose to.

I tested this on the devices mentioned before and on android devices (where it as already working fine), I do not have a device with `a5x` to test on.

### DIFF
--- a/linalg/arm64/arm64simd/arm64simd_mmm_f32_8x8_core.tmpl
+++ b/linalg/arm64/arm64simd/arm64simd_mmm_f32_8x8_core.tmpl
@@ -286,10 +286,11 @@
     b           .non_linear
 
 .packed_packed:
+
+{% if core == "a5x" %}
     ld1         { v0.4s, v1.4s }, [ x1 ], #32
     ld1         { v4.4s, v5.4s }, [ x2 ], #32
 
-{% if core == "a5x" %}
     cmp         x3, #4
     blt         .packed_packed_loop_1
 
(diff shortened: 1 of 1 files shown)


kali:
Hi ! Thanks for tracking this down. I'm not greatly surprised, the last read is indeed invalid. The fix is relatively easy, and I think we may have had it in the past : as we are operating on two "packed" matrices as inputs here, we can just have the packers pack a bit of padding at the end to make the read valid without requiring the assembly kernel to get smarter. I'll try to have a look at this as soon I am done with the long going refactoring as it is also in the conflicting area.

acrrd:
Ok, thanks.

kali:
@acrrd  finally managed to merge this big regactoring :) Could you try is #403 fix the issue ?

PR #403, opened 2020-11-12
setup kernel-configurable pack padding at end

kali:


kali:
temptative fix for #402

acrrd:
It works! Thanks

PR #414, opened 2020-11-27
Fix NNEF model reader

russellwmy:
Problem: The code for reading an NNEF model file with ext `.tgz` is not taking care of the GZ decoding.
Quick Solution: Added flate2 library to decode the file before read by tar.
Ideal Solution: Detect the file extension and add GZ decoding for `.tgz` file only. But it is not in scope as the NNEF tools always generate `.tgz` file.

### DIFF
--- a/nnef/Cargo.toml
+++ b/nnef/Cargo.toml
@@ -17,6 +17,9 @@ maintenance = { status = "actively-developed" }
 log = "0.4"
 nom = "5"
 tar = "0.4"
-flate2 = "1.0"
+flate2 = { version = "1.0", optional = true }
 tract-core = { path = "../core" }
 walkdir = "2"
+
+[features]
+default = ["flate2"]
--- a/nnef/Cargo.toml
+++ b/nnef/Cargo.toml
@@ -17,5 +17,6 @@ maintenance = { status = "actively-developed" }
 log = "0.4"
 nom = "5"
 tar = "0.4"
+flate2 = "1.0"
 tract-core = { path = "../core" }
 walkdir = "2"
--- a/nnef/src/framework.rs
+++ b/nnef/src/framework.rs
@@ -1,5 +1,5 @@
 use std::path::Path;
-
+use std::io::{Read, Seek, SeekFrom};
 use crate::ast::ProtoModel;
 use crate::internal::*;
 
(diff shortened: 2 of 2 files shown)


kali:
Hey! Thanks for your interest.

I need tract-nnef to stay independent of gziping. Zipping the model is not that beneficial: weights tensor do not compress well, while the cost of unzipping at launch time can get in the way. So I need to be able to feed the model as a tar stream, leaving the choice of compressing or not to the caller.

You will also notice that the command line interface (in cli/, crate name "tract") actually deal with some file name extension.

How about a compromise ? How about putting the dependency to flate2 plus a specific method on nnef() behind a a feature flag ?

russellwmy:
It is because the examples of NNEF is loading file with `.tgz`. That's why I am thinking it is a bug here. if that is the case, can I correct the file extension of  the examples ?

Here is the line has the issue: https://github.com/sonos/tract/blob/366d09824b36f1ce6f9259bb7c576bf581e405aa/examples/nnef-mobilenet-v2/src/main.rs#L6

kali:
Awww, right, this is a bug, indeed. I used to accept .tgz, actually, I just changed the library when I realized that the gzipping could be so counterproductive... and forgot this example (as a side note, examples should be covered by CI, and they're obviously not...)

This example model file is a tgz, and the example code should demonstrate how to run .tgz models as this is how nnef models exist in the wild, so we need to reconcile all of that.

So we have two options, I think.

1/ make the library accept .tgz and .tar, detecting the file name extension in the `*_for_path` methods, and expose different methods for a tar stream and a tgz stream. If we go this way, we still need to have the zipping features behind a feature flag (which could be enabled by default) so that I can still have the library without the gzipping bloat when I need it. In that case, the method accepting .tgz only should go away too, and the method with autodection should return an explicit error.
2/ Just document the fact that the _for_path method of tract-nnef only accept a tar file or a directory, and that "standard" tgz NNEF models should be loaded by calling flate2 and pipe it to model_for_read. And of course fix the example accordingly.

I have a small preference for 2/, because I'm not a big fan of feature flags, and 1/ also makes us introduce a runtime-only error if the feature is missing.

WDYT ?

russellwmy:
Lets think about the production use cases.

Option-1: if developer enabled the gzipping feature, developer can load the NNEF straight without any extras conversion. tract-nnef will take care that. It will need slightly more CPU usage.
Option-2: Since tract-nnef only supports tar, developer have to convert the file before load with tract-nnef.

I would prefer only accept `tgz`. Since people are using this format to store the model data. the `tar` will be less useful. But if you have concerns about the load time. I will go option-1.

Even NNEF github output models in `.tgz`. https://github.com/KhronosGroup/NNEF-Tools/tree/master/models#nnef-model-zoo

Let me know your thoughts, I will adjust the code accordingly.

kali:
Not accepting raw tar is not an viable alternative. On embedded devices, we have use cases where the loading time (and to a lesser extent code bloat) are a blocking issue. I do know that this is not what the specification suggests (and only suggest, iirc).

So ok, how about this:

* a `tgz` compilation feature on tract-nnef
* `tgz` is activated by default
* flate2 dependency is conditioned by the compilation flag
* the `*_for_path` methods check for a "tgz" extension, and wraps the file in a flate2 decoder accordingly, or returns an error if the feature is disabled

This way we have the example code working again as is. And the library will work as expected, out of the box, with models from the zoo.

I'm not too sure about what to do with the two `*_for_read` methods... how about we check the magic numbers for .gz and switch accordingly ?

russellwmy:
Sound good to me. Let me  update the PR.

russellwmy:
@kali , Please review the code changes.

kali (changes requested):
Rest LGTM

russellwmy:
It would be a permission issue. 
Somehow, I cannot see your comments on change request.

kali:
Weird. I was asking to please replace the eprintln + process::exit by a bail!() call.

kali (approved):

kali:
Thanks for doing this!

russellwmy:
I am very interested in this project. Let me know if you need any help in the future.

PR #415, opened 2020-11-29
Refactored the fix for NNEF gzip file detection

russellwmy:
I reviewed the code again, developer is still able to call the `proto_model_for_read` method with reader parameter. In this case the gzip detection is skipped. I moved the check inside `proto_model_for_read`, so that we can avoid this issue.

BTW, do you have `mobilenet_v2_1.0.onnx.nnef.tar` file for testing. I tried to create one with  decompress and compress again with tar command. But it give me this error: `Error: Wrong magic number`.

### DIFF
--- a/nnef/src/framework.rs
+++ b/nnef/src/framework.rs
@@ -1,5 +1,5 @@
 use std::path::Path;
-use std::io::{Read, Seek, SeekFrom};
+use std::io::{Read};
 use crate::ast::ProtoModel;
 use crate::internal::*;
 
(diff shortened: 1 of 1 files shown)


kali:
Good call moving the check at the proto_model place.

you should get the model in raw tar form by just doing: `gunzip mobilenet_v2_1.0.onnx.nnef.tar.gz`

russellwmy:
@kali  I tested both `tar` and `tgz`. In both cases, the example is looking good. I changed a better name for new variables.
Please review the PR.

kali (approved):

kali:
This looks good... Can we just have it squashed and rebased against the current main ?

russellwmy:
@kali  Finally, finish your change requests. it took me sometime to learn rebase.
I think it all good now.

kali:
Aha, yeah rebase is tricky, very powerful and a bit scary. You will use it again :)

PR #424, opened 2020-12-16
Add a "cargo deny" config.

hrydgard:
To keep our dependency tree sane, we use `cargo deny`. Thought it might be useful for tract as well.

After merging this, all you need to do is: `cargo install cargo-deny`, and then, `cargo deny check`, from the tract root directory.

It warns about security advisories for crates in the entire dependency tree, detects duplicate versions, bad licenses, and has a mechanism where you can ban crates you don't like from your whole dependency tree.

I've preconfigured this one to be suitable for tract, but of course change whatever you like - and I'd recommend to add this to CI as well.

As is, it will give you two security advisory warnings, about the `memmap` and `image` crates. I have suppressed a few bans of duplicate crates, you'll see those warnings if you remove those lines.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,0 +1,44 @@
+# add whatever else we support.
+targets = [
+    { triple = "x86_64-unknown-linux-gnu" },
+    { triple = "x86_64-apple-darwin" },
+    { triple = "x86_64-pc-windows-msvc" },
+    { triple = "aarch64-linux-android" },
+]
+
+[advisories]
+vulnerability = "deny"
+unmaintained = "warn"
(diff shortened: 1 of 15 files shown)


kali:
Thanks, that's useful indeed. I am very cautious about not letting GPL slip in, but you never know what can happen with transitive dependencies, so this can become a great mental load reliever.

I think I will include it as one of the "non-critical" CI, like the beta and nightly checks as a start.

I should probably also do that with rustfmt. And also clippy, but I'm terrified of the number of errors it will raise.

kali:
Merged as part of 425

PR #425, opened 2020-12-16
Embark add cargo deny config

kali:


kali:
@hrydgard any opinion on the right way to fix/ignore the few remaining ones ?

hrydgard:
Hm, I guess some of those showed up when you added the rest of your platform triples? 

cfg-if is a common one and not very critical, I'd just skip it for now.

crossbeam-channel, I'd skip it but also send a PR to "pbr" asking them to upgrade their dependency, since it's just a single crate that depends on the old version. 

miniz_oxide, I dunno, looks like a lot of crates using the old version. Might be more work than it's worth until some more of them come around. 

There's no huge value in being completely clean here, just not letting the duplicates spiral out of control. It's also good for the whole ecosystem to upgrade exceedingly old dependencies of various crates...

kali:
Mmm I don't think the platform list did it, actually... but I observe some variations in the list depending on it running it on a regular linux box and my M1 mac...

As for rand, I'm surprised, I don't know what is happening. I'll have a look :)

PR #434, opened 2021-01-12
Move pulse to dev deps

acrrd:
This is mainly because tract-pulse depends on ctor that does not work on wasm. tract-pulse is only used in tests.


acrrd:
Sorry there are commits that do not belongs here, I will clear it asap.

acrrd:
Cleared

kali:
I'm actually removing ctor from pulse, and I think I can also have the hir tests work without this dep... thanks for putting this under my nose.

acrrd:
Thank you!

PR #478, opened 2021-04-28
Fix groups in quantized convolution

TimotheeLeBerresonos:
This PR fix groups in quantized convolutions, and in fused operations.


kali:
@TimotheeLeBerresonos could you rebase this on main ? there have been some performance changes on main, and I want to double check we are not drifting with this branch...

PR #480, opened 2021-04-29
Added assert-op-count to cli

TimotheeLeBerresonos:
Added an assert-op-count argument to the run and dump subcommands

### DIFF
--- a/cli/src/dump.rs
+++ b/cli/src/dump.rs
@@ -34,6 +34,19 @@ pub fn handle(
             .collect::<TractResult<Vec<InferenceFact>>>()?;
         crate::utils::check_inferred(&*outputs_facts, &*asserts)?;
     }
+    if let Some(asserts) = &params.assertions.assert_op_count {
+        for (name, expected) in asserts {
+            let count = model
+                .eval_order()
+                .context("Cannot assert op count without an eval order")?
+                .into_iter()
+                .filter(|&i| &*model.node_op(i).name() == &*name)
+                .count();
+            if count != *expected {
+                bail!("Wrong number of {} operators: expected {}, got {}", name, expected, count);
+            }
+        }
+    }
 
     if let Some(path) = sub_matches.value_of("nnef") {
         let nnef = super::nnef(&matches);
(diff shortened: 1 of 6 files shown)


cli/src/dump.rs
```diff
@@ -34,6 +34,19 @@ pub fn handle(
             .collect::<TractResult<Vec<InferenceFact>>>()?;
         crate::utils::check_inferred(&*outputs_facts, &*asserts)?;
     }
+    if let Some(asserts) = &params.assertions.assert_op_count {
+        for (name, expected) in asserts {
+            let count = model
+                .eval_order()
+                .context("Cannot assert op count without an eval order")?
+                .into_iter()
+                .filter(|&i| &*model.node_op(i).name() == &*name)
```
TimotheeLeBerresonos, line 43: I used eval_order to iterate over all the nodes in the model. Is there a better way ?

kali: It's probably the best way to go. you could also iterate over all the nodes in the model, it would work too. In most cases it will do the same thing, the only case it is different is when there are dead nodes in the graph. For what we are doing it should not matter too much.

But it misses the nested models ! You also need to iterate on the subgraphs or we will miss all the Scan internal nodes... look for "nested_models" in the cli, there are already a few locations when this is done.

cli/src/params.rs
```diff
@@ -882,12 +883,24 @@ impl Assertions {
...
+                .values_of("assert-op-count")
+                .map(|vs| {
+                    vs.chunks(2)
+                        .into_iter()
+                        .map(|mut args| {
+                            Some((args.next()?.to_string(), args.next()?.parse().ok()?))
+                        })
+                        .collect()
+                })
+                .flatten();
```
kali, line 896: So how do we make this parser happy ? :) `--assert-op-count Add,2,Mul,4` ? 

We want an exemple in the help message (lines 323 and 269)

cli/src/utils.rs
```diff
@@ -1,4 +1,4 @@
-use crate::CliResult;
+use crate::{CliResult, model::Model};
```
kali, line 1: Nitpicking: I really find these hard to parse, tbh. I don't mind flat {} but prefer when submodules get their own line. I'd like it if we can tweak your environment so it does not generate them...

kali:
All right, I guess the last step here is to create a few of these tests. Add, Mul and Max on AM feel like right candidates.

kali (approved):

PR #484, opened 2021-05-03
more op count checks

kali:


kali:
FYI @TimotheeLeBerresonos

PR #510, opened 2021-06-29
Adding Range operator to ONNX

robin-des-bois90:
Adding Range operator to ONNX and example program running transformers.

### DIFF
--- a/examples/onnx-distilbert-vanilla/Cargo.toml
+++ b/examples/onnx-distilbert-vanilla/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "example-onnx-distilbert-vanilla"
+version = "0.1.0"
+authors = ["Can"]
+license = "MIT/Apache-2.0"
+edition = "2018"
+
+[dependencies]
+image = "*"
+tract-onnx = { path = "../../onnx" }
--- a/examples/onnx-distilbert-vanilla/src/main.rs
+++ b/examples/onnx-distilbert-vanilla/src/main.rs
@@ -1,0 +1,26 @@
+use tract_onnx::prelude::*;
+
+fn main() -> TractResult<()> {
+    let model = tract_onnx::onnx()
+    // load the model
+    .model_for_path("bert_for_classification.onnx")?    // Model : DistilBERT; HuggingFace transformers
+    // specify input type and shape : imported model works with 1x8 input tensors
+    .with_input_fact(0, InferenceFact::dt_shape(i64::datum_type(), tvec!(1, 8)))?
+    // optimize the model
+    .into_optimized()?
+    // make the model runnable and fix its inputs and outputs
+    .into_runnable()?;
+
+    // Inputting dummy tensor : ones(1x8)
+    let elems: i64 = 1;
+    let dummy_input: Tensor = tract_ndarray::Array2::from_shape_fn((1, 8), |(_,_)| {elems}).into();
+    println!("Input : {:?}", dummy_input);
+
+    // run the model on the input
+    let result = model.run(tvec!(dummy_input))?;
+
+    // PyTorch output = ONNX output = [[[ 0.01977395 -0.02752648 ]]]
+    // Printing Tract output :
+    println!("Output : {:?}", result[0]);
+    Ok(())
+}
--- a/onnx/src/ops/array/mod.rs
+++ b/onnx/src/ops/array/mod.rs
@@ -3,6 +3,7 @@ mod nonzero;
 mod one_hot;
 mod pad;
 mod slice;
+mod range;
 
 use tract_hir::internal::*;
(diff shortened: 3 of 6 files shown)


examples/onnx-distilbert-vanilla/src/main.rs
```diff
@@ -0,0 +1,26 @@
+use tract_onnx::prelude::*;
+
+fn main() -> TractResult<()> {
+    let model = tract_onnx::onnx()
+    // load the model
+    .model_for_path("bert_for_classification.onnx")?    // Model : DistilBERT; HuggingFace transformers
```
kali, line 6: Please add a minimum README file to the example to tell how to download the model.

onnx/src/ops/array/range.rs
```diff
@@ -0,0 +1,104 @@
...
+use tract_num_traits::*;
+
+use tract_hir::internal::*;
+
+use crate::model::ParsingContext;
+use crate::pb::*;
+
+#[derive(Debug, Clone, new, Hash)]
+pub struct Range {
+    dtype: DatumType,
```
kali, line 12: Except on very old TF ops, I don't try and store the operating type except if I need it explicitely. I think you can remove this.

onnx/src/ops/array/range.rs
```diff
@@ -0,0 +1,104 @@
...
+    not_a_typed_op!();
+}
+
+impl EvalOp for Range {
+    fn is_stateless(&self) -> bool {
+        true
+    }
+
+    fn eval(&self, inputs: TVec<Arc<Tensor>>) -> TractResult<TVec<Arc<Tensor>>> {
+        dispatch_numbers!(Self::eval_t(self.dtype)(self, inputs))
```
kali, line 57: here you can dispatch on inputs[0].datum_type

onnx/src/ops/array/range.rs
```diff
@@ -0,0 +1,104 @@
...
+        if let (Some(start), Some(limit), Some(delta)) = (
+            target.outlet_fact(mapping[&node.inputs[0]])?.konst.as_ref(),
+            target.outlet_fact(mapping[&node.inputs[1]])?.konst.as_ref(),
+            target.outlet_fact(mapping[&node.inputs[2]])?.konst.as_ref(),
+        ) {
+            let mut value = dispatch_numbers!(Self::eval_t(start.datum_type())(
+                self,
+                tvec!(start.clone(), limit.clone(), delta.clone())
+            ))?;
+            Ok(tvec!(target.add_const(&*node.name, value.remove(0))?))
```
kali, line 99: I'm not sure this is the right call here. I would expect one of the inputs of the Range op to be the a symbolic dimension (as it will likely be the input length)... Making it accept symbolic dimensions is a bit of work though : as it will no longer be a constant, it will have to be moved to core or an -opl crate.

The impact on performance could be significant, it should allow you to run into_optimized() before fixing the input length (which is usefull for NLP as input as utterances tend to have varrying sizes).

onnx/src/ops/array/mod.rs
```diff
@@ -35,6 +36,7 @@ pub fn register_all_ops(reg: &mut OnnxOpRegister) {
     reg.insert("Squeeze", squeeze);
     reg.insert("Tile", |_, _| Ok((expand(array::Tile::default()), vec![])));
     reg.insert("Transpose", transpose);
+    reg.insert("Range", range::range);
```
kali, line 39: Try to see in harness/onnx-test-suite if the `include-passing-ignored.sh` picks up more passing tests with Range and adds them to the list of passing tests.

You can also run `cargo test --ignored range`.

onnx/src/ops/array/range.rs
```diff
@@ -0,0 +1,104 @@
...
+        } else {
+            bail!("Can not type Fill op")
```
kali, line 101: typo ?

examples/onnx-distilbert-vanilla/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "example-onnx-distilbert-vanilla"
+version = "0.1.0"
+authors = ["Can"]
```
kali, line 4: typo ?

kali:
Yop, just realised that I was actually reviewing code I wrote 3  or 4 years ago and that you merely copied it around... most of the remarks still hold though.

The main question reamains: how will transformers be using Range ? Will the boundary values be always the same ? or are they dependend on the input length ? Ultimately, that's going to decide if we can substitue by a constant early or need an implementation in tract-core.

kali:
I have ported Range to hir instead (applying my earlier suggestions in the mean time), so the main goal of this PR is achieved. It still have merits in the inclusion of one more network example as a test/example if you want to pursue this.

Issue #413, opened 2020-11-25
i8xu8 (and friends) matrix products


kali:
Can we just use the i8xi8 path by applying an offset and playing with zeropoint ?

janpetschexain:
Hi, I'm working with @acrrd. In general it works, all signed and unsigned byte matrix multiplications can be expressed as a signed byte matrix multiplication with some correctional terms. The main question probably will be, who much work it is to support those additional terms in the assembly code. Please have a look at the following explanation.

Let the matrices of shape and data type (i'll try to use some pseudo notation due to the lack of Latex in markdown, if it helps i can also attach an image of the rendered formulas):
- `Ai = (ai)_mk` of shape `(M, K)` with signed elements `ai` of `i8`
- `Bi = (bi)_kn` of shape `(K, N)` with signed elements `bi` of `i8`
- `Au = (au)_mk` of shape `(M, K)` with unsigned elements `au` of `u8`
- `Bu = (bu)_kn` of shape `(K, N)` with unsigned elements `bu` of `u8`

We have to consider three different cases:

**1)**

`Ai * Bi` is the already implemented signed byte matrix multiplication with results in `i32`. The `i32` accumulator doesn't overflow for `K < 2^17` and the quantization scale and zero-point are as usual.

**2)**

`Ai * Bu` and `Au * Bi` are the mixed signed and unsigned byte matrix multiplications. Both can be transformed into each other via matrix transpose as `Au * Bi = ((Au * Bi)^t)^t = (Bi^t * Au^t)^t`, but for the sake of completeness let's look at them individually.

**2.i)**

Let `bi := bu - 128` and to simplify notation only consider the `K` index in the following sums. Then it holds for any element of the product `Ai * Bu` that
```
sum_k ai_k * bu_k = sum_k ai_k * (bu_k - 128 + 128) = sum_k ai_k * bi_k + 128 * sum_k ai_k
```
where the first term is the usual signed byte matrix multiplication and the second term can be precomputed once for all rows of `Ai`. The `i32` accumulator doesn't overflow for `K < 2^16`. The zero points `z_Bu` of `Bu` and `z_Bi` of `Bi` are related by the same offset, i.e. `z_Bi := z_Bu - 128`. With scales `s_Ai` of `Ai` and `s_Bu` of `Bu` and the usual mappings `v = s_Ai * (ai - z_Ai)` and `w = s_Bu * (bu - z_Bu)` the product can be expressed in terms of the original parameters as
```
[...]

PR #522, opened 2021-07-16
fix relative indexing in qparams inline static & add tests

janpetschexain:
**Problem**

during the work on #413 we encountered indexing errors in the qparams inlining during codegen under certain circumstances, depending on the relative order of the inputs compared to the order of qparams. for example at one point of the codegen we have the following node

```rust
Node {
    id: 119,
    name: "MatMul_101_quant",
    inputs: [117/0>, 117/2>, 118/0>],
    op: QMatMulUnary {
        a: 1,128,512,I8 35, 39, -1, 30, -13, -25, 32, 8, -19, 6, 33, 47...,
        bias: None,
        a_trans: true,
        b_trans: true,
        c_trans: true,
        output_type: I32,
        params: MatMulQParams {
            a0: Input(2),
            a_scale: Attr(,F32 1),
            b0: Input(1),
            b_scale: Attr(,F32 1),
            c0: Attr(,I32 0),
            c_scale: Attr(,F32 1),
        },
    },
    outputs: [1,64,512,I32 >120/0],
}
```

where the inputs are `b`, `b0` and `a0` and the inlining is going to replace the `a0`, resulting in a wrong index for `b0`

```rust
Node {
    id: 228,
    name: "MatMul_101_quant",
    inputs: [227/0>, 227/2>],
    op: QMatMulUnary {
        a: 1,128,512,I8 35, 39, -1, 30, -13, -25, 32, 8, -19, 6, 33, 47...,
        bias: None,
        a_trans: true,
        b_trans: true,
        c_trans: true,
        output_type: I32,
        params: MatMulQParams {
            a0: Attr(,I8 24),
            a_scale: Attr(,F32 1),
            b0: Input(0),
            b_scale: Attr(,F32 1),
            c0: Attr(,I32 0),
            c_scale: Attr(,F32 1),
        },
    },
    outputs: [1,64,512,I32 >229/0],
}
```

**Solution**

- fixes the relative indexing: the `remove_input()` must not be called on the `position` which is relative to the `MatMulQParams` order, but on `ix - number_removed` which is relative to the `node.inputs` order (where `number_removed` can be calculated as `ix - inputs.len()`)
- add three tests, where the second and third test wouldn't previously pass (and the first would just pass due to the coincidental same order of `position` and `ix`)

### DIFF
--- a/core/src/ops/matmul/mir_quant.rs
+++ b/core/src/ops/matmul/mir_quant.rs
@@ -76,9 +76,7 @@ impl MatMulQParams {
                 model.outlet_fact(*input)?.konst.as_ref(),
             ) {
                 *new.iter_mut().nth(position).unwrap().1 = AttrOrInput::Attr(k.clone());
-                for qp in new.iter_mut() {
-                    qp.1.remove_input(position);
-                }
+                new.remove_input(inputs.len());
             } else {
                 inputs.push(*input)
             }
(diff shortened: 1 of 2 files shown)


kali:
Nice catch ! Thanks :)

PR #536, opened 2021-08-05
pytorch_half_pixel coordinate transform mode

pYtoner:
I hit a todo when trying to run an onnx model that was created by `pytorch`. The coordinate transform for `"pytorch_half_pixel"` was not implemented. I think we can just use the `"half_pixel"` implementation for `"pytorch_half_pixel"`.

### DIFF
--- a/onnx/src/ops/resize.rs
+++ b/onnx/src/ops/resize.rs
@@ -11,6 +11,7 @@ pub fn resize(
         match node.get_attr_opt("coordinate_transformation_mode")?.unwrap_or("half_pixel") {
             "align_corners" => CoordTransformer::AlignCorners,
             "half_pixel" => CoordTransformer::HalfPixel,
+            "pytorch_half_pixel" => CoordTransformer::HalfPixel,
             s => todo!("coordinate_transformation_mode: {}", s),
         };
     let interpolator = match node.get_attr("mode")? {


pYtoner:
It seems to work for the simple example I tried at least.

kali:
I tried to fix it "the right way" (according to spec) in the `new-resize-coord-modes`. No test coverage as the pytoch_half_pixel mode is only covered with downsampling, which is still a todo. Could you validate with your simple example that my branch also does the right thing on your example ?

pYtoner:
No, unfortunately it doesn't work.
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Failed analyse for node #108 "Resize_53" Resize

Caused by:
    0: Infering facts
    1: Applying rule inputs[2].shape[0] == {(inputs[0].rank) as dim}
    2: Impossible to unify Val(0) with Val(4).', src/main.rs:58:10
```

The `unwrap` in question is the unwrap after `into_optimized`.
```rust
let model = tract_onnx::onnx()
        .model_for_path("detector.onnx")
        .unwrap()
        .with_input_fact(
            0,
            InferenceFact::dt_shape(f32::datum_type(), tvec!(1, 3, height, width)),
        )
        .unwrap()
        .into_optimized()
        .unwrap()
        .into_runnable()
        .unwrap();
```

pYtoner:
I'll make a public repo containing the example so you can get to the bottom of it if you want.

PR #542, opened 2021-08-30
Complex number support in Tensor

hdlj:


kali (approved):

kali:
It looks good. I have reservations about what is going to happen to the binary size when we actually start adding implementations for 6 new types (we just added the types, here, no casting, no arithmetics ops of any kind). If it becomes too big of an issue... we will have to find a way. We already have a bit of an issue like that with f64 for instance anyway.

Also I have done some work around this in the play-with-spleeter branch (type mapping from onnx and tf, iirc), but that other branch is too messy and obsolete to be of much use.

PR #549, opened 2021-09-16
Expose DelayState and declare Complex Datum as copy. 

hdlj:


pulse-opl/src/delay.rs
```diff
@@ -94,22 +104,15 @@ pub struct Delay {
...
-        Delay { datum_type: input_fact.datum_type, buffer_shape, axis, delay, overlap }
-    }
-
     pub fn new_typed(
         input_fact: &TypedFact,
         axis: usize,
         delay: usize,
         overlap: usize,
-    ) -> TractResult<Delay> {
+    ) -> Delay {
```
kali, line 112: mmm... changing this return type may break a few places in NNEF and pulse, i think... not a problem, this API is considered private, but we need to fix them.

kali (approved):
ok once nnef and pulse (and whatever else gets broken acks the API change :)

pulse-opl/src/delay.rs
hdlj: Indeed, I fixed the different places where it was used. The `new` API was returning directly `Delay` without Result. I remove this API that was a duplicate with `new_typed`.

hdlj:
Thanks for the review !

PR #556, opened 2021-10-13
Depth to space op

ro99:
Adding DepthToSpace onnx op to tract.


ro99:
Executed here the `include-passing-ignored.sh`, all good 👍

kali:
Thanks a lot and congrats :)

PR #574, opened 2021-11-03
Fix missing fmt argument in error message

dtolnay:
- The first commit fills in several missing arguments to `{:?}` in tract-onnx. Without this, the error message would literally be "Neither shape not scale makes sense: input_shape: {:?}, scale: {:?}, sizes: {:?}" with curly braces in it instead of the intended variables.

- The second commit is added for future compatibility with https://rust-lang.github.io/rfcs/2795-format-args-implicit-identifiers.html &mdash; if `{size}` and `{type}` in a format string without arguments were to start trying to interpolate a variable called `size` and `type`, the original code would be problematic.


kali (approved):
Thanks!

PR #577, opened 2021-11-09
linalg build system: Fix template partials when building for ARM64 on "backslashy" platforms.

hrydgard:
Minor indentation fix while at it.


kali:
Thanks! Merging blindly here, I won't setup a CI for this.

hrydgard:
That's fine - this works.

PR #581, opened 2021-11-16
Add ArrayFeatureExtractor operator

aeveris:
Add an implementation for `ArrayFeatureExtractor` using the `Gather` operator

PR #583, opened 2021-11-16
ArrayFeatureExtractor: Fix rank check for second input

aeveris:
This MR fixes an error introduced in #581 by properly checking the second input's rank.


kali:
Ha damn, this was making it pretty unusable, right ? Are we good to go for another release ?

aeveris:
Yeah, stupid mistake. I do believe we are good to go.

kali:
Yeah, easy one to make. This is not helped by the fact the "ML" operators do not have test coverage at all in ONNX...

kali:
FYI, release done.

PR #594, opened 2021-11-30
ArrayFeatureExtractor: Use rank check from gather, don't assume indices of rank 1

aeveris:
I am terribly sorry, but I seems I previously misunderstood the operators description with regards to the possible ranks and dimensions of the indices tensor.
This MR thus fixes the rank check for the output tensor, no longer assumes the indices to be of rank 1 and correctly calculates the outputs dimensions.


kali:
Haha, no worries mate, I've been there, ONNX operators spec is not great (and -ML operators are even worse than the rest).

PR #601, opened 2021-12-22
Fix inconsistent builds by sorting dir_entries

jul-sh:
std::fs::read_dir returns entries in a platform dependent order. This variability introduces inconsistent build artifacts across platforms. This is since when assembling source files, the output of the C compiler changes depending on the order in which source files are passed.

In practice this means inconsistent builds across machines, even when building from within an identical docker container.

### DIFF
--- a/linalg/build.rs
+++ b/linalg/build.rs
@@ -153,8 +153,12 @@ fn preprocess_files(
 ) -> Vec<path::PathBuf> {
     let out_dir = path::PathBuf::from(var("OUT_DIR"));
     let mut files = vec![];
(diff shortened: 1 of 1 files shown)


kali (approved):

kali:
Thanks!

kali:
FYI, I had to revert the changes, they do look harmless, but they are not compiling. First `.collect()` can not guess the container type, then DirEntry is actually not Ord, so .sorted() fails.

jul-sh:
oh noes! Thanks for the heads up, will look into updating this.

PR #603, opened 2021-12-23
Fix inconsistent builds by sorting dir_entries

jul-sh:
Thanks for bearing with me, reopening #601 with fixes to ensure it compiles properly.

PR #607, opened 2022-01-06
Fix inconsistent build in the tensorflow crate

jul-sh:
Same as https://github.com/sonos/tract/pull/603, but in a different project. :)


kali:
Now with CI :)

Issue #634, opened 2022-02-27
BERT example


kali:
I would like to have an working BERT model example in `examples/`.

Ideally, it should come with a README that tells how to download the model and other resources with simple curl/wget commands from some stable source, without having to setup a full python environment. See https://github.com/sonos/tract/tree/main/examples/tensorflow-mobilenet-v2 for an example / template.

PR #636, opened 2022-02-28
Add BERT example (#634)

Yevgnen:
This is a small example showing how to export a [transformers](https://huggingface.co/docs/transformers/index) model to ONNX and make predictions in tract.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ members = [
     "examples/nnef-mobilenet-v2",
     "examples/onnx-mobilenet-v2",
     "examples/onnx-mobilenet-v2-batch",
+    "examples/pytorch-albert-v2",
     "examples/pytorch-resnet",
     "harness/core-proptest-pulse",
     "harness/lstm-proptest-onnx-vs-tf",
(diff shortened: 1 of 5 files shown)


kali:
Aha, tract is supposed to run with rust 1.50 (this will probably change pretty soon). I think the fix might be simple enough...

You should be able to test it with `rustup install 1.50.0` then `cargo +1.50.0 run`.

Yevgnen:
I tried 1.50.0, it worked on my Mac though...

kali:
Mmm.. this is a weird one. Look at the `tokenizers` version your `Cargo.lock` contains and see if it is different than the one the CI picked. If it is different try a `cargo update` on your mac to see if it aligns to the same version and show the same problem...

Yevgnen:
Well...after I remove all build cache and rebuild, I can reproduce the issue. It seems that I have to downgrade `tokenizers` but  some API needs to be fixed.

kali:
As I have bumped MRSV on the head to 1.59, so you may want to rebase the example, it may jsut work.

PR #648, opened 2022-03-17
tweak build-override

kali:
ping @lqd


lqd:
Support for `strip` is quite recent IIRC, maybe more so than your MSRV

kali:
my msrv is 1.59, but gha uses 1.57, and rustup update fails because setup is one as root... fix coming.

kali:
🤜

lqd:
🤛

PR #656, opened 2022-03-22
[WIP] fix: allow to load tract_core in nnef if specified

DreamerMind:
- show available registries when incorrect key is set
- add availability for tract_core operator

### DIFF
--- a/nnef/src/lib.rs
+++ b/nnef/src/lib.rs
@@ -33,5 +33,5 @@ pub mod internal {
 }
 
 pub fn nnef() -> framework::Nnef {
-    framework::Nnef::new()
+    framework::Nnef::new().with_tract_core()
 }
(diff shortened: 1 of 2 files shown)


nnef/src/lib.rs
```diff
@@ -33,5 +33,5 @@ pub mod internal {
 }
 
 pub fn nnef() -> framework::Nnef {
-    framework::Nnef::new()
+    framework::Nnef::new().with_tract_core()
```
kali, line 36: This is not desirable, we want to be able to process nnef "strictly", without allowing tract-core extensions. If you are dealing with loading a model from the command line, you can use `--nnef-tract-core` to enable the extensions.

DreamerMind: Thanks for clarification.

PR #668, opened 2022-04-11
Fix torch to nnef issues

JulienBalianSonos:
- move tract_core arc and hyperbolic sin/tan/cos to default NNEF ops following the Spec
- replace repeat agument by repeats to follow NNEF Spec


kali:
Duplicate of #665 after very eager merge and revert.

PR #681, opened 2022-04-19
Added alternative jumptable imple on 32bit arm if os==android.

### LINKED ISSUE #661: tract-linalg generates TEXTREL for armeabi-v7a (android)
Depending on the NDK revision/whether `-z,text` (>=r22) or `-z,notext` (<r22) are set this will either cause a linking failure or will generate relocations which cause problems when opening a shared object on android.

The linker failure is:

```
= note: ld: error: can't create dynamic relocation R_ARM_ABS32 against local symbol in readonly segment; recompile object files with -fPIC or pass '-Wl,-z,notext' to allow text relocations in the output
          >>> defined in /home/user/projects/armv7-problems/target/armv7-linux-androideabi/debug/deps/libtract_linalg-e3b9658eda20774e.rlib(armvfpv2_mmm_f32_4x4.o)
          >>> referenced by armvfpv2_mmm_f32_4x4.S:58 (/home/user/projects/armv7-problems/target/armv7-linux-androideabi/debug/build/tract-linalg-afe625404b8b3add/out/armvfpv2_mmm_f32_4x4.S:58)
          >>>               armvfpv2_mmm_f32_4x4.o:(.text+0x40) in archive /home/user/projects/armv7-problems/target/armv7-linux-androideabi/debug/deps/libtract_linalg-e3b9658eda20774e.rlib
```

Repeating the last lines for the addresses `.text+0x40`, `0x44`,..., `0x8c`.

This seems to be related to the dispatch/jump table.

I already tried to use some alternative patterns for implementing the  dispatch table but to no avail (through my asm skills are pretty bad).

---

There are also some cases for `arm64-v8a` where the shared object will have a textrel flag, but no actual text relocations. Through they don't appear in a minified example, so I guess they probably have a a different cause.

rustonaut:
Fixes #661, see the issue for more details.

### DIFF
--- a/linalg/arm32/armv7neon/dispatcher.tmpliq
+++ b/linalg/arm32/armv7neon/dispatcher.tmpliq
@@ -8,19 +8,31 @@
 .non_linear_loop:
     add     r0, #20
     ldm     r0, { r2, r3, r4, r5, r6 }
+
     cmp     r2, #{{ jump_table | size }}
     movgt   r2, #{{ jump_table | size }}
     cmp     r2, #0
     movlt   r2, #{{ jump_table | size }}
 
-    adr     r7, .jmp_table
-    ldr     pc, [r7, r2, lsl #2]
+{% if os == 'android' %}
+        add     pc, pc, r2, LSL#2
+        nop     // pc in Rn above is start of the add instruction + 8, hence a nop is needed
+                // This is A32 asm, for T32/Thump2 use nop.w and b.w to avoid problems.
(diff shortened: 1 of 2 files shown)


rustonaut:
I tested it locally with an armv7a android phone with neon support (using dinghy), and cross.

kali (approved):

kali:
Thanks !

superbattery:
> Thanks !

+1

PR #684, opened 2022-04-20
Feat/nnef stack unstack

JulienBalianSonos:
- add NNEF stack and unstack operator

### DIFF
--- a/nnef/src/ops/nnef/deser.rs
+++ b/nnef/src/ops/nnef/deser.rs
@@ -599,9 +599,37 @@ pub fn stack(
     }
 
     for value in &mut values {
-        // add unsqueeze before maybe casting
+        // add unsqueeze
         *value = builder.wire(ops::change_axes::AxisOp::Add(axis as usize), &[*value])?[0];
     }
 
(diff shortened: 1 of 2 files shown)


nnef/src/ops/nnef/deser.rs
```diff
@@ -575,5 +575,61 @@ pub fn leaky_relu(
...
+pub fn unstack(
+    builder: &mut ModelBuilder,
+    invocation: &ResolvedInvocation,
+) -> TractResult<TVec<OutletId>> {
+    let wire = tvec!(invocation.named_arg_as(builder, "value")?);
+    let axis: usize = invocation.named_arg_as(builder, "axis")?;
+
+    let input_fact = builder.model.outlet_fact(wire[0])?.clone();
+
+    (0..input_fact.shape[axis].clone().to_i32().unwrap())
```
kali, line 623: I think we may want to use an error throwing ("?") and not an unwrap, this may fail if the tensor dimension is symbolic. Does not cost much and safer.

kali (approved):

PR #697, opened 2022-05-02
Add support for GPU inference

sh7dm:
Address #688 .

Tasks:
- [x] `GPUTensor`
  - [x] Import
  - [x] Export
  - [x] Intermediate data in GPU memory
- [ ] Ops
  - [ ] Convolution
  - [ ] Activations
    - [x] tanh
    - [x] sigmoid
    - [ ] relu
  - [ ] Fully-connected
  - [ ] Pooling
  - [ ] Softmax
- [ ] Runner for models
  - [ ] Managing GPU memory
    - [ ] Free buffers no longer in use to allow for models larger than GPU memory
- [ ] Examples working
  - [ ] `tensorflow-mobilenet-v2`

### DIFF
--- a/gpu/Cargo.toml
+++ b/gpu/Cargo.toml
@@ -1,0 +1,43 @@
+[package]
+name = "tract-gpu"
+version = "0.16.5-pre"
+license = "MIT/Apache-2.0"
+authors = ["Mathieu Poumeyrol <kali@zoy.org>"]
+description = "Tiny, no-nonsense, self contained, TensorFlow and ONNX inference"
+repository = "https://github.com/snipsco/tract"
+keywords = ["TensorFlow", "NeuralNetworks"]
+categories = ["science"]
+autobenches = false
+edition = "2018"
+
+[badges]
+maintenance = { status = "actively-developed" }
+
+[dependencies]
+derive-new = "0.5.9"
+downcast-rs = "1.2.0"
+dyn-clone = "1.0.4"
+educe = "0.4.18"
+lazy_static = "1.4.0"
+libc = "0.2.100"
+log = "0.4.14"
+num-traits = "0.2.14"
+tract-data = { path = "../data" }
+tract-core = { path = "../core" }
+tract-hir = { path = "../hir" }
+paste = "1.0.5"
+scan_fmt = "0.2.6"
+wgpu = "0.12"
+futures = "0.3"
+bytemuck = "1.9"
+
+[build-dependencies]
+cc = "1.0.69"
+liquid = "0.24"
+unicode-normalization = "0.1.19"
+smallvec = "1.6.1"
+walkdir = "2.3.2"
+
+[dev-dependencies]
+criterion = "0.3.5"
+proptest = "1.0.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ members = [
     "onnx",
     "kaldi",
     "cli",
+    "gpu",
     "examples/tensorflow-mobilenet-v2",
     "examples/jupyter-keras-tract-tf1",
     "examples/jupyter-keras-tract-tf2",
@@ -29,6 +30,7 @@ members = [
     "harness/tf-mobilenet-v2",
     "harness/tf-moz-deepspeech",
 ]
+resolver = "2"
 
 [profile.release]
 lto = true
--- a/examples/tensorflow-mobilenet-v2/src/main.rs
+++ b/examples/tensorflow-mobilenet-v2/src/main.rs
@@ -6,10 +6,12 @@ fn main() -> TractResult<()> {
         .model_for_path("mobilenet_v2_1.4_224_frozen.pb")?
         // specify input type and shape
         .with_input_fact(0, InferenceFact::dt_shape(f32::datum_type(), tvec!(1, 224, 224, 3)))?
-        // optimize the model
-        .into_optimized()?
-        // make the model runnable and fix its inputs and outputs
-        .into_runnable()?;
+        // declutter the model
+        .into_typed()?
+        .into_decluttered()?;
+
+    // GPU state
+    println!("{:#?}", SimplePlan::new(model.clone())?);
 
     // open image, resize it and make a Tensor out of it
     let image = image::open("grace_hopper.jpg").unwrap().to_rgb8();
@@ -21,7 +23,7 @@ fn main() -> TractResult<()> {
     .into();
 
     // run the model on the input
-    let result = model.run(tvec!(image))?;
+    let result = model.into_runnable()?.run(tvec!(image))?;
 
     // find and display the max value with its index
     let best = result[0]
--- a/gpu/examples/a.rs
+++ b/gpu/examples/a.rs
@@ -1,0 +1,20 @@
+use futures::executor::block_on;
+use tract_core::prelude::{DatumType, Tensor};
+use tract_data::tvec;
+use tract_gpu::GpuAccel;
+
+fn main() {
+    let gpu = block_on(GpuAccel::default()).unwrap();
+
+    let inp = gpu.import_tensor(
+        "inp".to_string(),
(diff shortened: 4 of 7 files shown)


gpu/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+repository = "https://github.com/snipsco/tract"
+keywords = ["TensorFlow", "NeuralNetworks"]
+categories = ["science"]
+autobenches = false
+edition = "2018"
+
+[badges]
+maintenance = { status = "actively-developed" }
+
+[dependencies]
```
sh7dm, line 16: Are to be decluttered later, some might not be necessary

kali:
This is a good start. 
* I'm a bit curious about what are you plans for the operator list ? Do you plan on supporting everything in tract-core ? Or just this subset ?
* Also, wondering about he "right" way to implement sigmoid and tanh on gpu. You used the exponentiation based definition, but I found out that using a rationale development was faster on cpu. See https://github.com/sonos/tract/blob/main/linalg/src/generic/sigmoid.rs for instance.

sh7dm:
Strides patch is on its way making tanh over 1000x1000 1.5x more efficient on RADV than on llvmpipe.

1. Everything I guess, implementing one-by-one with some goals like getting MobileNet running.
2. Hmm, they're builtin. I didn't care a lot about best algos on GPU, top priority for now is managing tensors out there. Then optimizations might take place.

sh7dm:
Also wanted to ask you: do we really have 4-dimensional tensors? If not, will it be fine to crash when one encountered? If so, how to handle them gracefully when wgpu only has 3D workgroup queuing?

kali:
4D dimensional tensors are the norm, not the exception. All image categorisers (inception, mobilenet, all of them really) work in NCHW or NHWC. Most frameworks will support up to 8 dimensions.

sh7dm:
Hmm, well, if I think about it then it's obvious: n feature maps each with 3 color planes each with x and y. How to get it into 3D properly? Split the tensor?

sh7dm:
Well, a solution might be that a uniform introduced and we run multiple workgroups with different values of the parameter to work on different 3D objects from 4D list of 3D objects

sh7dm:
@haixuanTao could you please give a piece of advice or some idea for fixing that issue?

haixuanTao:
> This is a good start.
> 
>     * I'm a bit curious about what are you plans for the operator list ? Do you plan on supporting everything in tract-core ? Or just this subset ?
> 
>     * Also, wondering about he "right" way to implement sigmoid and tanh on gpu. You used the exponentiation based definition, but I found out that using a rationale development was faster on cpu. See https://github.com/sonos/tract/blob/main/linalg/src/generic/sigmoid.rs for instance.

Fyi, exponential precision is pretty bad: https://github.com/webonnx/wonnx/issues/87#issuecomment-1086916156

sh7dm:
Okay, thank you, I'll take action on that later, after basic principles

haixuanTao:
> @haixuanTao could you please give a piece of advice or some idea for fixing that issue?

In wonnx, we only use 1-D array representing the concatenation of all dimensions and reference as C-like arrays.

sh7dm:
Well, that looks like z = z*w, will do that and invent other fix if we run into z*w > 65535.

sh7dm:
The problem with this way of handling is passing W coordinate into the shader. X, Y and Z are managed by workgroup creation, but to acquire W dividing Z by W size will be necessary.

haixuanTao:
> Well, that looks like z = z×w, will do that and invent other fix if we run into z×w > 65535. Thanks for info

There is many optimization like vectorisation that we have already implemented in wonnx. I genuinely feel like you're just redoing many of the things, we already did in wonnx. 

> The problem with this way of handling is passing W coordinate into the shader. X, Y and Z are managed by workgroup creation, but to acquire W dividing Z by W size will be necessary.

In wonnx, we have all the information required when creating workgroup, so it shouldn't be a problem.

I genuinely feel like you're going to reimplement many things we already did in wonnx, and I feel like, as a community, we should probably try to find a way to merge those two projects, avoid having two different implementations, find a compromise, as the maintenance burden is quite high, especially as `wgpu` is still not that stable.

But, as I'm not part of the `Tract` project, and therefore, this decision is not mine to make.

sh7dm:
Hmm, I'm not either, but open to collaboration.

sh7dm:
About integrating wonnx into Tract (from my point of view): implementation of stuff like tensors should be consistent with Tract (what I'm working on now in this PR), but stuff like shader template compilation might be reviewed and integrated.

kali:
Well, in terms of architecture, I think the best way to go is to make a bridge between tract-core and wgpu, using work done for wonnx. I am ok with light alterations to tract to make it easier, like making public stuff which is not yet in tract-core.

For a deeper integration in tract, I will only consider it with 1/ a (quasi-)total coverage of tract-core features and operators 2/ some form of maintainer commitment as I can not maintain the gpu stuff myself on top of the rest. So it's okay to "experiment" in tract repository, obviously, it gives us a convenient discussion medium, but I will not merge short of these two constraints.

sh7dm:
> Well, in terms of architecture, I think the best way to go is to make a bridge between tract-core and wgpu, using work done for wonnx. I am ok with light alterations to tract to make it easier, like making public stuff which is not yet in tract-core.
> 
> For a deeper integration in tract, I will only consider it with 1/ a (quasi-)total coverage of tract-core features and operators 2/ some form of maintainer commitment as I can not maintain the gpu stuff myself on top of the rest. So it's okay to "experiment" in tract repository, obviously, it gives us a convenient discussion medium, but I will not merge short of these two constraints.

This. I think compiler from wonnx can be used with tract-core. Coverage should be pretty much okay. Maintenance? Well, either me (not sure if able to respond quickly enough) or @haixuanTao, if you're willing to do so.
@haixuanTao now it's your time to think about that idea, are you willing to integrate some works of wonnx into Tract and maintain it. Not sure if integration of the two projects will open up all the greatest features of both, but that looks achievable.

sh7dm:
In this plan we do some model runner which runs Tract's `SimplePlan` over my `GPUTensor`s with help of wonnx's shader compiler. Seems workable.

haixuanTao:
For me, `wonnx` is still not at a (quasi-)total coverage of tract-core features and operators. And so, integration to tract at the moment seems not desirable. 

I'm ok to maintain it. What I would propose is that, `wonnx` becomes the `Tract-GPU` extension per se ( name might or might not change ). But this will require effort from wonnx to integrate tract logic. This will not add too much load on Tract side.

sh7dm:
Yes, seems like the solution does not need a lot, if any, support from tract-core. Most stuff is already exposed, you get a decluttered plan and do accelerated ops on it.

kali:
> For me, `wonnx` is still not at a (quasi-)total coverage of tract-core features and operators. And so, integration to tract at the moment seems not desirable.

Agreed. So this bridge should probably start its existence outside of tract, as a separate entity. Just prove the concept, see if it gets traction, if it becomes mature, we can discuss integration.
 
> I'm ok to maintain it. What I would propose is that, `wonnx` becomes the `Tract-GPU` extension per se ( name might or might not change ). But this will require effort from wonnx to integrate tract logic. This will not add too much load on Tract side.

Agreed, I think most of the work happens around or within wonnx. I don't mind keeping the discussion happening here if it helps, but at this stage, that's about it.

kali:
By the way folks, I spell `tract` with a lowercase initial ;)

haixuanTao:
Yes, I think it would be the best place to have it maintained and we would be able to run CI on it as well.

sh7dm:
Should I fork wonnx tomorrow and move all the code I have to there?

haixuanTao:
I think that a PR would be great yeah! It would be awesome :)

kali:
Since you've found a new home, I'm going to close this issue ;)

Issue #703, opened 2022-05-06
Getting error while loading a model


7r3nzy:
I am trying to load a silero vad model from [here](https://github.com/snakers4/silero-vad/releases)

Code:

```rust
         onnx()
             // load the model
             .model_for_path(model_path)?
             // specify input type and shape
             .with_input_names(["input", "h0", "c0"])?
             .with_output_names(["output", "hn", "cn"])?
             .with_input_fact(0,
                              InferenceFact::dt_shape( f32::datum_type(), tvec!(1, 1))
                              )?
             .with_input_fact(
                 1,
                 InferenceFact::dt_shape(f32::datum_type(), tvec!(2, 1, 64)),
             )?
             .with_input_fact(
                 2,
                 InferenceFact::dt_shape(f32::datum_type(), tvec!(2, 1, 64)),
             )?
             // optimize the model
             .into_optimized()?
             // make the model runnable and fix its inputs and outputs
             .into_runnable()
```

Error:

```
Failed analyse for node #121 "Pad_27" Pad

 Caused by:
     0: Infering facts
     1: Applying rule GivenRule { inputs[1] }
     2: Tensor datum type error: tensor is TDim, accessed as I64
```
I am able to load the model successfully if I skip `into_optimized()?`

On a separate note, on a successfully loaded model, I am not getting same output values as I get with `utils_vad.py` from silero. I will share the repro for that in a separate issue as soon as I get the chance.

kali:
Hey, thanks for the report. I had to make a few fixes, but I got it to work (I think). Use the linked branch if you want to try, I will merge it ASAP.

All in one python inference (+ exported wav tensor for reuse in tract).

```python
import torchaudio
import onnxruntime
import numpy

session = onnxruntime.InferenceSession("files/silero_vad.onnx")

wav, sr = torchaudio.load("en.wav")
assert sr == 16000

samples = wav.shape[1]
window_size_samples = 1536

h = numpy.zeros((2, 1, 64)).astype('float32')
c = numpy.zeros((2, 1, 64)).astype('float32')

io = dict()
io["wav"] = wav.numpy().astype('float32')
numpy.savez("io.npz", **io)

output = []

for (chunk, win) in enumerate(range(0, samples, window_size_samples)):
    x = wav[:,win:win + window_size_samples]
    if x.shape[1] < window_size_samples:
        x = torch.nn.functional.pad(x, (0, 0, 0, int(window_size_samples - len(x.shape[1]))))
[...]

PR #706, opened 2022-05-07
Fixes for valero vad model

kali:


kali:
Fixes for #703 (and a bit more)

PR #709, opened 2022-05-10
Fix f32 scaling

emricksinisonos:
Not really used, but it wasn't working.
I've also moved the function computing the multiplier and the shift of a given scale in a dedicated function and changed its validity range.

### DIFF
--- a/core/src/ops/quant.rs
+++ b/core/src/ops/quant.rs
@@ -369,7 +369,7 @@ impl crate::ops::binary::BinMiniOp for Scale {
 
 #[inline]
 // This function convert a scale (actually a fraction of two integers Q/D)
-// into an integer multiplier and a shift (the multiplier being 1/D in Q1_30).
+// into an integer multiplier and a shift (the multiplier being 1/2D in Q0_31).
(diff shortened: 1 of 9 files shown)


core/src/ops/quant.rs
```diff
@@ -350,27 +350,57 @@ impl crate::ops::binary::BinMiniOp for Scale {
...
+    // Get actual value of the exponent
+    let current_exponent = scale_bits >> 23;
+
+    // Extract fractional part of the float with:
+    // - bits & 0x007fffff
+    // Replace exponent value to 126 (set actual value to 126-127=-1)
+    // - bits | 0x3f0000000
+    // We devide by two because the fractional part is in [1.0, 2.0) and we want
+    // it to be in [0.5, 1.0) in order to be in Q0_31 instead of Q1_31
+    let bumped_multi = f32::from_bits(scale_bits & 0x007fffff | 0x3f000000);
```
kali, line 393: I think we need to makr sure we needs this conversion. I'm not convinced.

kali (changes requested):

core/src/ops/quant.rs
emricksinisonos: Yes you are right ! I've modify the code to get rid of this f32 multiplication

PR #722, opened 2022-05-24
Parsing for quantized datum in CLI inputs

emricksinisonos:
Add the ability to specify a quantized datum (`QI8`, `QU8`, `QI32`) type when specifying input facts.
It can be specified with the following pattern:
- QI8 with zp scale (zp: 95, scale: 0.036406454) -> `qi8{95-0.036406454}`
- QI8 with min max (min: 0.125, max: 0.525) -> `qi8{0.125-0.525}`

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -27,6 +27,7 @@ flate2 = "1.0.20"
 lazy_static = "1.4.0"
 log = "0.4.14"
 ndarray-npy = { version = "0.8.0", features = [ "compressed_npz" ] }
+nom = "7.0"
 py_literal = "0.4.0"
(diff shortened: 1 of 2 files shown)


kali:
I think I'd prefer we find a way to align display and spec so that it can be copy-pasta easily. Coma and space are going to be interfering with the shape side of the type spec, so I propose we compromise on:
`qi8(zp:12;scale:0.012) and qi32(min:-0.012;max:0.012)`

wdyt ?

kali:
true, parenthesis and semicolumn will be semantic for the shell, so we need to use double quotes or single quotes around the typespec. But it would be the same with curly braces on other shells actually, so...

PR #723, opened 2022-05-25
Add interpolator nearest for the Resize onnx op

cchudant:
This PR implements the `nearest` mode for the Resize onnx op.

This is following the description in the ONNX docs (which isn't super clear but I think my implementation is right): https://github.com/onnx/onnx/blob/main/docs/Operators.md#Resize 

I have tested it by comparing the results of a [YoloV5 model](https://github.com/ultralytics/yolov5), using tract and onnxruntime. YoloV5 uses the `nearest` mode in the Resize ONNX op.

Do I have to add a unit test somewhere?

### DIFF
--- a/harness/onnx-test-suite/node-1.6.0.txt
+++ b/harness/onnx-test-suite/node-1.6.0.txt
@@ -390,6 +390,8 @@ test_reshape_reordered_last_dims input:data
 test_reshape_zero_and_negative_dim input:data
 test_reshape_zero_dim input:data
 test_resize_upsample_scales_linear_align_corners                                    input:X not-nnef
+test_resize_upsample_scales_nearest not-nnef not-typable
+test_resize_upsample_sizes_nearest not-nnef not-typable
 test_rnn_seq_length
 test_round
 test_scan9_sum
--- a/harness/onnx-test-suite/node-1.7.0.txt
+++ b/harness/onnx-test-suite/node-1.7.0.txt
@@ -477,6 +477,10 @@ test_reshape_reordered_last_dims input:data
 test_reshape_zero_and_negative_dim input:data
 test_reshape_zero_dim input:data
 test_resize_upsample_scales_linear_align_corners                                    input:X not-nnef
+test_resize_upsample_scales_nearest not-nnef not-typable
+test_resize_upsample_sizes_nearest not-nnef not-typable
+test_resize_upsample_sizes_nearest_floor_align_corners not-nnef not-typable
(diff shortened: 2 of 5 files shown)


kali:
Looks nice. You may check if some of the ONNX examples pass. Go to harness/onnx-test-suite and run `cargo test --ignored` and check if you've made a few new ones pass.

Then you can run include-passing-ignored.sh to add them permanently to the test manifests.

cchudant:
@kali 
Thanks, I have done that, but it seems only the node-1.7.0 tests pass, and not the later ones (see new commit)

I have compared the node-1.7.0 and node-1.8.0 onnx files a bit and i found that node-1.8.0 does not specify the `roi` and `scales` arguments for resize, whereas node-1.7.0 does (and they are linked to float32[0] tensors)
I don't really get why node-1.8.0 tests and later don't pass, i think it's a problem with the optional argument parsing (which my PR does not touch)

As for earlier versions like node-1.6.0, i'm quite confused since the onnx model seems identical to node-1.7.0 and i don't get why it is not working

Do you have any idea of why that might be the case / what are the tools i can use to debug that somehow

kali:
Could it be because of a change of opset semantics ? Resize have been defined for opset 10, 11 and 13. I have not investigated the difference. Sometimes these difference are forward compatible, in these case I just implement the latest one. But in some other case the semantics are broken so we may need two operators, or change the way we build a common operator depending on the opset (example here https://github.com/sonos/tract/blob/main/onnx/src/ops/array/slice.rs#L10)

Mapping operator set to onnx versions is here: https://github.com/onnx/onnx/blob/main/docs/Versioning.md#released-versions .

As for tools, I hope you have tried using tract cli ? There are some resources here: https://github.com/sonos/tract/blob/main/doc/cli-recipe.md . Note that if you call `tract path/to/an/onnx/onnx/backend/test/data/node/test_XXX/ run` (omitting model.onnx) it should recognize it as a such, and run it. Additionaly, you may find --pass load and --proto useful to display the result of onnx lexing alongside each op.

Hope this helps. Tell me if it does not and I'll have a closer look.

kali:
FYI there is another PR just merged in the same area: #728 .

cchudant:
Thanks for the info. I've been preparing another PR for TorchHalfPixel format on this Op, (which was removed by accident i think?) and the scale fix will probably help me a lot.

I will get back to this PR very soon :)

cchudant:
I have updated the PR. It seems #728 fixed my issue :)

kali:
OK, let's try to land this one. What are we missing here ? Can you rebase against main to get rid of the android issues ?

PR #724, opened 2022-05-25
Support of external tensors on ONNX models

JoFrost:
This PR does two things: 

- Refresh the onnx.proto3 file as onnx got several updates since it was added to tract (needed to detect and support external tensors)
- Add the support of loading external tensors from file. Onnx allows to load tensors from file in case the model is bigger than 2gb. 
The support was added without breaking the compatibility of the existing API. 

The PR has been tested against GPT Neo (1.3B and 2.7B of parameters) models.


kali:
Looks good, let's ci.

kali:
Thanks !

PR #728, opened 2022-05-31
Fix `CoordTransformer::HalfPixel` formula

SludgePhD:
I was getting incorrect results from some MediaPipe models that use `Resize` nodes (after TFLite->ONNX conversion). This PR fixes at least one of the issues.

### DIFF
--- a/onnx/src/ops/resize.rs
+++ b/onnx/src/ops/resize.rs
@@ -46,7 +46,7 @@ enum CoordTransformer {
 impl CoordTransformer {
     fn transform(&self, x_out: usize, scale: f32, len_in: usize, len_out: usize) -> f32 {
         match self {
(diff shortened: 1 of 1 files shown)


SludgePhD:
To elaborate, I've observed lower-than-expected confidence outputs of the `face_detection_full_range` model that's part of the MediaPipe [Face Detection solution](https://google.github.io/mediapipe/solutions/face_detection), and also for the `pose_detection` model that's part of MP's [Pose Estimation](https://google.github.io/mediapipe/solutions/pose.html) solution, both of which seem to be fixed by this PR.

Another issue that *isn't* fixed by this PR is lower-than-expected confidence outputs of MP's `palm_detection_lite` and `palm_detection_full` models that are part of its [Hand Tracking](https://google.github.io/mediapipe/solutions/hands) solution.

kali (approved):

kali:
Thanks!

PR #736, opened 2022-06-09
Add more kernels for x86_64

tgolsson:
This is still a WIP but wanted to open up for discussion! I also want/need to do a clean-up pass on all the kernels and and ensure that all benchmarks are valid and reproducible. 

This PR adds specialized kernels for n=2,3,4,5 to go with the existing 1,6,8 sizes. I'm a bit sad that the 1-2-3 kernels are still proportionally very slow, so might dig into that a bit more. I've also opted to dropped the AVX-512 support from this PR (apart from the matmul bodies); it became a bit too much to deal with in a single PR. 

Below is the before/after for kernel scaling, which shows some very nice improvements for small batch sizes. The bad new cases are in general prime numbers, so 7, 11, 13, 17... 

batchsize-compare

Unfortunately there was a bug in my original 2x5 kernel which lead it to overcount the performance, it performs on par with all other kernels:

```
-- 2x6 kernels
2x6x1    original                                 158% ( 7.59/ 12 cy) 54.47 GFLOP/s
2x6x2    original-unroll                          154% (15.62/ 24 cy) 52.96 GFLOP/s

-- 2x5 kernels
2x5x1    avx                                      155% ( 6.46/ 10 cy) 53.37 GFLOP/s
2x5x2    avx-unroll                               150% (13.32/ 20 cy) 51.73 GFLOP/s

-- 3x4 kernels
3x4x1    avx                                      168% ( 7.15/ 12 cy) 57.86 GFLOP/s

-- 4x3 kernels
4x3x1    avx                                      167% ( 7.17/ 12 cy) 57.66 GFLOP/s

-- 5x2 kernels
5x2x1    avx                                      156% ( 6.39/ 10 cy) 53.93 GFLOP/s

-- 8x1 kernels
8x1x1    avx                                      117% ( 6.84/  8 cy) 40.33 GFLOP/s
8x1x2    avx-unroll                                34% (47.13/ 16 cy) 11.70 GFLOP/s

-- 8x8 kernels
8x8x1    avx                                      127% ( 6.28/  8 cy) 43.92 GFLOP/s
```


linalg/src/x86_64_fma.rs
```diff
@@ -11,9 +11,78 @@ mod intel;
...
+            match best_idx.0 {
+                0 => return mmm::fma_mmm_f32_8x8::mmm(),
+                1 => return mmm::fma_mmm_f32_16x6::mmm(),
+                2 => return mmm::fma_mmm_f32_16x5::mmm(),
+                3 => return mmm::fma_mmm_f32_24x4::mmm(),
+                4 => return mmm::fma_mmm_f32_32x3::mmm(),
+                5 => return mmm::fma_mmm_f32_40x2::mmm(),
+                _ => unreachable!("not a valid index"),
+            }
+        });
```
kali, line 78: I hope this approach will be good enough on intel. On arm there were too big discrepancies between proc that we went for training a small DNN model to pick teh right model...

doc/kernel-notes.md
```diff
@@ -0,0 +1,24 @@
+# Notes about implementing and working with the kernels
```
kali, line 1: Thanks for this.

linalg/benches/x86_64.rs
```diff
@@ -0,0 +1,133 @@
...
+                ));
+            }) / 128.;
+
+
+            println!("{} {:3.0}% ({:>5.2 }/{:3 } cy) {:.2} GFLOP/s", full_label, ($n as f64 / $ww as f64) / time * 100. * *TICK, time / *TICK, $n as f64 / $ww as f64, $n as f64 / time / 1e9 );
+        }
+    }
+}
+
+unsafe fn packed_packed_2x6(f: Option<&str>) {
```
kali, line 41: I would prefer we keep the size part of the kernel name in terms of item count and not register, so please call these 16x6 instead of 2x6, etc. These ids are popping here and there upstream (in the command line dumps for instance) where it generates a lot of cognitive cost to remember the register size on multiple architectures...

kali: Did you plan on using the same top-level template for avx2/2x5 and avx512/2x5 ? I understand it would make sense in that situation to switch the naming... If we want to go this way, we will need to put an indirection/renaming at some point, I really don't want to have to remember register sizes when looking at a high level profile of a model.

linalg/src/x86_64_fma.rs
```diff
@@ -11,9 +11,78 @@ mod intel;
...
+            let n = n.unwrap();
+
+            match n {
+                1 => unreachable!("should've been mmv"),
+                2 => return mmm::fma_mmm_f32_40x2::mmm(),
+                3 => return mmm::fma_mmm_f32_32x3::mmm(),
+                4 => return mmm::fma_mmm_f32_24x4::mmm(),
+                5 => return mmm::fma_mmm_f32_16x5::mmm(),
+                6 => return mmm::fma_mmm_f32_16x6::mmm(),
+                8 => return mmm::fma_mmm_f32_8x8::mmm(),
```
kali, line 29: Mmmm... ok. so high level view of multiplier keep the old kernel naming convention, actually. So I guess you just introduced the new notation as a kernel implementation internal after all...

kali (approved):

kali:
This looks pretty good! Really happy that you find tract relevant enough to invest so much time in this.

So my undestanding is, at this stage, we have: new kernel sizes, with implementation in avx2/fma, plus new avx512 inner loops just waiting to be plugged in, right ?

linalg/benches/x86_64.rs
tgolsson: Yeah, I've gone back and forth a few times. It makes more sense from a kernel design to go with register-based sizes, but I understand that it's a higher cognitive load. I.e., when comparing AVX-512 to AVX-2 the proper comparison is 32x6 vs 16x6 to keep "layouts" same, not just using larger registers. Otherwise the load/compute ratio gets skewed.

kali:
You can disregard the android failures, those ones are on me.

linalg/src/x86_64_fma.rs
tgolsson: I think that a DNN will perform better, as my heuristics are very handwavy. But since it performs better in almost all situations; I'm not too bothered right now -- rather do that as a separate PR if we deem it necessary.

linalg/src/x86_64_fma.rs
tgolsson: Yepp! My plan was to do something smart in the build scripts to sort this out and wrote all templates on the 2x6, 2x5 etc patterns. But when dropping AVX-512 from this PR I renamed them to this scheme directly to avoid confusion.

linalg/src/x86_64_fma.rs
kali: absolutely.

tgolsson:
> This looks pretty good! Really happy that you find tract relevant enough to invest so much time in this.

We've used it for 2,5 years so I think we're quite invested mentally already, just happy to contribute back. :-) 

> So my undestanding is, at this stage, we have: new kernel sizes, with implementation in avx2/fma, plus new avx512 inner loops just waiting to be plugged in, right ?

Yes! Well; almost. The reason I dropped AVX-512 support is that "non-linear" loads/stores are much more annoying to write (and has some perf footguns e.g. `VSCATTERDPS`). My idea was to have master-kernels and have AVX-512 be an opt-in feature; and we'd then have e.g. 2x5 with either AVX or AVX-512. But then it became an utter pain of IF-ELSE and testing; so I think they'll be separate kernel templates (or relying more on fragments).  You can see some of the work here before I pivoted: https://github.com/sonos/tract/blob/7b629320f82e2288769354db26de12a8fd4a60f1/linalg/x86_64/fma/fma_mmm_f32_16x6.tmpl

The second part of the problem is how that'll affect kernel selection if both AVX and AVX-512; but that's a future problem.

kali:
> Yes! Well; almost. The reason I dropped AVX-512 support is that "non-linear" loads/stores are much more annoying to write (and has some perf footguns e.g. `VSCATTERDPS`). My idea was to have master-kernels and have AVX-512 be an opt-in feature; and we'd then have e.g. 2x5 with either AVX or AVX-512. But then it became an utter pain of IF-ELSE and testing; so I think they'll be separate kernel templates (or relying more on fragments). You can see some of the work here before I pivoted: https://github.com/sonos/tract/blob/7b629320f82e2288769354db26de12a8fd4a60f1/linalg/x86_64/fma/fma_mmm_f32_16x6.tmpl

I see. Yeah, I did not bother with scatter at the time, because everybody was more or less saying it was usually counter productive. 

About testing... Can qemu run AVX512 code ? Because I'm not sure we will get AVX512 of github actions runners. That's how I'm proceeding for arm. Testing in dev iteration or in CI with qemu, and validation and bench on device using dinghy.
 
> The second part of the problem is how that'll affect kernel selection if both AVX and AVX-512; but that's a future problem.

Will we have real life situations where AVX512 is available and AVX2 kernels perform better ? I would think that as a first approximation, always using AVX512 seems a safe bet...

kali:
FYI I have merged a fix for the android build issue on main, feel free to rebase when convenient.
Also, I wanted to make sure you noticed there is a glitch with the windows/msvc build.

tgolsson:
> About testing... Can qemu run AVX512 code ? Because I'm not sure we will get AVX512 of github actions runners. That's how I'm proceeding for arm. Testing in dev iteration or in CI with qemu, and validation and bench on device using dinghy.

No idea unfortunately! GitHub doesn't support it as they run Dv family CPUs on Azure.

> Will we have real life situations where AVX512 is available and AVX2 kernels perform better ? I would think that as a first approximation, always using AVX512 seems a safe bet...

I'm not sure about real-life so maybe it's unimportant. It can theoretically happen with a single AVX-512 port if that port also runs a lower clock. But maybe that's enough of an edge-case to disregard until someone files a bug report.

tgolsson:
> FYI I have merged a fix for the android build issue on main, feel free to rebase when convenient.
Thanks!

> Also, I wanted to make sure you noticed there is a glitch with the windows/msvc build.
Yeah I've goofed some labels with a hard dot instead of {{L}}.

tgolsson:
@kali Rebased on latest main; can you retrigger workflows?

kali:
It looks like qemu can do avx512. So we could reuse the dinghy/qemu logic adding a target for the .travis/cross.sh script...

tgolsson:
@kali I think I'm happy now, tests seem to pass (at least locally!) and it works in my benchmarks. Went ahead and deleted the AVX-512 loops for now to reduce noise -- kept the parts in the benches since I plan to continue working on AVX-512 quite soon. 

I haven't tested this in production but just running it through my benchmark suite it generated identical outputs for all batch sizes.

kali (approved):

kali:
🎉

Issue #718, opened 2022-05-17
NNEF conversion fails when reshape depends on Batch Dimension


tgolsson:
When attempting to convert a model from https://github.com/Lednik7/CLIP-ONNX I get the following error:

```
$ tract image.onnx -i N,3,224,224,f32 --nnef-tract-core --nnef-tract-onnx dump --nnef-tar image.nnef.tar
[2022-05-17T13:23:05.692388987Z ERROR tract] Translating node #104 "Reshape_97" Reshape ToTypedTranslator

    Caused by:
        0: translating op Reshape
        1: shape input is variable
```

If I fix the dimension (`-i 1,3,224,224,f32`) it works fine. Looking in the output when it succeeds, this is the offending op:

```
...
┃┣┓
┃┃┣ 29 MatMulUnary MatMul_69.split-m.0..768
┃┃┣ 30 AddUnary Add_70.Slice_81.axis2_0_768
┃┃┣ 31 Reshape Reshape_97.0
┃┃┃   ━━━ 50,84,64,F32
┃┃┣ 32 MoveAxis Transpose_98.MoveAxis-0
┃┃┃   ━━━ 84,50,64,F32
┃┃┣ 33 MulUnary Div_127
┃┣━┓
...
```

Where I specialized for N=7, and the *second* dimension of the reshape depends on the batch size. (7 * 12 = 84).

The files are huge but should be able to share them via mail/Drive if you want to investigate.

tgolsson:
This seems to occur when calling `to_typed()` on the model specifically, unrelated to NNEF conversion.

kali:
Hey, thanks for the report.

So what we encountered is a reshape from: `50,N,12,64` to `50,12N,64` ?

hint: `dump --io-long` will show all input and outputs types for all operators...

kali:
Yeah, had a look at it, I think I will need the model. You know how to reach me.

tgolsson:
Hey, shared the model with you over Drive since it's quite big. I included another model  (`text.onnx`) which also fails to convert and does similar permutes, but the error message is about inplace uniforms. The input fact used there is `-i N,77,i32`. Not sure if related, or if you want me to open a separate issue for that.

kali:
I'll have a look later this week, things are a bit hectic over there. Don't create a separate issue, it's probably the same thing, or in the same vicinity.

kali:
Made some progresses, but there are multiple issues. See branch fixes-for-718 if you're curious.

[...]

PR #741, opened 2022-06-14
support serialization for core::array::Range

tgolsson:
For the second issue from #718.  This is what the graph looks like:

image

This ends up failing with:

```
    Caused by:
        0: Serializing tensor "Range_1523_end.dat": ,TDim N
        1: Undetermined symbol in expression: N
```

I know that the [ONNX spec for Range](https://github.com/onnx/onnx/blob/main/docs/Operators.md#Range) says the start,end,step are scalars, so I'm wondering if rewriting Range to do the same would be prudent? It doesn't seem like I can easily convert the Tensor to another RValue for serialization.

### DIFF
--- a/nnef/src/ops/core.rs
+++ b/nnef/src/ops/core.rs
@@ -8,6 +8,7 @@ mod gather;
 mod one_hot;
 mod qconv;
 mod qmatmul;
+mod range;
 mod reduce;
 mod scan;
 mod scatter;
@@ -39,4 +40,5 @@ pub fn register(registry: &mut Registry) {
     scatter::register(registry);
     scan::register(registry);
     source::register(registry);
+    range::register(registry);
 }
--- a/nnef/src/ops/core/range.rs
+++ b/nnef/src/ops/core/range.rs
@@ -9,29 +9,37 @@ pub fn register(registry: &mut Registry) {
 
 fn range_parameters() -> Vec<Parameter> {
     vec![
-        TypeName::Scalar.tensor().named("start"),
-        TypeName::Scalar.tensor().named("end"),
-        TypeName::Scalar.tensor().named("step"),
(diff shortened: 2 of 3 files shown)


kali:
I think I used the Tensor for the three Range in order to be type agnostic over the actual type of three scalars. As a matter of fact the Range code https://github.com/sonos/tract/blob/main/core/src/ops/array/range.rs#L106 make enough assumptions already about these values being scalar that I think it is safe to proceed like it (we could add an ensure!() in output_fact() though).

Then we have the eternal type snafu. ONNX accepts float or ints for Range. In tract, I assume these ints can be dynamic dimensions so I make a special case for TDim late resolution. NNEF has its own quirk: it uses "Scalar" to designate the tensor payload (which are actualy float or quantized values), and "Integer" are actually shape dimensions that should usually be mapped to TDim.

So... do we need two ranges in NNEF ? One expecting TDim, one expecting floats ? I dont think so. Because if the Range is float based, it will have been resolved by the constant folding happening in decluttering. The only thing that can prevent constant folding here is the presence of a dynamic dimension.

Bottom line, we can make the assumption that the three parameters are TDim, or integer castable to TDim (one of them at least being a "true" TDim). And they are Scalar in the tract sense, single value. 

So. Parameters type in NNEF should actually be `Integer.named("...")`. We can use `cast_to_scalar::<TDim>()` to extract the three values, and use `tdim()` to serialize them inline. And no more need for external tensor storage.

tgolsson:
Thanks @kali! So if I'm following correctly: change Range to contain three TDims, and then rewrite serialization code based on that?

kali:
Don't change Range in tract_core, we need it to handle the float case. Just assume you'll find single valeus TDim in ser/deser.

kali (approved):

PR #745, opened 2022-06-19
Fix UB by creating uninit Tensors with a non-null pointer

saethlin:
For types that can be converted to a slice, it is very dangerous to
use a null pointer as an initial pointer value. Even when a slice is
length zero, the pointer passed to from_raw_parts cannot be null. The
same goes for calls to copy/copy_nonoverlapping; even if the size is 0
the pointers may not be null (memcpy has the same rule). Both situations
are encountered through the test called crop_then_down_3. This can be
detected by running the tests -Zbuild-std
--target=x86_64-unknown-linux-gnu (or whatever your host is), which will
rebuild the standard library with its debug assertions enabled, which
check for these preconditions.

### DIFF
--- a/data/src/tensor.rs
+++ b/data/src/tensor.rs
@@ -157,7 +157,7 @@ impl Tensor {
         } as *mut u8;
         let mut tensor = Tensor { strides: tvec!(), layout, dt, shape: shape.into(), data, len: 0 };
         #[cfg(debug_assertions)]
-        {
+        if !data.is_null() {
             if dt == DatumType::F32 {
                 tensor.as_slice_mut_unchecked::<f32>().iter_mut().for_each(|f| *f = std::f32::NAN)
             } else {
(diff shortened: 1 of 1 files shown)


saethlin:
Haha I ran some tests before pushing this, clearly the wrong ones.

The pointer passed to `from_raw_parts` must also be correctly aligned (so you can't use 0x1 as your new sentinel value, which some people have tried). The standard library debug assertions will catch such a mistake as well.

Yes it would be sufficient to add a branch where you are calling `from_raw_parts` and `copy_*`. _In general_ I think this is the wrong place to put the branch because conversion to a slice tends to happen all over, but I don't think tract uses `Deref` to a slice so maybe this is different.

saethlin:
The tests don't pass under ASan (`RUSTFLAGS=-Zsanitizer=address cargo test --target=x86_64-unknown-linux-gnu ...`) even without this PR. I was going to use ASan to debug these test failures, but I'm not sure that's possible.

kali:
CI-like tests can be run with "./travis/native.sh" tests if you have a good connection and a bit of space on your drive.

Slicing tensor data happens all the time, but I think there are not that many places... and most will have to be in tract-data. Checking all from_raw_parts and copy is manageable by hand. And I could also make as_ptr() unsafe on tensor... This would not solve anything, but at least it would make user aware there might be an issue.

saethlin:
I rewrote the commit to add checks instead of using a different empty representation.

But also, I think this codebase has multiple other problems. I tried running with ASan and MSan, and they fail on different tests which suggests to me that there are at least two independent bugs, a heap buffer overflow and a read of uninitialized memory. I poked around these and they're so deep in some linear algebra algorithms that I just can't follow the code enough to help you out with them.

If you can understand the bugs, it would be worthwhile to add `RUSTFLAGS=-Zsanitizer=address` at least to CI.

PR #749, opened 2022-06-22
Sanitation

kali:


kali:
@saethlin I was able to fix linalg so test from linalg and data are now passing with sanitizer=address. Next up should be "core", but I get linking error which are apparently related to asan usage.

Any advice about these ?

PR #780, opened 2022-08-04
Update compiler flags for aarch64-unknown-linux-gnu

Nehliin:
Fixes compilation issues on the target by ensuring `-mcpu=cortex-a55` is added as a compiler flag. Before this you would end up with the error posted below. Let me know if you want a different approach (perhaps unconditionally add the flag for aarch64?)

Tested on physical Ampere Altra Arm64 machine. Issue could be seen in qemu builds as well.

```
  cargo:warning=/mnt/disks/storage_disk/ark/target/debug/build/tract-linalg-de04fa2611fd2023/out/arm64fp16_sigmoid_f16_8n.S:103:5: error: instruction requires: fullfp16
  cargo:warning=    fmla v28.8h, v20.8h, v0.h[2]
  cargo:warning=    ^
  cargo:warning=/mnt/disks/storage_disk/ark/target/debug/build/tract-linalg-de04fa2611fd2023/out/arm64fp16_sigmoid_f16_8n.S:105:5: error: instruction requires: fullfp16
  cargo:warning=    fmla v24.8h, v20.8h, v28.8h
  cargo:warning=    ^
  cargo:warning=/mnt/disks/storage_disk/ark/target/debug/build/tract-linalg-de04fa2611fd2023/out/arm64fp16_sigmoid_f16_8n.S:106:5: error: instruction requires: fullfp16
  cargo:warning=    fmul v16.8h, v16.8h, v24.8h
  cargo:warning=    ^
  cargo:warning=/mnt/disks/storage_disk/ark/target/debug/build/tract-linalg-de04fa2611fd2023/out/arm64fp16_sigmoid_f16_8n.S:109:5: error: instruction requires: fullfp16
  cargo:warning=    fmla v24.8h, v20.8h, v0.h[5]
  cargo:warning=    ^
  cargo:warning=/mnt/disks/storage_disk/ark/target/debug/build/tract-linalg-de04fa2611fd2023/out/arm64fp16_sigmoid_f16_8n.S:111:5: error: instruction requires: fullfp16
  cargo:warning=    fdiv v16.8h, v16.8h, v24.8h
  cargo:warning=    ^
  cargo:warning=/mnt/disks/storage_disk/ark/target/debug/build/tract-linalg-de04fa2611fd2023/out/arm64fp16_sigmoid_f16_8n.S:112:5: error: instruction requires: fullfp16
  cargo:warning=    fadd v16.8h, v16.8h, v7.8h
  cargo:warning=    ^
  exit status: 1

  --- stderr


  error occurred: Command "cc" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "-static" "-Wall" "-Wextra" "-o" "/mnt/disks/storage_disk/ark/target/debug/build/tract-linalg-de04fa2611fd2023/out/arm64fp16_mmm_f16_128x1_a55.o" "-c" "/mnt/disks/storage_disk/ark/target/debug/build/tract-linalg-de04fa2611fd2023/out/arm64fp16_mmm_f16_128x1_a55.S" with args "cc" did not execute successfully (status code exit status: 1).
```

### DIFF
--- a/linalg/arm64/arm64fp16/arm64fp16_mmm_f16_128x1_core.tmpl
+++ b/linalg/arm64/arm64fp16/arm64fp16_mmm_f16_128x1_core.tmpl
@@ -20,7 +20,7 @@
 .text
 .align 4
 
-{% if clang == false %}
+{% if needs_pragma == false %}
 .cpu generic+fp+simd+fp16
 {% endif %}
 .global {{G}}arm64fp16_mmm_f16_128x1_{{core}}_{{suffix}}
--- a/linalg/arm64/arm64fp16/arm64fp16_mmm_f16_128x1_core.tmpl
+++ b/linalg/arm64/arm64fp16/arm64fp16_mmm_f16_128x1_core.tmpl
@@ -20,7 +20,7 @@
 .text
 .align 4
 
-{% if needs_pragma == false %}
+{% if needs_pragma == true %}
 .cpu generic+fp+simd+fp16
 {% endif %}
 .global {{G}}arm64fp16_mmm_f16_128x1_{{core}}_{{suffix}}
--- a/linalg/arm64/arm64fp16/arm64fp16_mmm_f16_16x8_core.tmpl
+++ b/linalg/arm64/arm64fp16/arm64fp16_mmm_f16_16x8_core.tmpl
@@ -22,7 +22,7 @@
 .text
 .align 4
 
-{% if clang == false %}
+{% if needs_pragma == false %}
 .cpu generic+fp+simd+fp16
 {% endif %}
 .global {{G}}arm64fp16_mmm_f16_16x8_{{core}}_{{suffix}}
--- a/linalg/arm64/arm64fp16/arm64fp16_mmm_f16_16x8_core.tmpl
+++ b/linalg/arm64/arm64fp16/arm64fp16_mmm_f16_16x8_core.tmpl
@@ -22,7 +22,7 @@
 .text
 .align 4
 
-{% if needs_pragma == false %}
+{% if needs_pragma == true %}
 .cpu generic+fp+simd+fp16
 {% endif %}
 .global {{G}}arm64fp16_mmm_f16_16x8_{{core}}_{{suffix}}
--- a/linalg/arm64/arm64fp16/arm64fp16_mmm_f16_32x4_core.tmpl
+++ b/linalg/arm64/arm64fp16/arm64fp16_mmm_f16_32x4_core.tmpl
@@ -12,7 +12,7 @@
 .text
 .align 4
 
-{% if needs_pragma == false %}
+{% if needs_pragma == true %}
 .cpu generic+fp+simd+fp16
 {% endif %}
 .global {{G}}arm64fp16_mmm_f16_32x4_{{core}}_{{suffix}}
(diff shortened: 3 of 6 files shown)


kali:
Not super convinced about the approach, tbh, the function is called "use_clang". I assume your platform defaults to clang, but a debian / ubuntu arm64 would probably pick gcc by default.

Maybe we should call "cc --version" and try to figure out my maching "clang" in there ? maybe the `cc` crate has gotten better at guessing the compiler too, it may be worth having a look.

kali:
As for putting the flag all the time, we need to make sure gcc does not choke on it.

kali:
Could you have a look / give a shot to #781 ?

Nehliin:
I'm actually running debian so gcc by default but I agree with your points of this being a bandaid solution. I was a bit confused to why `cortex-a55` was specifically selected where you instead can use `native` and do extension detection automatically. Both gcc and clang supports this (at least the versions I'm running). However I still need to pretend that my gcc compiler actually is clang thanks to the assembly templates adding [this section](https://github.com/sonos/tract/blob/main/linalg/arm64/arm64fp16/arm64fp16_mmm_f16_128x1_core.tmpl#L24) when using gcc which gives the same problem as previously. I'll push the changes that I've done for better compiler detection + using `native` unconditionally.

Nehliin:
Saw your pr now! That won't work most likely since I actually need it to pretend gcc is clang for this to work. My initial commit worked since it would unconditionally assume I'm using clang when in fact I was using gcc. I didn't actually realize this at the time but I do now that not only do I need the `-march` argument but also the template not generating the .cpu section

Nehliin:
I'm not sure why that section is added for non clang? It seems to work fine without it for gcc on aarch64 at least but perhaps there are other platform cpu combinations that need it?

kali:
Mmm... if "-march=native" works everywhere without anything more, that would be ideal... Let's see what the CI says...

kali:
OK, so that does not work. Not surprisingly, for cross-platform builds, "native" hardly make sense.

Let's just take a step back. Both the .cpu pragma and the -mcpu option were meant for the same thing: allowing translation of fp16 instructions. It does not matter if the host (aka native) supports them or not: we have runtime feature detection. I have added these (the .pragma and -mcpu) because it was not working without them, but there may be more robust / portable ways to tell the assemblers that I don't know about.

Nehliin:
Okay well if the .cpu pragma was made for the same thing it seems unnecessary at least for my configuration. Like I mentioned earlier the current state of this pr won't build on gcc because of that additional pragma (which is strange to me) but works without it. Since it works with `cortex-a55` I can revert to that if that makes cross compilation happier

Nehliin:
Sorry my mistake, I meant to say it DOES need that for both clang and gcc. (Which also makes a lot more sense to me) one sec I'll fix this and push again

Nehliin:
Welp ignore my (now deleted comment) I thought the check looked at == true not == false so my original assessment that's it's not needed for my set up was correct. However I'm not sure if other platform/compiler scenarios require it if not it could probably be removed completely

Nehliin:
What's on this branch now works for me with both gcc and clang

kali:
Ignore the two musl targets, I'm working on them on the side.

kali:
Looks good. I have merged fixes for the two musl toolchains on main. Would you mind rebasing this so we can check that the musl toolchains are consistent with this ?

kali:
All good! Thanks for your patience sorting this out :)

Nehliin:
No problem at all :)

Nehliin:
@kali When do you expect the next tract release? :) Would be nice to avoid patches but ofc don't rush anything for us, just curious!

kali:
Let's have @cchudant close on the Multinomial impl, and we'll make one. Should come quick.

PR #785, opened 2022-08-10
Add ONNX Multinomial Operator

cchudant:
Add [ONNX Multinomial Operator](https://github.com/onnx/onnx/blob/main/docs/Operators.md#Multinomial).

> Generate a tensor of samples from a multinomial distribution according to the probabilities of each of the possible outcomes.

This is useful for GPT-like models, as a source of randomness for text generation.

There are no harness tests for this operator, because the outputs are random, and even though you can seed this random on the operator ; i would guess that the ONNX runtimes are not expected to output the same results.

This makes it difficult to test this operator.
What do you think? Should I add a bare minimal CI test that only checks if the eval functions doesnt crash, without checking any output?

### DIFF
--- a/onnx/Cargo.toml
+++ b/onnx/Cargo.toml
@@ -31,3 +31,7 @@ mapr = "0.8.0"
 
 [build-dependencies]
 prost-build = "0.10.0"
+
+[features]
+default = []
+getrandom-js = ["tract-onnx-opl/getrandom-js"]
--- a/onnx/Cargo.toml
+++ b/onnx/Cargo.toml
@@ -25,7 +25,6 @@ smallvec = "1.6.1"
 tract-hir = { version = "0.17.4-pre", path = "../hir" }
 tract-nnef = { version = "0.17.4-pre", path = "../nnef" }
 tract-onnx-opl = { version = "0.17.4-pre", path = "../onnx-opl" }
-rand = "0.8.4"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 mapr = "0.8.0"
--- a/onnx/Cargo.toml
+++ b/onnx/Cargo.toml
@@ -25,6 +25,7 @@ smallvec = "1.6.1"
 tract-hir = { version = "0.17.4-pre", path = "../hir" }
 tract-nnef = { version = "0.17.4-pre", path = "../nnef" }
 tract-onnx-opl = { version = "0.17.4-pre", path = "../onnx-opl" }
+rand = "0.8.4"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 mapr = "0.8.0"
--- a/.travis/cross.sh
+++ b/.travis/cross.sh
@@ -195,7 +195,7 @@ case "$PLATFORM" in
 
     "wasm32-unknown-unknown")
         rustup target add wasm32-unknown-unknown
-        cargo check --target wasm32-unknown-unknown -p tract-onnx -p tract-tensorflow
+        cargo check --target wasm32-unknown-unknown --features getrandom-js -p tract-onnx -p tract-tensorflow
         ;;
     *)
         echo "Don't know what to do for platform: $PLATFORM"
--- a/cli/src/params.rs
+++ b/cli/src/params.rs
@@ -305,7 +305,12 @@ impl Parameters {
                 let graph = onnx.proto_model_for_read(&mut *location.read()?)?;
                 info_usage("proto model loaded", probe);
                 let path = &location.path().clone();
-                let parsed = onnx.parse(&graph, path.to_str())?;
+                let mut parsed = onnx.parse(&graph, path.to_str())?;
+
+                if matches.is_present("determinize") {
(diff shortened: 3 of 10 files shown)


kali:
Looks good... random operators are difficult to test... But there is something we should do nonetheless: allow for a way to fix the seed to have reproducible runs. I did something like that for tensorflow already: see https://github.com/sonos/tract/blob/main/cli/src/params.rs#L325.

Implem is here:
https://github.com/sonos/tract/blob/4da4cefbe436c2d078994b1a3b2ce12471e0e35c/tensorflow/src/model.rs#L75

It was done on the proto model, but I am not sure why. I think we could do with just a determinisation over the InferenceModel or TypedModel...

kali:
Mmmm and we need to clever dependence trick for webasm build...

cchudant:
Okay for determinize for cli, simple enough
I think i just need to add a feature flag for getrandom in wasm. I'll try.

cchudant:
@kali Okay, so I may have found why you did not use InferenceModel for determinize :upside_down_face: 
This is a bit cursed, but here is the problem:

Traversing the nodes from the InferenceModel, the naive way to do it would be
```rs
    pub fn determinize(model: &mut InferenceModel) -> TractResult<()> {
        use crate::ops::multinomial::Multinomial;
        for node in model.nodes_mut() {
            if let Some(op) = node.op_as_mut::<Multinomial>() {
                op.seed.get_or_insert(1)
            }
        }
        Ok(())
    }
```
This does not work, as Multinomial does not implement Op. Instead, it implements Expansion.
`Box<dyn Expansion>` implements Op though. As such, doing
```rs
node.op_as_mut::<Box<dyn Expansion>>()
```
works and gives the operator from the graph like we want. But we need to filter further, as well as downcast to our operator:
```rs
if let Some(op) = node.op_as_mut::<Box<dyn Expansion>>() {
  if let Some(op) = node.downcast_mut::<Expansion>() {
    println!("yatta!");
  }
}
```
This does not work, because `dyn Expansion` does not implement `Any`. Using `Box<dyn Expansion + Any>` would not work since the typeid would not be the same. Also, it doesn't work in stable since Any is not an auto trait.
You could fix that by making Any a supertrait of Expansion though.

Or, maybe we want to determinize on the model proto directly, as in tensorflow.

Another option would be to add a `determinize` method into Op directly, which by default would do nothing, and could be overriden by the operators.

Or, maybe determinize only after expansions are done (maybe on the typed graph or something), but that would mean that running an untyped graph with `--determinized` on the CLI would not actually return a determinized output.

kali:
Ha, thanks for going down the rabbit hole on this one. How about putting the Any as a trait bound on Expansion? Would it be enough ?

cchudant:
i think so, the code would do 2 downcasts in the end though which feels weird but that may be the simpler fix

kali:
Let's try it... there are other places where we need two nested cast...

cchudant:
i think that's it, although
- concretize is implemented in onnx, not onnx-opl, meaning nnef graphs with onnx-opl operators won't be determinized. Should i add a determinized method somewhere in onnx-opl / nnef too?
- is my wasm fix satisfactory?

onnx/Cargo.toml
```diff
@@ -31,3 +31,7 @@ mapr = "0.8.0"
 
 [build-dependencies]
 prost-build = "0.10.0"
+
+[features]
+default = []
+getrandom-js = ["tract-onnx-opl/getrandom-js"]
```
kali, line 37: https://docs.rs/getrandom/latest/getrandom/#indirect-dependencies

They suggest to put it all the time as it has no impact on non-webasm targets (that way, you don't need it on the cargo invocation too...). wdyt ?
getrandom = { version = "0.2", features = ["js"] }

kali:
And we'll leave determinisation of nnef/core graphs for later.

onnx/Cargo.toml
cchudant: i think this section of the getrandom documentation is relevant to people making an app that depends on getrandom indirectly, not for libraries?
if I understand correctly, they expect libraries to not specify any backend, and binaries to specify the backend even if they're not depending on getrandom directly.
I'm still a bit confused though, not sure I understand correctly

cchudant: i think that would mean that people using the `custom` feature in getrandom, with the `wasm32-unknown-emscripten` target will have an error when using tract because we force the `js` backend.
https://docs.rs/getrandom/latest/src/getrandom/lib.rs.html#232
https://docs.rs/getrandom/latest/getrandom/macro.register_custom_getrandom.html#registering-a-custom-getrandom-implementation

cchudant: actually, you're right
by setting the js feature we don't force it according to the getrandom sourcecode

kali: Let's ask @bminixhofer what he thinks...

bminixhofer: I haven't used getrandom, but I would also go for `getrandom = { version = "0.2", features = ["js"] }`.

From the docs:

> This feature has no effect on targets other than wasm32-unknown-unknown.

so should be all good.

PR #786, opened 2022-08-11
Make yolov5 postprocessing work

cchudant:
Hello :)

This PR is made to make the yolov5 postprocessing (which uses the If operator with a non-constant condition!).

This PR is not supposed to be merged as-is, as it removes (maybe?) important checks...
Instead, i'd like to know what you think.

I'll make some comments directly on the code using the github review interface.

### DIFF
--- a/core/src/ops/array/dyn_slice.rs
+++ b/core/src/ops/array/dyn_slice.rs
@@ -51,23 +51,29 @@ impl EvalOp for DynSlice {
     }
 
     fn eval(&self, inputs: TVec<Arc<Tensor>>) -> TractResult<TVec<Arc<Tensor>>> {
-        unsafe {
-            let start =
-                if self.start_input { inputs[1].cast_to_scalar::<i64>()? as usize } else { 0 };
-            let end = if self.end_input {
-                inputs[1 + self.start_input as usize].cast_to_scalar::<i64>()? as usize
-            } else {
-                inputs[0].shape()[self.axis]
-            };
-            if start >= end {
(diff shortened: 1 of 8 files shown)


kali:
I'm curious, what does yolo do that needs a non-constant if conditional ?

cchudant:
> I'm curious, what does yolo do that needs a non-constant if conditional ?

I think it's conditionally calling NonMaxSuppression depending on the number of boxes found in the image. This is unmodified yolov5-rt-stack, so there is probably a way to rewrite the graph so that it does not use If without constant condition..

but i kinda want that onnx models that are often used Just Work without having to resort to weird rewrites

kali:
No worries here. I fully agree that "common" networks should run out of the box in tract. Within reason of course: Tensor List and/or Sequence support would be another story as they would be a massively invasive change, but I don't think there hype anymore.

My question was more about... only one internal branch of the operator will be evaluated. This is good. But if there is heavy computation in some inputs which are not used by the "active" internal branch, then that can lead to waste. If it's too bad, we may need to make changes to the Plan, making the eval order conditional, and getting the plan one step closer to a bytecode runner.
One other way would be to detect the pattern (input used only by one of the branches) and pull these operation inside the iff.

But let's not get ahead of ourselves :) I'm happy to merge this once it is just working, we'll figure out what need to be optimised when it arises, if it arises.

0xBYTESHIFT:
what's the status of it now? does yolov5 work in tract?

kali:
Scavenged the relevant bit as part of #2155. Long time no see !

PR #799, opened 2022-09-06
Split --input-bundle into --input-facts-from-bundle and --input-from-bundle

cchudant:
Motivation #797

I did not change the CI (yet?)
Not sure this is the cleanest

I only changed the `run` command for `--input-from-bundle`. You were talking about `dump` / `compare` too but i'm not sure I understand: dump does not run the model, compare only uses random input, right?

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -88,7 +88,8 @@ fn main() -> tract_core::anyhow::Result<()> {
                   "Set input shape and type (@file.pb or @file.npz:thing.npy or 3x4xi32)."))
 
         .arg(arg!(--"const-input" [const_input] ... "Treat input as a Const (by name), retaining its value."))
-        .arg(arg!(--"input-bundle" [input_bundle] "Path to an input container (.npz)"))
+        .arg(arg!(--"input-bundle" [input_bundle] "Path to an input container (.npz). This sets input facts and tensor values.").hide(true))
+        .arg(arg!(--"input-facts-from-bundle" [input_bundle] "Path to an input container (.npz). This only sets input facts."))
         .arg(arg!(--"allow-random-input" "WIll use random generated input"))
 
         .arg(arg!(--"kaldi-adjust-final-offset" [frames] "Adjust value of final offset in network (for reproducibility)"))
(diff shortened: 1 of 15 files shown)


kali:
`dump` needs data for `--profile`, `compare` can use pre-determined input if you ask. And same for `bench` actually. See the various invocation of `retrieve_or_make`...

kali (approved):
It's not great, but good enough. The CLI code is quite far from great anyway.

kali:
Thinking about this --input-from-bundle... It may make sense to have --allow-random-input on the subcommand as well, even if it implies fixins the CI.
We could also group these two parameters decl in a function (like the display options) which we can can apply to dump, compare, run and profile.

cchudant:
I can add `--allow-random-input` & other args like it as a subcommand too yes, i'll try and do that

cchudant:
Arg --const-input seems to be unused / unimplemented
https://github.com/sonos/tract/blob/328c7acfb41ababa2086b74d554d8f1356bef490/cli/src/main.rs#L90

At least, i can't find the implementation if it exists. Can it be removed?

cchudant:
@kali I think that's it

Other arguments affected:
--allow-random-input
--allow-float-casts

Should I change the CI? The changes I made are backward-compatible.

kali:
I am surprised about const-input. It serves when running onnx test cases from the cli (which is useful for debugging mostly). As in the onnx test manifest file, we sometimes need to say that some inputs are actually const (to avoid supporting operator configuration with variable inputs that are never used).

I may have broken it while refactoring around the TensorValues struct... It will fix it if/when I need it, I guess... It's not that often.

kali:
Yes, I think it would be nice to fix the various invocation in the CI scripts in case somebody uses them as examples...

cchudant:
@kali 
I changed the CI, and fixed a problem. I think this is ready to merge!

cli/src/main.rs
```diff
@@ -99,7 +102,7 @@ fn main() -> tract_core::anyhow::Result<()> {
         .arg(arg!(--"onnx-test-data-set" [data_set] "Use onnx-test data-set as input (expect test_data_set_N dir with input_X.pb, etc. inside)"))
         .arg(arg!(--"onnx-ignore-output-shapes" "Ignore output shapes from model (workaround for pytorch export bug with mask axes)"))
 
-        .arg(arg!(--"input-node" [node] ... "Override input nodes names (auto-detects otherwise)."))
+        .arg(arg!(--"input-node" [node] ... "Oveide input nodes names (auto-detects otherwise)."))
```
kali, line 105: oupsie :)

kali (approved):

kali (approved):

PR #829, opened 2022-10-06
NNEF resource parsing

hdlj:


nnef/src/resource.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    fn into_boxed(self) -> Box<dyn ResourceLoader> where Self: Sized + 'static {
+        Box::new(self)
+    }
+}
+
+impl Resource for Document {}
+
+#[derive(Clone, Debug, Copy, PartialEq, Eq, Hash)]
+pub struct NnefDocumentLoader;
```
kali, line 46: I think we should call this guy `GraphNnefLoader`

nnef/src/resource.rs
```diff
@@ -0,0 +1,119 @@
...
+        } else {
+            Ok(None)
+        }
+    }
+}
+
+impl Resource for Tensor {}
+
+#[derive(Clone, Debug, Copy, PartialEq, Eq, Hash)]
+pub struct NnefTensorLoader;
```
kali, line 72: `DatLoader` ?

nnef/src/resource.rs
```diff
@@ -0,0 +1,119 @@
+use std::path::Path;
+
+use crate::ast::{Document, QuantFormat};
+use crate::internal::*;
+use tract_core::downcast_rs::{impl_downcast, DowncastSync};
+
+pub const NNEF_DOCUMENT_FILE: &'static str = "graph.nnef";
```
kali, line 7: If we're isolating these in constants, then I would prefer them to be GRAPH_NNEF_FILENAME and GRAPH_QUANT_FILENAME.

nnef/src/resource.rs
```diff
@@ -0,0 +1,119 @@
...
+        } else {
+            Ok(None)
+        }
+    }
+}
+
+impl Resource for HashMap<String, QuantFormat> {}
+
+#[derive(Clone, Debug, Copy, PartialEq, Eq, Hash)]
+pub struct QuantFormatLoader;
```
kali, line 97: `GraphQuantLoader` ?

nnef/src/ast.rs
```diff
@@ -8,8 +8,9 @@ pub mod quant;
 #[derive(Clone, Debug)]
 pub struct ProtoModel {
     pub doc: Document,
-    pub tensors: Vec<(String, Arc<Tensor>)>,
+    pub tensors: HashMap<String, Arc<Tensor>>,
```
kali, line 11: I hope this is not going to bite us someplace it hurts. The reason that was a Vec was (I think) to have a reproducible ordering or tensor. We'll see.

nnef/src/ops/resource.rs
```diff
@@ -0,0 +1,23 @@
...
+fn resource_get(
+    builder: &mut ModelBuilder,
+    invocation: &ResolvedInvocation,
+) -> TractResult<Value> {
+    let label: String = invocation.named_arg_as(builder, "label")?;
+    let key: String = invocation.named_arg_as(builder, "key")?;
+    let resource = builder.proto_model.resources.get(&label)
+        .with_context(|| anyhow!("No resource found for label {:?} in the model", label))?;
+    resource.get(&key)
+}
```
kali, line 23: Can we hand-craft a couple of usage examples ? as test-cases maybe ? either inline (but It may be complicated) or as integration tests (to make it easier to have external files)

nnef/src/resource.rs
hdlj: 👍

nnef/src/resource.rs
hdlj: 👍

nnef/src/resource.rs
hdlj: 👍

nnef/src/resource.rs
hdlj: 👍

nnef/src/framework.rs
```diff
@@ -185,52 +198,88 @@ impl tract_core::prelude::Framework<ProtoModel, TypedModel> for Nnef {
...
+            Arc::try_unwrap(quant)
+                .map_err(|_| anyhow!("Error while extracting quantization format resource from shared reference. Only one reference to it is expected"))
+        })
+        .transpose()?;
+
+    let proto = ProtoModel { doc, tensors, quantization, resources };
+    proto.validate()?;
+    Ok(proto)
+}
+
```
kali, line 256: good factorization here :)

kali (approved):
It looks pretty good.

nnef/src/ast.rs
kali: Yeah, I guess we will if we have to. I'm just not sure why I picked a Vec instead of a HashMap in the first place... Was it for some unit tests ? Everything seems to be green right now, so I'm not going to worry about it for now.

PR #847, opened 2022-10-27
Avoid trying to check /proc/cpuinfo when the system isn't Linux

VariantXYZ:
This is just a quick fix for systems that aren't Linux (e.g. FreeBSD), falling back on /proc/cpuinfo causes trouble.

I think for special platforms, it's probably better to have some form of override available like getrandom does, because unfortunately I'm not aware of any way to pull the target_cpu if it's been specified...

### DIFF
--- a/linalg/src/arm64.rs
+++ b/linalg/src/arm64.rs
@@ -86,10 +86,13 @@ impl Kind {
                         part
                     );
                     part
-                } else {
+                } else if cfg!(target_os = "linux") {
                     let part = max_cpuid().unwrap_or("0x00".to_string());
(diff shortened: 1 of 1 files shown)


kali:
It is possible to override the detector with environment variables... It's not documented, I thought it was only making sens in qemu test contexts where detection is confused. See https://github.com/sonos/tract/blob/main/linalg/src/arm64.rs#L61/ 

Is that good enough for your use-case on FreeBSD ? Which part of what getrandom does do you refer to ? The ability to register a generator at runtime ? What would be the equivalent for use ? Offer a programatic way to set up the cpu family ?

Issue #854, opened 2022-10-28
Aarch64 targets fall back on the GenericMmm4x4 kernel instead of arm64simd


VariantXYZ:
As per the discussion in #852 , the expectation is that tract should just rely on the ```arm64simd_mmm_f32_*_generic``` family, but in testing on a cortex-a57 device (Ubuntu 18.04, cortex-a57+simd, aarch64-unknown-linux-gnu) running the resnet example, a profile shows that it's spending a lot of time in the Generic 4x4 kernel.

I modified the KIND-selector to always return cortex-a53 out of curiosity, and found that performance improved and it did not spend time in the generic 4x4 function, so my assumption is that something's up with the kernel selection (or I just messed up when building).

VariantXYZ:
Oh, it's because the generic blob ops.mmm that's set [here](https://github.com/sonos/tract/blob/662d51af704370640d546c9002738059682b3129/linalg/src/lib.rs#L103) is never actually overridden in [plug](https://github.com/sonos/tract/blob/main/linalg/src/arm64.rs#L142).

Unless I'm missing something else.

kali:
Yeah, that's definitely a bug. I'm patching to fallback to picking a _gen kernel with simple heuristics.

kali:
@VariantXYZ want to give the branch a shot ? See what it picks on the Cortex-A57 ?

VariantXYZ:
Looks like it correctly relies on the simd kernels with that change.

kali:
I think we can close this one. Please reopen if you feel otherwise.

VariantXYZ:
@kali , I noticed this fix wasn’t in 0.18.5, not sure how that happened…

kali:
0.18.x is a maintenance branch. main will be released soon-ish (probably in january) as 0.19 with some breaking changes. This fix has only been made to main only. I can backport it to 0.18.x if you're in a rush.

VariantXYZ:
Ah, no problem. I wasn’t aware of the release methodology. I can work around it, I just thought it was an error.

PR #855, opened 2022-10-28
add simple heuristic for picking a _gen aarch64 kernel on cpu w/o model

kali:


kali:
Partially address #854.

PR #868, opened 2022-11-08
Fix facts inference on example pytorch-albert-v2

llenotre:
Hi there,

I found a bug while running the example `pytorch-albert-v2`. It gives the following error:

```
Error: Failed analyse for node #1470 "Add_1433" Add

Caused by:
    0: Infering facts
    1: Applying rule outputs[0].shape == 1,9,30000
    2: Unifying shapes b,s,30000 and 1,9,30000
    3: Impossible to unify Sym(Symbol('b', 0)) with Val(1).
```

After a short discussion with @cchudant, he told me calls to the function `with_input_fact` were not required anymore.
Removing them fixes the problem


kali:
Thanks !

PR #876, opened 2022-11-11
Fix NNEF pool operator padding

sinitame:
When trying to execute `inception_v1` from NNEF model Zoo: [here](https://sfo2.digitaloceanspaces.com/nnef-public/inception_v1.onnx.nnef.tgz). I've got the following error:

```
[2022-11-11T15:34:27.734378000Z ERROR tract] In ModelBuilder::translate

    Caused by:
        0: In ModelBuilder::translate
        1: Wiring root graph body
        2: Plugging in assignement for "concat"
        3: Resolving invocation concat
        4: Interrogating registry tract_nnef
        5: Deserializing op `concat'
        6: inputs are [202/0>, 195/0>, 181/0>, 210/0>]
        7: wiring concat.Concat (TypedConcat { axis: 1, slices: [Var, Var, Var, Var] }), determining output_facts
        8: in output_facts invocation
        9: Inconsistent concat TypedConcat { axis: 1, slices: [Var, Var, Var, Var] } inputs: [1,64,27,27,F32, 1,128,27,27,F32, 1,32,27,27,F32, 1,32,25,25,F32]
```

After looking at how the padding parameters are used to infer the output shape of the `MaxPool` node, I it seems that the first two pad values are considered instead of the one corresponding to the HW dimensions.

Example:
```
max_pool(input, size = [1, 1, 3, 3],padding = [(0, 0), (0, 0), (1, 1), (1, 1)], border = 'constant', stride = [1, 1, 1, 1], dilation = [1, 1, 1, 1])
```

The first two tuples of the `padding` parameters were considered resulting on a wrong output shape for the `MaxPool` node.

### DIFF
--- a/.travis/cli-tests.sh
+++ b/.travis/cli-tests.sh
@@ -49,13 +49,11 @@ echo
 echo $WHITE • harness/nnef-test-cases $NC
 echo
 
-export CARGO_OPTS="--no-default-features"
 for t in `find harness/nnef-test-cases -name runme.sh`
 do
     echo $WHITE$t$NC
     $t
 done
-unset CARGO_OPTS
 
 echo
 echo $WHITE • kaldi/test_cases $NC
--- a/.travis/cli-tests.sh
+++ b/.travis/cli-tests.sh
@@ -49,11 +49,13 @@ echo
 echo $WHITE • harness/nnef-test-cases $NC
 echo
 
+export CARGO_OPTS="--no-default-features"
 for t in `find harness/nnef-test-cases -name runme.sh`
 do
     echo $WHITE$t$NC
     $t
 done
+unset CARGO_OPTS
 
 echo
 echo $WHITE • kaldi/test_cases $NC
--- a/nnef/src/ops/nnef/deser.rs
+++ b/nnef/src/ops/nnef/deser.rs
@@ -347,18 +347,7 @@ pub fn conv_or_deconv(
         reshaped_bias.set_shape(&[output_channels])?;
         Some(reshaped_bias)
     } else {
-        if bias.rank() > 1 {
-            let output_channels = kernel.shape()[0];
-            ensure!(
-                output_channels == bias.len(),
-                "Bias tensor should be scalar or have one value per output channel"
-            );
-            let mut reshaped_bias = bias.into_tensor();
-            reshaped_bias.set_shape(&[output_channels])?;
-            Some(reshaped_bias)
-        } else {
-            Some(bias.into_tensor())
-        }
+        Some(bias.into_tensor())
     };
 
     let op: Box<dyn TypedOp> = if deconv {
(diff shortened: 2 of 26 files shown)


.travis/cli-tests.sh
```diff
@@ -49,11 +49,13 @@ echo
 echo $WHITE • harness/nnef-test-cases $NC
 echo
 
+export CARGO_OPTS="--no-default-features"
```
kali, line 52: I'm not too sure about this: unless I'm wrong, we are likely to speed up this (small) set of test, but then we will need to build tract again for the following test as the active feature set will be different.

kali (approved):

.travis/cli-tests.sh
sinitame: Yes that's true, I'll remove it

nnef/src/ops/nnef/deser.rs
```diff
@@ -411,14 +424,16 @@ fn pool_spec_for_pools(
...
+        PaddingSpec::Explicit(before_hw, after_hw, false)
     };
     Ok(PoolSpec::new(
         DataFormat::NCHW,
-        shape[2..].into(),
+        shape_hw,
         padding,
-        if dilation.len() > 2 { Some(dilation[2..].into()) } else { None },
-        if stride.len() > 2 { Some(stride[2..].into()) } else { None },
+        if dilation_hw.iter().all(|it| *it == 1) || dilation_hw.len() == 0 { None } else { Some(dilation_hw) },
```
kali, line 435: Are you sure about these ? dilations and strides in Rust are 1 in the "trivial" case, not 0.

sinitame: Yes I think ! If all dilations/strides are 1 then there is nothing to do so it can be `None`.

sinitame:
I have tried to align with the NNEF specification. It is not mentioned explicitly for pooling nodes but looking at Sliding window operations [here](https://registry.khronos.org/NNEF/specs/1.0/nnef-1.0.5.html#sliding-ops) it seems that pooling parameters should be of the same length as the input. I've updated the pooling NNEF serialization accordingly (some tests of the ONNX test suite were failing because of that).

I have also added new `DataFormat` to cover the 5 dimensional `shape`. This change is not directly related to the padding issue (I just wanted to make explicit the conversion between a `shape` and the spatial dimensions of that `shape`) so tell me if you want me to remove this part or not.

kali:
I don't like very much the renaming from hw_* to spatial_*, and it make the actual changes impossible to review. Can we revert it ? We'll discuss it separately if you want.

nnef/src/ops/nnef/deser.rs
```diff
@@ -387,20 +387,31 @@ pub fn conv_or_deconv(
     builder.wire(op, &[input])
 }
 
+fn get_spatial_shape(shape: &[usize]) -> TractResult<TVec<usize>> {
```
kali, line 390: This logic feels a bit brittle. If I understand this right, the data format can actually be NCHW or... just HW. Where does this come from ? Is it operator dependant ? (like Conv VS pooling ?)

What if we have a 3D-HW ? This test may make the wrong decision... Can we have a more robust test to choose the format ? Like propagate the info to pool_spec_for_pools ?

sinitame: Yes that's a good point. Instead of overcomplicating the PR, I've just add a strict check enforcing the NNEF specification.

kali (approved):

PR #881, opened 2022-11-18
Fix `x86_64-apple-darwin` cross compilation

Jake-Shadle:
This fixes an issue where targeting `x86_64-apple-darwin` would unconditionally invoke `xcrun` to get `ar`, which won't work on non-macos hosts (usually). Initially I made it respect the `AR/_<cfg>` etc environment variables, but then realized the code could just, be the default, and I'm unsure why it was explicitly assembling each file and manually adding them to the archive? This method works on a linux host at least, so if CI works on a normal macos host I think it should be fine.

### DIFF
--- a/linalg/build.rs
+++ b/linalg/build.rs
@@ -80,88 +80,53 @@ fn main() {
         "x86_64" => {
             let files = preprocess_files("x86_64/fma", &[], &suffix, false);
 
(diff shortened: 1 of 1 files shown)


kali:
TBH honest, I am not sure either why the macos code diverged from the linux code, and the CI seems happy enough...

PR #886, opened 2022-11-25
NNEF documentation

hdlj:


nnef/src/registry.rs
```diff
@@ -17,11 +17,35 @@ pub type Extension = Box<
         + Sync,
 >;
 
+#[derive(Clone)]
+pub struct PrimitiveDef {
```
kali, line 21: should if be named `PrimitiveDecl` instead of `Def` ?

hdlj: 👍

kali (approved):

Issue #916, opened 2022-12-15
DepthWise conv Inner loop f16 support 


VariantXYZ:
https://github.com/Rikorose/DeepFilterNet/pull/211#issuecomment-1353637586

Digging in a bit into why I was seeing so many f32/f16 conversions despite the A55 supporting fp16 storage and arithmetic, it seems like this is just a limitation of Rust’s f16 support.

To fully take advantage of FP16, I think avoiding these conversions is necessary… though, I’m not sure what the best solution is…

Maybe just rewriting the inner loop in assembly for f16 when the CPU says it supports f16?

kali:
Yeah, this is right, rust just does not have f16 support, so we must do it by hand. Assembly is a way, and we have infrastructure in linalg for this already, but intrinsic may work too. It would be a new "thing" in tract, so it needs a bit of thinking, but overall it may be less awful than going all the way through assembly and we have a couple of operators where intrinsic could help, at a lower cost than assembly.

I would be tempted to have them in linalg, the idea is that this is were the platform dependent code live, and linalg compiles and optimize pretty fast so it gets the dev iteration loop significantly more pleasant.

Do you already have some kind of list of the ops that should be ported to f16?

VariantXYZ:
> Do you already have some kind of list of the ops that should be ported to f16?

DeepFIlterNet2 (a very convenient example) at least is primarily bottlenecked by this function https://github.com/sonos/tract/blob/main/core/src/ops/cnn/conv/depth_wise.rs#L209 though I noticed some others. Every add/mul instruction on F16 requires a conversion to FP32.

> I would be tempted to have them in linalg, the idea is that this is were the platform dependent code live, and linalg compiles and optimize pretty fast so it gets the dev iteration loop significantly more pleasant.

To be honest, for something like this, I would suggest handling F16 in C while Rust support is lacking, and just expose a C API that gets called within Rust... but that probably isn't ideal for tract's general organization. In this case, just handwriting assembly per platform is probably fine.

VariantXYZ:
@kali ,

I tried looking into doing some of the inner loop in assembly but had a few questions:

`            for (ix, v) in iter {
                let k = *kptr.add(ix);
                let i = *iptr.offset(v);
                sum = sum + k * i;
            }
`

[...]

PR #920, opened 2022-12-24
WIP: Use external C implementation for f16 dot product

VariantXYZ:
Ping #916

### DIFF
--- a/core/src/ops/cnn/conv/depth_wise.rs
+++ b/core/src/ops/cnn/conv/depth_wise.rs
@@ -208,7 +208,11 @@ impl DepthWise {
     ) {
         let optr = optr.offset(visitor.output_offset);
         *optr = *bias.offset(c);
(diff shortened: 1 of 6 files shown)


VariantXYZ:
The build failures are a bit of a pain:

> https://clang.llvm.org/docs/LanguageExtensions.html#half-precision-floating-point

Not sure how to best handle this. Maybe just building conditionally with fp16 support is available on the target at compile-time?

kali:
About the build issues, well, we can't assume we'll always have clang.

* So, first, in build.rs, we must only try to compile this code if the c compiler is clang.
* we must inform from build.rs the main Rust code whether the clang-optimized function is available.
* we must have an alternative fallback Rust function, which will be set by default on Ops.
* finally if the clang-optimized function is available we can overwrite the default

kali:
This is great news. Keep me posted on your progress :)

VariantXYZ:
Closed the PR. I think the change in the half-crate is the correct way to go.

I'll see if I can fit it in to that mold somehow.

PR #938, opened 2023-01-15
Consider using easyfft

WalterSmuts:
`easyfft` is a wrapper that uses `rustfft` internally. It makes it harder to make mistakes and manages boilerplate for you.

```
commit edfd69987535e98e36ec5f24a768076c1eb5795c
Author: Walter Smuts <smuts.walter@gmail.com>
Date:   Sun Jan 15 08:54:20 2023 +0200

    Replace rustfft with easyfft

    This avoids re-generating the Planner struct each time the `eval_t`
    method is invoked. As a bonus it also handles initialization and sizing
    calculations for you, resulting in less boilerplate.

commit 0484590e18d269adf293f729a7c8c3da9961b931
Author: Walter Smuts <smuts.walter@gmail.com>
Date:   Sun Jan 15 08:53:03 2023 +0200

    Remove rustfft from onnx-opl module

    The module has it declared as a dependency but it's unused.
````

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,7 +78,7 @@ rand_distr = "0.4"
 readings-probe = "0.1.3"
 regex = "1.5.4"
 reqwest = { version = "0.11.4", features = [ "blocking", "rustls" ], default-features = false }
-rustfft = { version = "6.1", features = [ "neon" ] }
+easyfft = "0.3.5"
 rustls = "0.20.4"
 smallvec = "1.6.1"
 scan_fmt = "0.2.6"
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -27,7 +27,7 @@ maplit.workspace = true
 ndarray.workspace = true
 num-integer.workspace = true
 num-traits.workspace = true
-rustfft.workspace = true
+easyfft.workspace = true
 smallvec.workspace = true
 tract-linalg = { version = "=0.19.1-pre", path = "../linalg" }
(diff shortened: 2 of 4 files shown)


kali:
Sorry, I don't think I'm going to merge this.

First, it's loosing me the neon optimisations on arm64.
Secondly it adds a dependency and some opacity for a marginal benefit. tract infrastructure is already equipped to deal with the planner and scratch space by using its op state. Also, there is no scaling benefit, there should never be another place where FFT is needed in tract, so ergonomics matter very little.

Nice work on improving rustfft ergonomics, though. I'm sure a lot of project will appreciate it. I just don't think this is for tract.

WalterSmuts:
> First, it's loosing me the neon optimisations on arm64.

What makes you think so? `easyfft` also has the neon optimisations... It's a default feature in `rustfft`.


> First, it's loosing me the neon optimisations on arm64.
Secondly it adds a dependency and some opacity for a marginal benefit. tract infrastructure is already equipped to deal with the planner and scratch space by using its op state

It's quite a large performance improvement to the code you currently have. Currently you allocate a new planner each time `eval_t` is called and you allocate a new scratch buffer *for each iteration of the loop* in the `eval_t` functions. Sure you *could* do all the handling of the planner and scratch buffers yourself, but that's adding extra overhead and does exactly what `easyfft` does. So IMO the negative effects that adding an extra dependency has would also be added when you implement it yourself, except now *you* have to maintain it.

PR #949, opened 2023-01-24
nnef: ignore directories when loading a model from a path

lqd:
`File::open`ing a directory can fail on an unnamed OS.

With this, a `cargo test`  in `nnef` completes successfully, whereas the `nnef::test_json_resource::load_model_with_json_resource` test currently fails on the `main` branch.

Building the python wheels with https://github.com/sonos/tract/pull/948 also succeeds locally. (After installing `cibuildwheel` locally, `python -m cibuildwheel ffi/py --output-dir wheelhouse --config-file "" --only "" --platform windows` can be used without `pipx` and so on)


kali:
🚙

PR #959, opened 2023-02-07
remove "real" test

kali:


kali:
Broken by https://github.com/onnx/onnx/issues/4857

PR #986, opened 2023-03-02
Allow bool values to be resolved by their identifier

emricksinisonos:


linalg/src/frame/mmm/tests.rs
```diff
@@ -7,403 +7,409 @@ use tract_data::internal::*;
```
kali, line 7: what happened here ? reformatting ?

kali (changes requested):

linalg/src/frame/mmm/tests.rs
emricksinisonos: Yep 💀

PR #989, opened 2023-03-03
Avx 512f

cchudant:
Hello, this is initially the work of @tgolsson, and @feldspath took over and implemented some kernels. My only contribution in this is the kernel selection.

Microkernel benchmarks:

image

Some notes:
- this one was a bit surprising to me: most kernels are not that fast, i don't know why. Only 3x4 / 4x3 have a good throughput
- unroll on k is slower in all cases
- reordering instructions in the microkernels has no effects on perf
- using more register / reusing registers also has generally no effect on perf

Now, for kernel selection:
image

Some kernels are never optimal, regardless of the value of N
After removing them from the graph, it looks like this
image
64x3 is almost always better, except when N is two, and sometimes (when N divisible by 3 but not 4) 48x4 is better by a tiny bit

## future work

lots of assembly here is duplicated from between x86_64/fma and x86_64/avx512 folders; tidying this up would be nice


kali (approved):

tgolsson:
@cchudant and @feldspath Awesome work! 

The conclusions mirror my expectations where I left off the work - AVX512 holds a lot of promise but it's finicky with L1/L2 caching, double-pump ports and downclocking. I'd love to repro and see what happens on my machine, in case it's different than your results - are the scripts to generate those graphs checked in? I'll also see if I can find the time to test his on our production workloads and see what happens there.

Out of curiosity, what CPU have you tested on?

kali:
@cchudant do you need help with the rebasing conflicts ?

cchudant:
@tgolsson 
> are the scripts to generate those graphs checked in?

Kind of, i have this script <https://github.com/sonos/tract/blob/a80d91e3cc1ac4f2ab7fcd7c6a1619520d0b72c3/linalg/benches/intel.rs#L148> that outputs a CSV and I just load it on google sheets (sorry)

> Out of curiosity, what CPU have you tested on?

Intel(R) Xeon(R) Gold 6334 CPU @ 3.60GHz

kali:
Do we have any concern about platforms where these changes can be counter-productive? I'm inclined to merge this relatively aggressively, it feels like a very valuable contribution and I don't want to let it on the side again.

Picking the best kernel in all situations is a super hard problem. We did a relatively expensive attempt at fixing it with a DNN on the aarch64 side of things, which is super critical for us. I would prefer we avoid this trajectory for avx-512 if we can (mostly because I would not be in a position to maintain it).

As you noticed, the variables at play in the intel case are relatively different, so I'm happy with geometry-based heuristics as long as we are not too concerned about them doing anything very bad on a case that actually matter to people around the table here ;)

tgolsson:
@kali 

 I think we talked about that a long while back; but in theory - yes. Maybe that was in the discussion thread. On single-port AVX512f or ones that aggressively downclock with AVX512 instructions I think it could be negative. But I don't have hardware to verify that this would happen, and well; it's not that common to not have double AVX-512 VPUs nowadays.

https://github.com/jeffhammond/vpu-count shows that most recent generations have double ports, at least.

@cchudant Interesting, that is very similar to what I have at least. I'm on an older but beefier W-2295, primarily differing by amount of cache in practice. That was something I noticed had a big impact on the AVX256 kernels; most kernels are optimal on small MMs but for my production workloads (1024x1024 weights) they're all memory bound. Will try to benchmark over the weekend or early next week - I'm out of disk space in WSL so can't even build tract right now 😂

kali:
@tgolsson Ha right, sorry, I remember now. Damn, Intel is... mmm... complicated.

What about only activating these kernels on double-port cpus and leaving the single-ports using the FMA ones ? Is there a flag that we could use ? or some other cpuid trickery ?

Also, if we are only concerned about a hypothetical issue... if you don't have any single-port cpu you can test, and if @cchudant does not have any either, I'm still ok to merge and deal with the consequences later.

tgolsson:
I don't know of any CPUID tricks; Intel themselves provide this heuristic:

https://github.com/intel/optimization-manual/blob/main/chap18/ex25/fma_unit_count.c

cchudant:
> Also, if we are only concerned about a hypothetical issue... if you don't have any single-port cpu you can test, and if @cchudant does not have any either, I'm still ok to merge and deal with the consequences later.

I don't have one sorry :/

> I don't know of any CPUID tricks; Intel themselves provide this heuristic:
> 
> https://github.com/intel/optimization-manual/blob/main/chap18/ex25/fma_unit_count.c

That's interesting, I guess it could be done. Do you want me to do this @kali?

I fixed the clippy things, I think this should pass CI now

kali:
Naaa, too complicated to do before we know for sure we actually have a problem. Plus we know we have a workaround if it comes to that. Let's merge this!

kali:
🚀

PR #992, opened 2023-03-06
Submodel support

emricksinisonos:
Introduce a new operator called `tract_core_submodel` (part of `tract_core` registry) that allows the definition of submodels in a NNEF graph.
```
fragment tract_core_submodel(
    input: tensor<scalar>,
    label: string )   # Name of the submodel in the resources (can be a directory: `nnet1` or an archive: `nnet1.nnef.tgz`, it must end by `nnef.tgz` in case of an archive)
-> ( output: tensor<scalar> )
```

- [x] Add submodel operator
- [x] Implement deserialization (from archive + from directories)
- [x] Implement serialization (to archive)
- [x] Generalise operator to support different kind of submodels
      - Abstract inner submodel with a trait
      - Add possibility to have alternative deserialiser (submodel loading can be based on external loading strategy)
- [x] Be agnostic of the profiling method for submodels 

New arg added to the CLI (`--condensed`) that allows to display the graph without sub-models


nnef/src/framework.rs
```diff
@@ -80,11 +82,17 @@ impl Nnef {
         self.write_to_tar(model, w)?;
         Ok(())
     }
-
+
```
kali, line 85: I think we should try to make it possible to give more control about the nesting of compression and networks: decompressing at loading time is not negligible, and compressing neural networks is not super effective in the first place. And here we are even inducing double compression...

PR #1008, opened 2023-03-21
vectorized activation vm

kali:
This is a side-project I'm working on during live-coding sessions on twitch. Primary time for streams is Monday evenings (8:30 CET or so) on https://www.twitch.tv/kalicoding . The unedited videos are also  https://www.youtube.com/playlist?list=PL2L_7bZXXxQuNpcxjg8gipegP32pmwVZF .

Landing page for twitch with some explanations is here. https://github.com/kali/coding .

* [x] Activation function inventory and Rust POC (parts 1 and 2)
* [x] Early benchs (part 3)
* [x] Bringing up assembly integration in tract (parts 4, 5, 6)
* [x] Rework the protocol so constant are in the flow of instructions (Rust, assembly) (part 7)
* [x] Jump table generation (part 8)
* [x] Support for "bloc-1" activation (ReLu to HardSwish) (part 8, part 9)
* [x] Test should be one-liner (Rust) (part 8)
* [x] bench (part 10)
* [ ] integrate in tract-core
* [ ] ARMv8.2 f16 support
* [ ] other activations: Sigmoid (part 10), Tanh, Exp2f
* [ ] implement on ARMv7
* [ ] implement on intel (primary target AVX2+FMA)
* [ ] consider compiling instead of interpreting

### DIFF
--- a/linalg/Cargo.toml
+++ b/linalg/Cargo.toml
@@ -97,5 +97,10 @@ name = "x86_64"
 harness = false
 
 [[bench]]
+bench = false
 name = "intel"
+harness = false
+
+[[bench]]
+name = "activations"
 harness = false
--- a/linalg/activations/Cargo.toml
+++ b/linalg/activations/Cargo.toml
@@ -9,4 +9,12 @@ members = []
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dev-dependencies]
+criterion = "0.4.0"
 proptest = "1.1.0"
+
+[dependencies]
+
+
+[[bench]]
+name = "vm"
+harness = false
--- a/linalg/activations/Cargo.toml
+++ b/linalg/activations/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "activations"
+version = "0.1.0"
+edition = "2021"
+
+[workspace]
+members = []
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dev-dependencies]
+proptest = "1.1.0"
--- a/linalg/activations/benches/vm.rs
+++ b/linalg/activations/benches/vm.rs
@@ -27,6 +27,15 @@ fn crit(c: &mut Criterion, name: &str, r: impl Fn(f32) -> f32, prog: &Program) {
                 BatchSize::LargeInput,
             )
         });
+        group.bench_with_input(BenchmarkId::new("VMVec", size), size, |b, size| {
+            b.iter_batched(
+                || vec![1.0f32; *size as usize],
+                |mut v| {
+                    prog.compute_slice(black_box(&mut v));
+                },
+                BatchSize::LargeInput,
+            )
+        });
     }
 }
 
--- a/linalg/activations/benches/vm.rs
+++ b/linalg/activations/benches/vm.rs
@@ -1,8 +1,8 @@
-use activations::{definitions, reference};
-use criterion::{black_box, criterion_group, criterion_main, BenchmarkId, Criterion, BatchSize};
+use activations::{definitions, reference, Program};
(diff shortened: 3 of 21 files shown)


kali:
* We are interested in Element wise operations:
  * result is independent of tensor geometry (only accept scalar operations, no broadcasting of tensors)
  * output = f(x) for x in input 

* "simple ops + if/then/else"

  * Relu(x) = max(0, x)
  * Affine = alpha * x + beta
  * LeakyRelu = if x >= 0 { x } else { alpha * x }
  * ThresholdRelu = if x >= alpha { x } else { 0 }
  * HardSigmoid = min(max(alpha * x + beta), 0, 1)
  * Sofsign = x / (1 + abs(x))
  * HashSwish = x * max(0(min(1, alpha * x + beta))) (alpha = 1/6 and beta 1/2)
 
## Proposed virtual machine def

 * 4 big registers (mapped to one or several hardware vector registers)
   * armv8, maybe one big register may be 4 NEON registers (4*4 f32 values or 4*8 f16 values)
   * 16 NEON registers are used, 16 for housekeeping (pre-fetch caching + operators)
 * focusing on "simple" batch
   * BigRegs are a,b,c,d
   * min, max, abs, +, *, /, ite, ifpos
   * constants
 * calling convention: framework preloads x into a, expects y in a

## Moves, load consts to any reg, unary, binary, ternary on a,b,c only

 * Relu: load(b, 0) | max
 * Affine: load(b, alpha) | mul | load(v, beta) | add
 * LeakyRelu: b <- a | load(a, alpha) | mul | c <- a | ifpos
 * ThresholdRelu: c <- a | load(b, alpha) | sub | b <- c | load(c, 0) | ifpos
 * Softsign: c <- a | abs | load(b, 1) | add | recip | b <- c | mul
 * HardSwish: load(b, alpha) | mul | load(b, beta) | add | load(b, 1) | min | load(b, 0) | max

## With a <- a#K (K constant)

 * Relu: max(0)
 * Affine: mul(alpha) | add(beta)
 * LeakyRelu: b <- a | mul(alpha) | c <- a | ifpos
 * ThresholdRelu: b <- a | sub(alpha) | load(c, 0) | ifpos
 * Softsign: b <- a | abs | add(1) | recip | mul
 * HardSwish: mul(alpha) | add(beta) | min(0) | max(1)

# combination of operands
   * all moves 
      * 16 moves ops (any to any)
      * other ops have all fixed registers: unary ops a <- a, binary a <- a#b, ternary a <- a#b#c
   * all operands can be const or any registers
      * with 4 big resisters, if a then b else c needs 64 combination
   * ifte (a <- cond in a, then in b, else in c)

* Complicated activations functions
  * computed from exp, log, tang, etc as seperate operators

  * ScaledTanh = alpha * tanh(beta * x)
  * Softplus = log(1 + e^x)
  * Celu = max(0,x) + min(0,alpha * exp(x / alpha) - 1)
  * Elu = if x < 0 { alpha * (exp(x) - 1) } else { x }
  * Selu = if x < 0 { gamma * (alpha * e^x - alpha) } else { gamma * x }
  * Mish(x) = x * tanh(softplus(x)) = x * tanh(ln(1 * e^x))

## Current state of tract implementation in assembly on armv8, armv7, intel fma

    Tanh(x) - (1 - e^{-2x})/(1 + e^{-2x})
    Sigmoid(x) - 1/(1 + e^{-x})
    Erf(x)
    * rational approximation:
      P(x)/Q(x) where P and Q are polys
    * Do we want to support rational function evaluation ?

## Can we make a faster exp ? what about log ?
  *  e^x <-

kali:
Parking this. The interpreter design works but on Zen 5 + AVX-512 a
  handcrafted sigmoid runs ~4× faster than the VM doing the same math, and
  LLVM autovec beats the VM on relu/hardswish by 2-4×. The bytecode is a
  fine portable representation; the interpreter that runs it isn't the
  right shape. If we come back to this it'll be as a JIT or build-time
  monomorphizer. Branches archived under kali/coding:

PR #1024, opened 2023-03-30
Ensure callee saved registered are properly restored

emricksinisonos:
According to ARM documentation:
- x19-x29 needs to be callee preserved
- v8-v15 needs to be callee preserved

I might have been a bit too exhaustive (some registers might not be altered during the subroutine processing but I took the choice to persist all the resisters that needs to be persisted anyway in that first version of the PR). For sure it's a tradeoff between minimising load/stores instructions from the stack and ensuring the code is safe without having to look into the subroutine code.

Let's discuss that here


emricksinisonos:
I made it a bit less exhaustive by keeping the old logic for regular registers (x19, x28, x29) are not in use.

kali (approved):

Issue #1029, opened 2023-04-03
Wrong number of outputs | Error on If operator


7r3nzy:
I am trying to load [silero-vad-v4 onnx](https://github.com/snakers4/silero-vad/archive/refs/tags/v4.0.tar.gz) but it is giving me the following error:

```
Error Failed analyse for node #74 "If_25" If

Caused by:
    Wrong nnumber of outputs. Op says 1, node says 3.
```

This is how I am loading it:

```rust
        onnx()
            .model_for_path(model_path)?
            .with_input_names(["input", "sr", "h", "c"])?
            .with_input_fact(
                0,
                InferenceFact::dt_shape(f32::datum_type(), tvec!(1, voice_detection_frame_size)),
            )?
            .with_input_fact(1, InferenceFact::default())?
            .with_input_fact(
                2,
                InferenceFact::dt_shape(f32::datum_type(), tvec!(2, 1, 64)),
            )?
            .with_input_fact(
                3,
                InferenceFact::dt_shape(f32::datum_type(), tvec!(2, 1, 64)),
            )?
            .with_output_names(["output", "hn", "cn"])?
            .with_output_fact(
                0,
                InferenceFact::dt_shape(f32::datum_type(), tvec!(1, 1)),
            )?
            .with_output_fact(
                1,
                InferenceFact::dt_shape(f32::datum_type(), tvec!(2, 1, 64)),
            )?
            .with_output_fact(
                2,
                InferenceFact::dt_shape(f32::datum_type(), tvec!(2, 1, 64)),
            )?
            .into_optimized()?
            .into_runnable()
```

model overview

![2023-04-03_10-00](https://user-images.githubusercontent.com/97305194/229415672-34b66e03-ac35-408a-bdc3-f3cf557e416a.png)

`If_25` node is the one with 3 outputs

tract version: `0.19.8`

**Update:**

Running through command line has the same behavior:

```
tract silero-vad/files/silero_vad.onnx
┏ 1 Source sr
┃   ━━━ I64
┣┻ 73 Equals Equal_23
┃   ━━━ Bool
┃┏ 3 Source c
┃┃   ━━━ 2,batch,64,F32
┃┃┏ 2 Source h
┃┃┃   ━━━ 2,batch,64,F32
┃┃┃┏ 0 Source input
┃┃┃┃   ━━━ batch,sequence,F32
[...]

PR #1038, opened 2023-04-10
Improve if op

kali:
ONNX If only works has a loading macro expansion: the condition must be known and fixed at loading time, the right branch is substituted in the type phase. This blocks Silero VAD v4 (see #1029).

* [x] make analyse work when input is dynamic
* [x] implement an If operator in core and translate ONNX op into it
* [ ] find out how to encode this in NNEF
* [ ] into_optimized as in Scan
* [ ] declutter as in Scan


kali:
Moving mid-term tracking to #1062. This PR is merged to land in 0.20 with partial support (analysis and basic core support).

PR #1073, opened 2023-05-02
Introduce Load/Store/ForceEval ops

emricksinisonos:
### Load
Can load a tensor from the `SessionState` during the graph execution. The tensor in the context can be retrieved by its ID. If the tensor is not found, the default value provided as input will be used.
```
load(default_value, id = "string-id");
```

### Store
Can store a tensor to the `SessionState` during the graph execution with a given ID. An input is provided and directly propagated as output to ensure the node has an output.
```
out = store(input = in, state = state_tensor, id = "string-id");
```

### ForceEval
Force the evaluation of given nodes even if they are not producing any useful outputs. This node will forward inputs that are not indicated in the slot list and won't do anything with the other inputs (these inputs are the outputs of previous nodes that need to be force eval).
```
useful_output = force_eval([useful_input, dummy_1, dummy_2], slots = [1, 2]);
```


kali (changes requested):
* Can we group them three in a module ? Maybe called "vars" ?
* We will also need to consider how to react to axis changes requests. ForceEval is pretty straightforward, but load and store must be in agreement wrt to changes requests.
(* I am ok merging this PR and opening a separate issue for axis optimisation.)

kali (approved):

nnef/src/ops/core/force_eval.rs
```diff
@@ -0,0 +1,37 @@
...
+use crate::{
+    internal::*,
+    ser::{array, ints},
+};
+use tract_core::ops::force_eval::ForceEval;
+
+pub fn register(registry: &mut Registry) {
+    registry.register_dumper(TypeId::of::<ForceEval>(), ser_force_eval);
+    registry.register_primitive(
+        "force_eval",
```
kali, line 12: we need to call them tract_core_force_eval, and so on and so forth

Issue #1066, opened 2023-04-27
ERROR : BUILDING MOBILENET 


KhelifaSaifEddine:
i have main.rs like this : 
```rust
use tract_tensorflow::prelude::*;
use std::env;


fn main() -> TractResult<()> {

    let model = tract_tensorflow::tensorflow()
        // load the model
         .model_for_path("mobile_net.pb")?
	.with_input_fact(1, i32::fact([ 1,900, 1352, 3]).into())?

        // make the model runnable and fix its inputs and outputs
         .into_optimized()?
        // make the model runnable and fix its inputs and outputs
        .into_runnable()?;






    Ok(())

}
``` 

Caused by:
    ShapeError/OutOfBounds: out of bounds indexing', examples/tensorflow-mobilenet-v2/src/main.rs:8:52
stack backtrace:
   0: rust_begin_unwind
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/std/src/panicking.rs:579:5
   1: core::panicking::panic_fmt
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/core/src/panicking.rs:64:14
   2: core::result::unwrap_failed
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/core/src/result.rs:1750:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/core/src/result.rs:1090:23
   4: example_tensorflow_mobilenet_v2::main
             at ./src/main.rs:8:17
   5: core::ops::function::FnOnce::call_once
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

Anyone can help please ??

kali:
This looks like a protobuf parsing error. Check that your .pb file is indeed a protobul model, in binary form not text form (i think the library does not like them).

Upload the file if you need more help.

KhelifaSaifEddine:
Here is the file kali and i think its protofbuf model (i trained and froze it myself) i can even provide google colab or jupter notbook  you can find file here 
https://drive.google.com/file/d/1_LKs2RrpsqhVCExFAG6qE3zrR-BjrpNd/view?usp=share_link

[...]

PR #1074, opened 2023-05-05
Fix Runtime Error of the ONNX Example due to `memmap2` on WebAssembly Platform

fgh1999:
Fix Issue https://github.com/sonos/tract/issues/1066#issuecomment-1528933145.


kali (approved):

kali:
Look pretty good, thanks for taking the time. One more tiny bit that we could do to improve reliability would be to add the wasm commands to the ci.sh script in the example (so I'll know if I break wasm again).

fgh1999:
Okay. Wasm runtime added in ci.sh.

kali:
Awesome :) Thanks a lot !

PR #1131, opened 2023-07-31
hold back time crate

kali:


poliorcetics:
@kali was this related to the project's MSRV ?

kali:
@poliorcetics yes

Dirreke:
Could we use `time = "0.3.23"` and `time-macros = "0.2.10" ?

Is there any compatibility issues? If not, I think the above is better. Those who use `rust 1.65`  can use `time 0.3.23`, while others can use newer `time` to ensure compatibility with other deps .

kali:
Well, I have no notes or specific memories, but you can assume I put the hold in place because the build was broken. I guess time metadata and cargo version dependencies were conflicting at that point and I have not checked again since. I will increase the MSRV of tract at some (undetermined) point in the future, and at that point I will release all the crate holds, so this is not forever.

What problem are you trying to solve and how bad is it ? If you're convinced your solution works, feel free to make a PR and we can see if the CI is happy with it.

PR #1296, opened 2024-01-07
 fix: set constant inner tensor values upon "setting" variables 

### LINKED ISSUE #1279: Setting variables does not set values that depend on those variables
Have a simple mean calculation model which when loaded creates a fairly simple table of operations. We also forcefully set the batch_size to be 1 and concretize those dimensions as follows: 

```rust


        let mut model = model.into_typed()?.into_decluttered()?;
        let mut symbol_values = SymbolValues::default();
        for (symbol, value) in run_args.variables.iter() {
            let symbol = model.symbol_table.sym(symbol);
            symbol_values = symbol_values.with(&symbol, *value as i64);
            info!("set {} to {}", symbol, value);
        }
        model = model.concretize_dims(&symbol_values)?;

```

[network.onnx.zip](https://github.com/sonos/tract/files/13568998/network.onnx.zip)

The pretty printed version of this model is as such however

```bash 
model     0 |                   -> >1/0              | Source                    input                                              [] => [1,1,300,F32]
    1 | 0/0>              -> >5/0              | Reduce<Sum>               /ReduceMean.sum                                    [1,1,300,F32] => [1,1,1,F32]
    2 |                   -> >3/0              | Const                     /ReduceMean.size                                   [] => [1,1,1,TDim 300*batch_size]
    3 | 2/0>              -> >4/0              | Cast                      /ReduceMean.cast                                   [1,1,1,TDim 300*batch_size] => [1,1,1,F32]
    4 | 3/0>              -> >5/1              | Recip                     /ReduceMean.norm-recip                             [1,1,1,F32] => [1,1,1,F32]
    5 | 1/0>     4/0>     -> >6/0              | Mul                       /ReduceMean.norm                                   [1,1,1,F32, 1,1,1,F32] => [1,1,1,F32]
    6 | 5/0>              -> >7/0              | RmAxis                    /ReduceMean-dispose-dims-2                         [1,1,1,F32] => [1,1,F32]
    7 | 6/0>              -> >8/0              | RmAxis                    /ReduceMean-dispose-dims-1                         [1,1,F32] => [1,F32]
    8 | 7/0>              ->                   | RmAxis                    /ReduceMean-dispose-dims-0                         [1,F32] => [F32]
outputs: 8/0>



```
The batch size has been correctly set but as you can see the values `300*batch_size` persist within the tensors themselves (scroll the codeblock to the right). 

I'm not sure this is expected behaviour -- one would expect the symbolic values to also be filled within the Const nodes -- unless there's a separate command for that ?

alexander-camuto:
Resolves #1279 on the cli by creating a routine within the --set feature of the tract cli which: 

1. Iterates over all constants in the graph 
2. If const's the inner tensor type is of type TDim then: 
    a) extract that tensor
    b) _evaluate_ the inner values using the symbol_values map previously established
3. _then_ concretize_dims for the entire graph 

Have confirmed that this update works on the network file provided in #1279

### DIFF
--- a/cli/src/params.rs
+++ b/cli/src/params.rs
@@ -736,7 +736,31 @@ impl Parameters {
                 let key = typed_model.as_ref().unwrap().get_or_intern_symbol(key);
                 values.set(&key, value);
             }
-            stage!("set", typed_model -> typed_model, |m: TypedModel| {
+            stage!("set", typed_model -> typed_model, |mut m: TypedModel| {
+                for node in m.eval_order()? {
+                    let node = m.node_mut(node);
+                    if node.op_is::<Const>() {
+                        // map option to err
+                        let op = node
+                            .op_as_mut::<Const>()
+                            .unwrap();
+                        // get inner value to Arc<Tensor>
+                        let mut constant = op.0.as_ref().clone();
+
+                        match constant.datum_type() {
+                            DatumType::TDim => {
+                                // Generally a shape or hyperparam
+                                constant
+                                    .as_slice_mut::<tract_onnx::prelude::TDim>()?
+                                    .iter_mut()
+                                    .for_each(|x| *x = x.eval(&values));
+
+                                op.0 = constant.into_arc_tensor();
+                            }
+                            _ => {}
+                        }
+                    }
+                }
                 m.concretize_dims(&values)
             });
             stage!("set-declutter", typed_model -> typed_model, TypedModel::into_decluttered);
--- a/cli/src/params.rs
+++ b/cli/src/params.rs
@@ -740,8 +740,7 @@ impl Parameters {
                 for node in m.eval_order()? {
                     let node = m.node_mut(node);
                     if let Some(op) = node.op_as_mut::<Const>() {
-                        match op.0.datum_type() {
-                            DatumType::TDim => {
+                        if op.0.datum_type() == DatumType::TDim { {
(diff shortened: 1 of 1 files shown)


cli/src/params.rs
```diff
@@ -736,7 +736,31 @@ impl Parameters {
...
             }
-            stage!("set", typed_model -> typed_model, |m: TypedModel| {
+            stage!("set", typed_model -> typed_model, |mut m: TypedModel| {
+                for node in m.eval_order()? {
+                    let node = m.node_mut(node);
+                    if node.op_is::<Const>() {
+                        // safe to unwrap as we know it is a Const
+                        let op = node
+                            .op_as_mut::<Const>()
+                            .unwrap();
```
kali, line 746: I think you can simplify this a bit more using a `if let Some(op) = node.op_as::<Const> { ... }`

kali (changes requested):
As I was saying before, I'm happy merging this in the cli code for now. Please just consider the `if let` change.

kali (approved):

kali:
dont worry too much about the CI. credentials for aws access need some work. I'll tell you if something need attention.

kali:
I'm afraid this one is for you: https://github.com/sonos/tract/actions/runs/7440343618/job/20241787343?pr=1296 .

PR #1297, opened 2024-01-08
avoid config-dependant type path

kali:


kali:
@alexander-camuto FYI

PR #1327, opened 2024-02-08
fix: correct elm wise mul with quantized tensors

JulienBalianSonos:
Formulation of elm wise quantization was wrong from my understanding:
`zero_point` addition is needed after `scale` not before.


kali:
Ignore the clippy thing.

PR #1333, opened 2024-02-09
[RFC] feat: more precise qbinary handling in tract core with unaligned entry

JulienBalianSonos:
Change ways binary ops are done for math with quantized tensors, by allowing unaligned qparams in input/output tensors.

This change allows increasing precision of quantization and reduce casting/alignment cost, but we now need to handle the qparams of output tensor as function input.

We need to handle all ops implied by `bin_to_super_type macro` already supporting QU8, that may use unaligned qparams quantization.

- [x] elementwise mul
- [x] div

in next cases we directly apply dequant to f32 and requant to targeted qtype for now 
- [x] add 
- [x] sub
- [x] min
- [x] max (here we set a small optim to make it slightly faster for ReLU case) 

we propose to leave aside from this PR: rem, and all "bool" or "bit" operations


nnef/src/registry.rs
```diff
@@ -217,20 +217,29 @@ impl Registry {
...
-                .wire_as_outlets(tract_core::ops::binary::TypedBinOp(bin.1.clone()), &inputs)?[0];
-            if let Some(Some(out_dt)) = dt.first() {
-                if out_dt != &a_dt {
-                    wire =
-                        builder.wire_as_outlets(tract_core::ops::cast::cast(*out_dt), &[wire])?[0];
+
+            let c_dt: Option<DatumType> = match dt.first().cloned() {
+                Some(c_dt) => c_dt,
+                None => None,
+            };
```
kali, line 232: Nitpicking: I think there's a more idiomatic way :) with `then` or `and then` maybe ?

test-rt/suite-unit/src/q_binary.rs
```diff
@@ -0,0 +1,397 @@
...
+    suite.add_arbitrary::<QBinaryOpProblem>("proptest", ());
+
+    // simplification 0 at declutter constant
+    suite.add(
+        "trivial_mul_0_case",
+        QBinaryOpProblem {
+            operator: tract_core::ops::math::mul(),
+            tensor_a: tensor0(0u8)
+                .cast_to_dt(
+                    u8::datum_type().quantize(QParams::ZpScale { zero_point: 0, scale: 1. }),
```
kali, line 222: I think if we define something like

```
fn qu8(zp: i32, scale: f32) -> DatumType { ... } 
```

we could reduce widely the bloat in test definition. WDYT ?

Note that `?` should actually work there instead of `unwrap()`.

core/src/ops/binary.rs
```diff
@@ -514,10 +538,99 @@ macro_rules! bin_to_super_type {
...
+                    /// Apply to all Q types
+                    /// Take more memory but hopefully faster than memory_optimised_q_binary_as_float_op
+                    /// especially once cast_to_dt will have will have vectorized implementations
+                    fn generic_q_binary_as_float_op(
+                        a: &TValue,
+                        b: &TValue,
+                        c_dt: &DatumType,
+                        accumulator_dt: DatumType
+                    ) -> TractResult<Option<Tensor>> {
+                        if let (Some(QParams::ZpScale {zero_point: a_zp, scale: a_scale}),
```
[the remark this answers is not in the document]
JulienBalianSonos: That was my bad. I expected other kind of quantization scheme in future and wanted to enforce linear quant only.
Given that looking more closely even `MinMax` is seamlessly transformed to zp_scale, and that the only 2 kinds are for now linear, what I did was premature “type” enforcing.
This is now fixed to be more cohesive with rest of the project.

PR #1341, opened 2024-02-23
fix: tdim for other logic operators

JulienBalianSonos:
- support TDim for others cmp operators


kali:
I am not sure this is going to do what we want on the long run. It will compare TDim based on... alphabetic order of symbols, or numbers before symbols and the like. My inclination would be:
* remove Ord and PartialOrd on TDim altogether
* implement comparison by casting to integer first during expansion, adding an actual Cast to i64 node in the net before getting into the comparison.
* fix whatever else is broken, including the Einsum/MatMul heuristics.

WDYT ?

JulienBalianSonos:
Indeed that seems a better way to handle what we realy want. I will update this PR accordingly.

PR #1358, opened 2024-03-22
Abstract file access for tensor as external resources

kali:


onnx/src/data_resolver.rs
```diff
@@ -0,0 +1,81 @@
...
+pub fn default() -> Box<dyn ModelDataResolver> {
+    Box::new(MmapDataResolver)
+}
+
+#[cfg(target_family = "wasm")]
+pub fn default() -> Box<dyn ModelDataResolver> {
+    Box::new(FopenDataResolver)
+}
+
+pub trait ModelDataResolver {
```
kckeiks, line 18: This should work @kali 

Not too important but I wonder if these methods can be consolidated to make the trait simpler.

kali: I'm not sure we can without having it lead to extra copies here and there. Can you see a way ?

kckeiks: Hmm where is `read()` used? Couldn't we just have `read_bytes_from_path`?

kali: indeed. not sure why I incuded it in the first place.

PR #1407, opened 2024-05-09
spurious debug in onnx resize loader

### LINKED ISSUE #1406: Stray `dbg!()` in several crate versions
https://github.com/sonos/tract/blob/5576526b7ddda189d7dbe834f7a14d700022836d/onnx/src/ops/resize.rs#L11

This has made its way into 0.21 crate releases and prints to stderr every time I open an ONNX file.

I suggest using a CI check that fails when the string `dbg!` is encountered in a Rust source file, that typically prevents these issues with essentially zero false positives in most projects.

kali:
Fix #1406

### DIFF
--- a/onnx/src/ops/resize.rs
+++ b/onnx/src/ops/resize.rs
@@ -8,7 +8,7 @@ pub fn resize(
     ctx: &ParsingContext,
     node: &NodeProto,
 ) -> TractResult<(Box<dyn InferenceOp>, Vec<String>)> {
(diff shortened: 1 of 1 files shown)


kali:
Only one of these dbg! is in actual production code, the rest are in tests. I have mixed feeling about removing them. I completely hear you, dbg!() in production code is absolutely not acceptable. On the other hand, it's to my knowledge the first time it happens in tract. I'm gonna leave the ones in tests for now, and will reassess if it happens again.

PR #1415, opened 2024-05-23
F32 To F16 transform with node filter predicate

hdlj:


core/src/floats.rs
```diff
@@ -33,58 +111,69 @@ impl<T1: Datum + Float, T2: Datum + Float>
...
-        } else if let Some(op) = node.op_as::<EinSum>() {
-            Box::new(EinSum {
-                operating_dt: dt_float_precision_conversion::<T1, T2>(op.operating_dt),
-                ..op.clone()
-            })
-        } else if let Some(op) = node.op_as::<Pad>() {
-            if let PadMode::Constant(t) = &op.mode {
-                Box::new(Pad {
-                    mode: PadMode::Constant(tensor_float_precision_conversion::<T1, T2>(t)),
+        let is_source = node.op_as::<TypedSource>().is_some();
```
kali, line 119: Not sure why Source need a special treatment ?

core/src/floats.rs
```diff
@@ -12,8 +12,86 @@ use crate::ops::scan::Scan;
...
+    }
+
+    fn should_translate_node(&self, node: &Node<TypedFact, Box<dyn TypedOp>>) -> bool {
+        self.node_predicate.as_ref().map(|it| (it)(node)).unwrap_or(true)
+    }
+
+    /// Cast node input to the working float precision for the operator
+    /// Only input using float datumtype are impacted. This will add cast operations
+    /// in the model. The function return the new input outlet ids.
+    fn cast_input_if_required(
```
kali, line 35: cast_node_inputs_if_required (NB plural)

core/src/floats.rs
```diff
@@ -12,8 +12,86 @@ use crate::ops::scan::Scan;
...
+                mapped_inputs.push(mapping[i])
+            }
+        }
+        Ok(mapped_inputs)
+    }
+
+    /// Cast node output outlet ids to the destination float precision,
+    /// after insertion in the target mode. This preserves the model output float
+    /// precision.
+    fn cast_model_output_if_required(
```
kali, line 64: cast_node_outputs_if_required

core/src/floats.rs
hdlj: I am treating the model input differently in order to preserve that input and ouput at the model level are still f16 after a f32-to-f16 transformation. That is also the purpose of the function `cast_model_outputs_if_required ` but on the model output side. 

If a source is needed in f32 from an input of f16, cast operation will be added.

PR #1425, opened 2024-05-31
Improve reduce sum

emricksinisonos:


core/src/ops/nn/reduce.rs
```diff
@@ -142,44 +143,58 @@ impl Reducer {
...
+        f32: AsPrimitive<T>,
     {
         if axes.len() == 0 {
             return input.to_owned();
         }
+
         let mut output: Option<ArrayD<T>> = None;
-        for axis in axes.iter() {
-            let current_input = output
+        for axis in axes.iter().cloned() {
```
kali, line 155: Nitpicking: can we dispense with the cloned() ? These are just usize, right ? so `&axis` or at least `.copied()` should be enough?

linalg/src/frame/reduce/mod.rs
```diff
@@ -280,7 +249,7 @@ pub mod test {
         let found = values;
         let red = op.run_with_params(found, params).unwrap();
         tensor0(red)
-            .close_enough(&tensor0(expected), true)
+            .close_enough(&tensor0(expected), Approximation::SuperApproximate)
```
kali, line 252: do we still need SuperApprox when we use integer-as-float inputs ?

linalg/src/frame/reduce/sum.rs
```diff
@@ -0,0 +1,60 @@
...
+    use proptest::test_runner::TestCaseResult;
+
+    #[macro_export]
+    macro_rules! sum_frame_tests {
+        ($cond:expr, $t: ty, $ker:ty) => {
+            proptest::proptest! {
+                #[test]
+                // Addition is not associative in f16 a + b + c != (a + b) + c because of precision
+                // issues. To avoid that, test vector should be aligned with kernel size (to avoid
+                // having to sum partial sums together).
```
kali, line 16: is this still true ? it is a bit misleading in that f32 is not associative either.

linalg/src/generic/reduce.rs
```diff
@@ -0,0 +1,189 @@
...
+    #[cfg(test)]
+    #[macro_use]
+    pub mod h {
+        use super::*;
+        crate::sum_frame_tests!(true, f16, crate::generic::reduce::sum::HSum8);
+    }
+}
+
+// Softmax generic implementation
+pub mod softmax {
```
kali, line 109: Should this be softmatxl2 instead ?

kali (approved):
Looks good, just a couple of nitpicking.

core/src/ops/nn/reduce.rs
emricksinisonos: Yes, I just didn't want to dereference axis everywhere in the scope. I'll change that to `.copied()`.

linalg/src/frame/reduce/mod.rs
emricksinisonos: You're right, it should be fine with `Approximate`

PR #1436, opened 2024-06-13
Improve Mul (f32, f16) on aarch64

emricksinisonos:


core/src/ops/math/mod.rs
```diff
@@ -52,6 +54,125 @@ where
 bin_to_super_type!(mul, Mul,
                    cost: |dt| tvec!((Cost::FMA(dt), 1)),
                    declutter: declutter_mul,
+                   eval_in_a: |a: &mut Tensor, b: &Tensor| -> TractResult<bool>{
```
kali, line 57: any reason not no make this a function ? if only for the stake of having rustfmt processing it :)

linalg/src/frame.rs
```diff
@@ -1,6 +1,8 @@
 #[macro_use]
 pub mod element_wise;
 
+#[macro_use]
+pub mod binary;
```
kali, line 5: This is actually the unicast case, right ? Should we call it bin_unicast ? (and maybe rename by_scalar bin_scalar) ?

linalg/src/frame/binary/mul.rs
```diff
@@ -0,0 +1,40 @@
...
+        };
+    }
+
+    pub fn test_binary_mul<K: BinaryKer<T>, T: LADatum + Float>(values: &[f32]) -> TestCaseResult
+    where
+        f32: AsPrimitive<T>,
+        T: AsPrimitive<f32>,
+    {
+        crate::setup_test_logger();
+        let a: Vec<T> = values.iter().copied().map(|x| x.as_()).collect();
```
kali, line 36: we need two random inputs here, reusing them will make dozens of bugs not showing.

linalg/src/frame.rs
emricksinisonos: yep, I've done the renaming:
- binary -> unicast
- by_scalar unchanged
(just to remain short on the naming)

linalg/src/frame/binary/mul.rs
emricksinisonos: Ok, I'll change that !

PR #1439, opened 2024-06-14
Opaque metadata attached to TypedFact

hdlj:


core/src/model/fact.rs
```diff
@@ -286,6 +301,15 @@ impl TypedFact {
     pub fn without_value(&self) -> Self {
         Self::dt_shape(self.datum_type, self.shape.clone())
     }
+
+    pub fn opaque_metadata<T: OpaqueMetadata>(&self) -> Option<&T> {
```
kali, line 305: I'm not too sure about this one, because there is no way to distinguish the case where metadata is missing, and the case where metadata as the wrong type.

It can be written `typed_fact.opaque_metadata.and_then(|om| om.downcast_ref::<...>)`, and it's not much longer...

WDYT ?

data/src/opaque.rs
```diff
@@ -7,10 +8,47 @@ use std::sync::Arc;
 use downcast_rs::{impl_downcast, Downcast};
 use dyn_hash::DynHash;
 
-pub trait OpaquePayload: DynHash + Send + Sync + Debug + Display + Downcast {}
+pub trait OpaquePayload: DynHash + Send + Sync + Debug + Display + Downcast {
+
+    fn metadata(&self) -> Option<Box<dyn OpaqueMetadata>> {
```
kali, line 13: do we even need this ? Or is it up to the code instantianting the TypedFact to manage the OpaquePayload directly ? Asking because it does not look like we're using it anywhere in the PR.

Also it makes it a bit murky in the case where we have several OpaquePayload in a tensor. Better leave it entirely to the TypeFact instantiator, i think.

core/src/model/fact.rs
hdlj: yes I am changing this

data/src/opaque.rs
hdlj: The only situation, I see this function call it is when we receive a tensor in the `eval` and we want to get the metadata. But we could consider that in such situation, the user needs to downcast the opaque tensor to a concrete type and build the metadata itself ? WDYT ?

kali: Well, i just don't think we should assume the metadata should be constructed from the Opaque. If we build a TypedFact with metadata, we will just explicitely add it. This solves the problem where we have a tensor with multiple opaque inside: they could potentially generate different metadata, but the fact only stores one.

So yeah, If the metadata is still helpful at the `eval` stage, its code will just generate it again from the concrete Opaque type.

PR #1468, opened 2024-07-12
Refactor binary

emricksinisonos:


core/src/ops/mod.rs
```diff
@@ -11,6 +11,7 @@ pub mod macros;
 pub mod element_wise;
 #[macro_use]
 pub mod binary;
+//pub mod binary_new;
```
kali, line 14: ...

data/src/tensor/indices.rs
```diff
@@ -0,0 +1,73 @@
+pub struct IndexIterator {
```
kali, line 1: is this better / different than ndarray::indexes() ?

linalg/src/arm64/arm64fp16/by_scalar.rs
```diff
@@ -34,8 +34,183 @@ ew_impl_wrap!(
...
+        assert!(buf.len() % 16 == 0);
+        assert!(buf.len() > 0);
+        #[target_feature(enable = "fp16")]
+        unsafe fn run(buf: &mut[f16], s: f16) {
+            let len = buf.len();
+            let ptr = buf.as_ptr();
+            std::arch::asm!("
+            dup v0.8h, v0.h[0]
+            2:
+                ld1 {{v4.8h, v5.8h, v6.8h, v7.8h}}, [{ptr}]
```
kali, line 53: we need to give a shot at unrolling to amortize the loading latency, and prefetch too (but this is another task/PR)

linalg/src/frame/by_scalar.rs
```diff
@@ -1,3 +1,61 @@
...
+
+use crate::{LADatum, element_wise::ElementWiseKer, LinalgFn};
+
+use super::{ElementWise, element_wise_helper::map_slice_with_alignment};
+
+
+/// Generic implementation struct that unify all by scalar kernels.
+/// A by scalar operation is an ElementWise operation with a scalar paramerer.
+#[derive(Debug, Clone, new)]
+pub struct ByScalarImpl<K, T>
```
kali, line 13: I'm slowing moving away from this pattern in the matmul code. We may need to discuss it. (mmm kernels are now represented by the instance of a struct, lazy-initialized const)

linalg/src/lib.rs
```diff
@@ -183,13 +181,73 @@ lazy_static::lazy_static! {
...
+    };
+}
+
+pub fn bin_by_scalar(dt: DatumType, bin: BinOp) -> Option<Box<LinalgFn>> {
+    let map = BIN_BY_SCALAR_OPS.lock().unwrap();
+    map.get(&(bin, dt)).map(|it| (it)())
+}
+
+pub fn bin_unicast(dt: DatumType, bin: BinOp) -> Option<Box<LinalgFn>> {
+    let map = BIN_UNICAST_OPS.lock().unwrap();
```
kali, line 238: we should not need the mutex here, as we are initialized in a lazy_static. is this visible in profiles ? i think we may want to consider non-hashmap options (we can make BinOp repr(C) and use it as integer for instance, maybe something similar for dt ?)

kali (commented):
Lets's hace a chat later today :)

data/src/tensor/indices.rs
emricksinisonos: After re-checking, `ndarray::indexes()` does the job !

PR #1495, opened 2024-08-08
[RFC] fix: opaque tensors and declutter stage

JulienBalianSonos:
Hello,
This PR aims to be a base for discussion about:
`Const::new_with_opaque_fact` typed op that does not support some declutter steps. 

Indeed from debuging session so far (if intepretted correctly):
-  `tract_data::Tensor` with dt Opaque  has .shape empty, it's only in `OpaqueFact` that hold proper values, hence declutter like `tract_core::optim::slice::PushSliceUp` will fail because there is no .rank.

I tried to add the shape in the opaque tensor at nnef deserialization (not in this PR)
but this is not sufficient because when slice eval operate on it: it is not handled today, and I am unsure about how to make this work well (+ accounting for the 32 divisibility constraint of sliced output).

Any advice or guidance ?


kali:
There are definitely issues with Opaque and OpaqueFact, but I don't think this is one of these. Here, we are trying to push up a Slice operator over the m axis from below a Matmul (represented by Einsum), but the weights are in block quant form, so axes are... weird.

 This is not theoretically impossible, but I'm not sure we want to go there, so I'm disabling it from now in #1506 which supersedes this one. Thanks for the helpful test-case :)

PR #1497, opened 2024-08-13
[WIP] debug regression in mul operator with 'big' dimensions - post tract 0.21.5

JulienBalianSonos:
Regression appeared after tract v0.21.5.

Simple mul operator between an input and a "variable" tensor fail when specific dimensions implied.
 
Original failing test is with shapes:
- input(1, 4, 256, 128, 2)
- variable (1, 256, 1, 1) 

we manage to reduce the the problem to
- input (2, 2, 17, 2)
- variable (1, 2, 1, 1)

But not further (reducing any dim from here allow exact results).

The problem is not sensible to the values in input tensor, nor parameter,
This is likely a multiplication layout issue.


emricksinisonos:
As discussed, this bug was introduced in the `Mul` operator while adding the uniform/unicast optimisation.
This PR fixes the issue: #1500

JulienBalianSonos:
I confirm after testing: the new PR fixes the issue.

PR #1500, opened 2024-08-19
Fix mul uniform/unicast selection strategy

emricksinisonos:
When checking if uniform/unicast optim should be applied, consistency of the remaining shapes wasn't checked.
For example, shapes A: [1, 2, 1, 1] x B: [2, 2, 2, 16] would have been selected for uniform (by_scalar) evaluation while a A[0] != B[0], leading to false output results.


JulienBalianSonos:
LGTM, solve [Mul issue](https://github.com/sonos/tract/pull/1497) 
but opaque CI fail in Native Script

kali (approved):

PR #1518, opened 2024-09-04
draft: Face similarity example

heabeounMKTO:


Cargo.toml
```diff
@@ -50,7 +51,7 @@ members = [
     "test-rt/test-unit-core",
     "test-rt/test-onnx-core",
     "test-rt/test-nnef-cycle",
-    "test-rt/test-tflite", 
+    "test-rt/test-tflite", "examples/face_similarity_arcface_onnx", 
 ]
```
kali, line 55: Can you adjust the formatting / ordering please ?

kali (changes requested):
Looks pretty nice.

* We may need a (short) Readme.
* We need a ci.sh file. At minimum running "cargo check".
* Can we turn the main() into a #[test] and make it part of the ci check ?

Cargo.toml
heabeounMKTO: yes, i will update it shortly

kali (approved):

PR #1540, opened 2024-09-26
fix: op_set version v14 needed at ONNX export with newer version of torch for Albert

JulienBalianSonos:
- to avoid this kind of error in the future, you may wish to freeze torch/transformers/onnx versions in pip install from `ci.sh`


kali:
thanks a bunch

PR #1580, opened 2024-11-25
ensure nnef tars can be deterministic

tgolsson:
Hey!

Minor patch here to ensure that the NNEF tars can be made deterministic. Before this, the files would be in random order and the timestamps would change each time the conversion ran, which makes it show up as modified in VCS systems even if the contents are technically the same. 

Happy to bike-shed if you want more/less configurability here, I figured the key sorting overhead is small enough to not matter. The 1980-01-01 date is a common placeholder, for example a bunch of container tools use it. I've found in the past it's more reliable than 1970-01-01, which might be interpreted as "not set" even when explicitly set to 0. Upstream defines this to be 2006-07-23, which has symbolic value but looks more like a real date.


kali:
Hey @tgolsson !

So this is actually something I have refused to do in the past for my internal client teams... I told them that the .tgz content was deterministic, but making the .tgz itself deterministic would be too much of a hassle...

I assume your plan is to use this from the API. Maybe we should still add a `--nnef-deterministic` option in the `dump` command ?

kali:
you can ignore the wasm32 build issue, i have it fixed on main.

tgolsson:
Added the flag, but I can never remember the magic words for tract's CLI...

```
cargo run --bin tract -- ../cervo/brains/test.onnx -O --onnx-ignore-output-shapes  --nnef-tract-onnx  --nnef-tract-core dump --nnef-tar foo.nnef.tar --nnef-deterministic
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.16s
     Running `target/debug/tract ../cervo/brains/test.onnx -O --onnx-ignore-output-shapes --nnef-tract-onnx --nnef-tract-core dump --nnef-tar foo.nnef.tar --nnef-deterministic`
[2024-11-26T10:00:30.152081112Z WARN  tract_onnx::model] ONNX operator for your model is 8, tract is only tested against operator set 9 to 18 (included). Your model may still work so this is not a hard fail.
[nnef/src/framework.rs:109:9] timestamp = 315532800s
[2024-11-26T10:00:30.231729468Z ERROR tract] Writing model to tar

    Caused by:
        0: Translating model to proto_model
        1: Translating model to AST
        2: translating node #1 "StatefulPartitionedCall/model_11/mlp_encoder_4/dense_8/MatMul.pack_b" OptMatMulPack
        3: No serializer found for node #1 "StatefulPartitionedCall/model_11/mlp_encoder_4/dense_8/MatMul.pack_b" OptMatMulPack
```

tgolsson:
Works fine, and the diff is empty!

PR #1591, opened 2024-12-03
Metal GPU profiling

LouisChourakiSonos:
Most Metal model operators currently run on the Metal GPU. However, the current profiler only shows the CPU timespan for operations, which does not provide an accurate insight into chip performance due to the asynchronous nature of the process (i.e Sync operators are shown as the most consuming ones).

The PR enables Metal GPU profiling using metal-rs API.
It also updates the CLI to nicely print GPU performances (or more generally accelerator performances if any) next to CPU ones


metal/src/command_buffer.rs
```diff
@@ -0,0 +1,157 @@
...
+        }
+    }
+
+    pub fn add_node_entry(&mut self, node_id: usize) {
+        self.profile_buffers.insert(node_id, vec![]);
+        self.current_node_id = Some(node_id);
+    }
+
+    pub fn add_buffer(&mut self, buffer: Buffer) {
+        let current_node_id = &self.current_node_id.unwrap();
```
hdlj, line 71: unwrap should be avoid. We could return an Error instead with TractResult:

```
self.current_node_id.ok_or_else(|| anyhow!("Metal profile doesn't have any current node id which is unexpected while adding the sampling buffer"))?;
```

metal/src/command_buffer.rs
```diff
@@ -0,0 +1,157 @@
...
+    pub fn get_profile_data(&self) -> HashMap<usize, u64> {
+        let mut formatted_hashmap: HashMap<usize, u64> = HashMap::new();
+        self.profile_buffers.borrow().iter().for_each(|(key, v)| {
+            let mut node_duration_ns = 0;
+            v.iter().for_each(|buffer| unsafe {
+                let slice = std::slice::from_raw_parts(
+                    buffer.contents() as *const u64,
+                    NUM_SAMPLES as usize,
+                );
+                node_duration_ns += slice[1] - slice[0];
```
hdlj, line 88: the sampling is a cycle number or a true duration ?

hdlj:
Thanks @LouisChourakiSonos  for the PR. Could you add the metal session handler support in the bench subcommand line also ?

metal/src/command_buffer.rs
LouisChourakiSonos: It is a true duration in ns

cli/src/bench.rs
```diff
@@ -37,11 +37,29 @@ pub fn handle(
...
         params.tract_model.downcast_ref::<TypedModel>().context("Can only bench TypedModel")?;
     let inputs = tract_libcli::tensor::retrieve_or_make_inputs(model, &run_params)?.remove(0);
-    let plan = SimplePlan::new_with_options(model, &plan_options)?;
 
     limits.warmup(model, &inputs)?;
 
-    let mut state = SimpleState::new(plan)?;
+    let mut state = {
+        #[cfg(any(target_os = "macos", target_os = "ios"))]
+        {
```
kali, line 45: should this only happen with `--metal` ?

LouisChourakiSonos: Good catch! Creating a MetalSessionHandler from a non-Metal plan is not really an issue, as it will just do nothing and the model will execute as normal. But for user readability I'll add a check on the flag

PR #1599, opened 2024-12-09
[Metal] Rewrite rules: Remove MetalSyncToCPU between Const and CPU Operator

hdlj:
After making MetalSyncToCpu stateful to avoid PropConst with Metal Op, some MetalSyncToCPUs were reintroduced which slow down the execution. This rewrite rule remove these ones.

### DIFF
--- a/metal/src/rewrite_rules/rewire_metal_sync.rs
+++ b/metal/src/rewrite_rules/rewire_metal_sync.rs
@@ -46,10 +46,8 @@ pub fn rewire_metal_sync_after_const(
     rule_ensure!(sync_cpu_op.kind == MetalSyncKind::ToCpu);
 
     let mut patch = TypedModelPatch::default();
-
-    let konst_input = patch.taps(model, &node.inputs)?;
     let out =
-        patch.wire_node(format!("{node_name}"), Const(cpu_const.into(), None), &konst_input)?;
+        patch.wire_node(format!("{node_name}"), Const(cpu_const.into(), None), &[])?;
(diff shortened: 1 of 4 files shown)


metal/src/rewrite_rules/rewire_metal_sync.rs
```diff
@@ -25,3 +27,29 @@ pub fn rewire_metal_sync(
...
+    // Identify successor ToCpu
+    let Some(sync_cpu) = next_node(model, node) else { return Ok(None) };
+    let Some(sync_cpu_op) = sync_cpu.op_as::<MetalSync>() else { return Ok(None) };
+    rule_ensure!(sync_cpu_op.kind == MetalSyncKind::ToCpu);
+
+    let mut patch = TypedModelPatch::default();
+
+    let konst_input = patch.taps(model, &node.inputs)?;
+    let out =
+        patch.wire_node(format!("{node_name}"), Const(cpu_const.into(), None), &konst_input)?;
```
kali, line 52: Mmmm what ? Const takes zero input, so why do we need `konst_input` ?

(Also, if we just want to kill the MetalSync, we can do so with https://docs.rs/tract-core/latest/tract_core/model/struct.ModelPatch.html#method.shunt_one_op)

hdlj: Yes I can remove the tap. However I need to rewrite the Const with the CPU tensor. I can use the replace op also but it is more concise to do a dedicated path ?

PR #1600, opened 2024-12-10
isolated metal translation phases test

kali:


hdlj:
Good catch thanks a lot @kali !

hdlj (approved):

PR #1643, opened 2025-02-14
Metal GGML Q40 support

LouisChourakiSonos:


metal/src/kernels/matmul/ggml_gemm/mod.rs
```diff
@@ -33,14 +34,14 @@ impl GemmKernel for GgmlGemm {
...
-        if dts[0] == DatumType::F32 {
-            Ok(dts[1] == DatumType::F32)
-        } else {
-            Ok(dts[0] == DatumType::F16 && matches!(dts[1], DatumType::F32 | DatumType::F16))
-        }
+        (as_q40_fact(&facts[1]).is_some()
+            && matches!(facts[0].datum_type, DatumType::F16 | DatumType::F32))
+            || ((facts[1].datum_type == DatumType::F32) && (facts[0].datum_type == DatumType::F32))
+            || ((facts[1].datum_type == DatumType::F16)
+                && matches!(facts[0].datum_type, DatumType::F32 | DatumType::F16))
```
kali, line 44: lol, this is very hard to parse :) maybe we could split the tricky q40 case apart then do somethings like
```
use DatumType::*;
match (facts[0].datum_type, facts[1].datum_type) {
  (F32, F32) => true,
  (F16, F16) => true,
  (F16, F32) => true,
  _ => false
}
```

metal/src/kernels/matmul/mod.rs
```diff
@@ -205,14 +213,18 @@ impl<M: GemmKernel> GemmImpl<M> {
...
-    pub fn output_facts(&self, a: &TypedFact, b: &TypedFact) -> TractResult<TVec<TypedFact>> {
-        let out_shape = self.output_shape(&a.shape, &b.shape).to_vec();
-        let out_dt = self.matmul.output_dt(a.datum_type().unwrap(), b.datum_type().unwrap())?;
+    pub fn output_facts(
+        &self,
+        shape: &[TDim],
+        a_dt: DatumType,
+        b_dt: DatumType,
+    ) -> TractResult<TVec<TypedFact>> {
+        let out_dt = self.matmul.output_dt(a_dt, b_dt)?;
```
kali, line 222: how about something that respect the symmetry more ?
```
ensure!([DatumType::F16, DatumType::F32].contains(out_dt));
Ok(tvec!(out_dt.fact(shape))
```

metal/src/kernels/matmul/mod.rs
```diff
@@ -636,36 +717,51 @@ mod tests {
...
-                    let datum = (0usize..10).prop_map(|x| x.as_());
+        fn arbitrary_with(params: MmmProblemParams) -> Self::Strategy {
+            (1usize..4, 1usize..128, 1usize..256, 1usize..128)
+                .prop_flat_map(move |(b, m, mut k, n)| {
+                    if params.q4_0_weights {
+                        k = k.div_ceil(32) * 32
+                    };
+
+                    let mut rng = rand::thread_rng();
+                    let lhs_data: Vec<F> = (0..b * m * k)  // Create a vector with 10 elements
```
kali, line 731: i suspect we have a lying comment here.

metal/src/kernels/matmul/mod.rs
```diff
@@ -636,36 +717,51 @@ mod tests {
...
+                    .map(|_| F::from(rng.gen_range(0.0..1.0)).unwrap()) // Generate a random float in [0.0, 1.0)
```
kali, line 732: mmm... this is not good. we should not rand() ourselves, but let the proptest strategies do it in predictable and simplifiable manner. The idea is to use `proptest::collections::vec()` or something like that.

kali (approved):

PR #1660, opened 2025-02-28
Get Arc tensor in Syncs instead of cloning inner value

LouisChourakiSonos:


cli/src/utils.rs
```diff
@@ -93,18 +93,16 @@ pub fn check_inferred(got: &[InferenceFact], expected: &[InferenceFact]) -> Trac
...
         .map(|t| {
-            t.to_scalar::<Opaque>()
-                .and_then(|ot| ot.clarify_to_tensor().transpose())
-                .ok()
-                .flatten()
+            Ok(t.to_scalar::<Opaque>()?
+                .clarify_to_tensor()?
                 .map(|ot| ot.into_tvalue())
-                .unwrap_or(t.clone())
+                .unwrap_or(t.clone()))
```
kali, line 103: unwrap_or_else ?

kali: mmm... not too bad here as we are just shallow-cloning the tvalue, actually (i wonder if clippy is not going to catch this one anyway)

metal/src/kernels/array/concat.rs
```diff
@@ -147,7 +147,7 @@ mod tests {
                     axis,
                     &inputs.iter().map(|it| it.to_cpu()).collect::<Result<Vec<_>>>()?,
                 )?;
-                assert_eq!(ref_output, output.to_cpu()?);
+                assert_eq!(ref_output, output.to_cpu()?.into_tensor());
```
kali, line 150: should this be made a debug_assert_eq ?

metal/src/kernels/array/permute_axes.rs
```diff
@@ -147,8 +147,8 @@ mod tests {
                 let a = Tensor::from_shape(shape, &a_data)?.into_metal()?;
 
                 let output = PermuteAxes.eval(context, &a, axes)?;
-                let ref_output = a.to_cpu()?.permute_axes(axes)?;
-                assert_eq!(ref_output, output.to_cpu()?);
+                let ref_output = a.to_cpu()?.into_tensor().permute_axes(axes)?;
+                assert_eq!(ref_output, output.to_cpu()?.into_tensor());
```
kali, line 151: same

metal/src/kernels/array/concat.rs
kali: ho ! it's test code. nevermind.

kali (approved):

PR #1698, opened 2025-03-28
Fix Einsum rule

LouisChourakiSonos:
Einsum_as_matmul rule is adding unecessary axes to A and B when m axis is not present in the output (m dimension is necessarily 1 in those cases).
The PR makes the rule more flexible to this


kali:
Extra "full" CI: https://github.com/sonos/tract/actions/runs/14170452524

kali (approved):

kali:
Extra LLM CI: https://github.com/sonos/tract/actions/runs/14170878427

PR #1733, opened 2025-04-30
call simplify on hand-made partition size tdim expr

kali:


kali:
@aaronjonessonos This could influence very significantly the cost of reasoning over the partition size :)

PR #1761, opened 2025-06-13
Cuda: Unary + Binary Ops

LouisChourakiSonos:


cuda/src/kernels/launch_args.rs
```diff
@@ -0,0 +1,17 @@
+use cudarc::driver::{DeviceRepr, LaunchArgs, PushKernelArg};
+use tract_core::internal::tract_smallvec::SmallVec;
+use tract_gpu::tensor::DeviceTensor;
+
+use crate::tensor::CudaBuffer;
+
+pub trait LaunchArgsExt<'a> {
+    fn set_slice<T: DeviceRepr>(&mut self, slice: &'a SmallVec<[T; 4]>);
```
kali, line 8: there may be a more generic parameter here, something like &[T] or even `impl AsRef<[T]>`. As a rule of thumb, it rarely make sense to take Vec or SmallVec by reference.

PR #1774, opened 2025-07-01
Fix optim transpose match case

LouisChourakiSonos:


kali:
merging despite the perf regression on transformers cpu

PR #1796, opened 2025-07-25
CUDA GPU tracing + CLI refactor

LouisChourakiSonos:


LouisChourakiSonos:
> > Not sure about moving the --metal-gpu-trace and --cuda-gpu-trace to output_options() parser. Tell me what you think @kali . I am pretty convinced it only makes sense to use those options in a profiling context, so moving them to the subcommand sounds better
> 
> What about the "bench" subcommand ? How about adding the to the "run" options set ?

Sounds good! I was a bit reluctant to add the dirty closure thing in other commands like I did for profile_gpu but actually, not that big of a deal. In that case, I guess adding the options to the run option set makes sense too.

kali (approved):

PR #1828, opened 2025-08-24
fix: loosen half crate dependency for compatibility with wgpu

jackrr:
Thanks for open-sourcing and maintaining such a fast onnx model inference implementation! It's been great for my project https://github.com/jackrr/eymo! Much more performant than when I was using [ort](https://github.com/pykeio/ort) to call out to ONNX runtime.

I've recently run into this minor issue using the crate alongside [wgpu](https://github.com/gfx-rs/wgpu).
wgpu depends on half=^2.5, so is mutually incompatible with tract pinned to half=2.4.1.

I don't know the `tract` implementation details well enough to claim this is a safe change to make, nor do I believe that the integration testing with `half=2.6.0` verified by my project is sufficient enough. But I figured I'd open this up since I'd made the fork to unblock my project.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -151,7 +151,7 @@ foreign-types = "0.5"
 fs-err = "2"
 fs2 = "0.4.3"
 getrandom = "0.2"
-half = { version="2.4.1", features = [ "std", "num-traits" ] }
+half = { version=">=2.4,<3.0", features = [ "std", "num-traits" ] }
 home = "0.5.9"
 icu_collections = "1.5.1"
 icu_normalizer = "1.5.0"


jackrr:
Still verifying the fix compiles and runs without issue in my project (https://github.com/jackrr/eymo). Will move to ready once verified.

kali (approved):
Looks good. Impact, if any, will only be on speed, and only be observable on main. I'll merge it and have a look on the numbers tomorrow.

kali:
Will reopen if nnumbers are not good tomorrow.

kali:
All good, no regression.

PR #1874, opened 2025-10-08
Fix Qwen2.5 GPU

emricksinisonos:
Following error occurs when running Qwen2.5 model families, it happens on Metal only in `f16` (should also on CUDA but tests models are too big to run on dev machines):

```
[2025-10-08T08:18:44.007653000Z ERROR tract] Error at stage "metal"

    Caused by:
        0: Evaluating rewriting rule "einsum-to-prefix-matmul" on node #9 "model_model__0_selfAttn_qProj_linear0__linear" EinSumMatMul
        1: Trying to substitute a 1,S,3584,F32 by 1,S,3584,F16 as output #0 of #9 "model_model__0_selfAttn_qProj_linear0__linear" EinSumMatMul.
```

Error is specific to Qwen2.5 families because linear projection are expressed as `einsum` operators (+ bias + cast) for this model while others directly use the `linear` operator. This is a consequence of a [specific configuration enforcement in T2N](https://github.com/sonos/torch-to-nnef/blob/1bd50d1fc549251006d30ee5b7467c83cca9b3e5/torch_to_nnef/llm_tract/exporter.py#L744-L749).

Qwen2.5 q projection:
```
    model_model_layers_0_self_attn_q_proj_weight = variable<scalar>(label = 'model.model.layers.0.self_attn.q_proj.weight', shape = [3584, 3584]);
    model_model_layers_0_self_attn_q_proj_bias = variable<scalar>(label = 'model.model.layers.0.self_attn.q_proj.bias', shape = [3584]);
    model_model__0_selfAttn_qProj_linear0__linear = tract_core_einsum([model_model__0_inputLayernorm_hiddenStatesMul1, model_model_layers_0_self_attn_q_proj_weight], expr = 'bij,kj->bik', acc = 'f32', output = '');
    model_model__0_selfAttn_qProj_linear0__biasf32 = tract_core_cast(model_model_layers_0_self_attn_q_proj_bias, to = 'f32');
    model_model__0_selfAttn_qProj_linear0__biasf32_unsqueezed = unsqueeze(model_model__0_selfAttn_qProj_linear0__biasf32, axes = [0, 1]);
    model_model__0_selfAttn_qProj_linear0__biased = add(model_model__0_selfAttn_qProj_linear0__linear, model_model__0_selfAttn_qProj_linear0__biasf32_unsqueezed);
    model_model__0_selfAttn_qProj_linear0 = tract_core_cast(model_model__0_selfAttn_qProj_linear0__biased, to = 'f16');
```

Llama q projection:
```
    model_model_layers_0_self_attn_q_proj_weight = variable<scalar>(label = 'model.model.layers.0.self_attn.q_proj.weight', shape = [2048, 2048]);
    model_model_layers_0_self_attn_q_proj_weight_aligned_rank_expanded = unsqueeze(model_model_layers_0_self_attn_q_proj_weight, axes = [0]);
    model_model__0_selfAttn_qProj_linear0 = linear(model_model__0_inputLayernorm_hiddenStatesMul0, model_model_layers_0_self_attn_q_proj_weight_aligned_rank_expanded, 0.0);
```

## Issue with the einsum
When converting Einsum to `PrefixMatmul` for GPU gemm conversion, the output type is wrongly chosen to be `f16` as this decision is only made by choosing the same datum type as the input.

This raises the error reported above as the expected output for this node is expected to be `f32` from the semantic of the typed model.

The fix is to rely on `Einsum` operating dt

### DIFF
--- a/core/src/ops/einsum/prefix_matmul.rs
+++ b/core/src/ops/einsum/prefix_matmul.rs
@@ -32,9 +32,9 @@ fn rule(
 ) -> TractResult<Option<TypedModelPatch>> {
     // F: 2 inputs
     // Q: 9 inputs
-    if !((op.q_params.is_none() && node.inputs.len() == 2)
-        || (op.q_params.is_some() && node.inputs.len() == 9))
-    {
+    let is_fp_mm = op.q_params.is_none() && node.inputs.len() == 2;
+    let is_q_mm = op.q_params.is_some() && node.inputs.len() == 9;
+    if !(is_fp_mm || is_q_mm) {
         return Ok(None);
     }
     if op.q_params.is_some()
@@ -132,10 +132,11 @@ fn rule(
         let input_facts = model.node_input_facts(node.id)?;
         let a_dt = input_facts[0].datum_type;
         let b_dt = input_facts[1].datum_type;
+        let operating_dt = quantize_output.unwrap_or(op.operating_dt);
         let allowed_dt = matmul_semantic_output_dt(&a_dt, &b_dt);
 
         ensure!(
-            op.operating_dt == allowed_dt,
+            operating_dt == allowed_dt,
             format!(
                 "Strict matmul semantic require operating_dt to be {allowed_dt:?} \
                 for (a: {a_dt:?}, b:{b_dt:?}) but got {:?}.",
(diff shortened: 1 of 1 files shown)


kali (approved):

kali:
Is this actually enough ? Or do we need to cascade it to the next operators in the conversion chains ?

kali:
I think nnef-serialization may need a fix. PrefixMatMul is serialized to nnef "matmul", which will not support the type enforcing, so we will loose the information if we go through a nnef cycle... (Now that I think about it, I think I designed the PrefixMatMul to be matching "nnef" matmul, so we may have a couple of issues lurlking here and there.)

We could fallback to serializing as einsum when the accumulator does not match the input.

kali:
Then this operator is consumed by cuda and metal, so we need to figure this out, but also (maybe less important) tflite serialization, but as_blas() function for blas-based support...

PR #1887, opened 2025-10-20
Add Borrow trait implementation for Graph to support model caching

dbaumberger:
ROBLEM:
  TypedRunnableModel<TypedModel> cannot be stored in OnceCell, Mutex, or other
  containers due to missing Borrow<Graph<...>> trait implementation.
  
  This forces re-parsing of ONNX models on every
  inference call, causing severe performance degradation.

SOLUTION:
  Add trivial Borrow<Self> implementation for Graph<F, O> in:
  core/src/model/graph.rs
  
  This is a zero-cost, non-breaking change that enables model caching.

### DIFF
--- a/core/src/model/graph.rs
+++ b/core/src/model/graph.rs
@@ -4,6 +4,7 @@ use crate::ops::Op;
 use crate::plan::PlanOptions;
 use crate::prelude::*;
 
+use std::borrow::Borrow;
 use std::fmt;
 use tract_data::internal::*;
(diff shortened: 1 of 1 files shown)


kali:
I'm ok with the proposed patch, but I'm not sure I understand the gist of the problem. 

I typically put a TypedRunnableModel into an Arc and never ran into issues...

kali:
Not compiling, and probably not needed, so closing.

PR #1935, opened 2026-01-06
Added support for LinearClassifier, LinearRegressor and Normalizer

threeseed:
* Adds support for ONNX LinearClassifier, LinearRegressor, Normalizer ML operations.
* Adds benchmarks for LinearClassifier, LinearRegressor.

Verified across dozens of models.


onnx/src/ops/ml/linear_classifier.rs
```diff
@@ -0,0 +1,320 @@
...
+                tract_core::ops::cast::cast(f32::datum_type()),
+                &[x],
+            )?[0];
+        }
+
+        let w = model.add_const(format!("{prefix}.coefficients"), self.coefficients.clone())?;
+        let mut scores = {
+            model.wire_node(
+                format!("{prefix}.matmul"),
+                PrefixMatMul {
```
kali: Mmm... this will not play well with the optimiser. PrefixMatMul is meant to be an intermediate step in converting tract-core back to nnef or tflite, or to some optimisations.
tract-core favours and einsum reprensentation for matmul, as it gives the framework maximum flexibility for axis reordering.
The consequence is, this PrefixMatMul will not be optimised at all on cpu.
Fixing should not be too hard, look at this couple of lines for inspiration:
https://github.com/sonos/tract/blob/main/onnx/src/ops/math/gemm.rs#L61C1-L66C15

onnx/src/ops/ml/linear_classifier.rs
```diff
@@ -0,0 +1,320 @@
...
+                        },
+                        &logits2,
+                    )?
+                }
+            }
+        } else {
+            let mut tmp = scores.clone();
+            match self.post_transform {
+                None => {}
+                Some(PostTransform::Softmax) => {
```
kali: Mmm... instantiating Softmax is not easy :)
how about using `Softmax  { axes: tvec!(1), ..Softmax::default() }` to avoid exposing this code to this complexity ?

onnx/src/ops/ml/linear_regressor.rs
```diff
@@ -0,0 +1,164 @@
...
+                format!("{prefix}.to_f32"),
+                tract_core::ops::cast::cast(f32::datum_type()),
+                &[x],
+            )?[0];
+        }
+
+        let w = model.add_const(format!("{prefix}.coefficients"), self.coefficients.clone())?;
+        let mut y = model.wire_node(
+            format!("{prefix}.matmul"),
+            PrefixMatMul {
```
kali: Einsum again

onnx/Cargo.toml
```diff
@@ -24,18 +24,29 @@ log.workspace = true
 memmap2.workspace = true
 num-integer.workspace = true
 prost.workspace = true
+rand.workspace = true
+rayon.workspace = true
```
kali: same here, right ?

kali (changes requested):

PR #1938, opened 2026-01-08
Precompute more metadata during plan

threeseed:
For our use case we are executing basic ONNX models e.g. LinearClassifier millions of times.
And so the overhead of Tract is our biggest performance impact.

This PR aims to "front-load" more work during the plan stage.

Based on benchmarks we see a ~3x performance improvement.

### DIFF
--- a/core/src/plan.rs
+++ b/core/src/plan.rs
@@ -76,10 +76,17 @@ where
     M: Borrow<Graph<F, O>>,
 {
     model: M,
+    inputs: Vec<OutletId>,
+    input_facts: Vec<Option<TypedFact>>,
     outputs: Vec<OutletId>,
+    output_facts: Vec<Vec<Option<TypedFact>>>,
     order: Vec<usize>,
     flush_lists: Vec<TVec<usize>>,
     has_unresolved_symbols: bool,
+    node_has_symbolic_output: Vec<bool>,
+    single_symbolic_output_syms: Vec<Vec<Vec<Option<Symbol>>>>,
+    symbols: Vec<Symbol>,
+    symbolic_output_dims: Vec<Vec<Vec<usize>>>,
     executor: Option<Executor>,
     session_handler: Option<Arc<dyn SessionStateHandler + 'static>>,
     _casper: PhantomData<(F, O)>,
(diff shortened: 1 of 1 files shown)


kali:
Wow, ok, this one may be more tricky than the new ops in onnx/ml :)
 First question, can we split it into two PR to discuss the plan side and the symbols side separately or are they strongly coupled one way or the other ?

core/src/plan.rs
```diff
@@ -76,10 +76,17 @@ where
     M: Borrow<Graph<F, O>>,
 {
     model: M,
+    inputs: Vec<OutletId>,
+    input_facts: Vec<Option<TypedFact>>,
```
kali: Are these two denormalization worth it ? model.borrow().inputs should be pretty fast (input_facts, maybe less so). Why are input_facts optional ? is it only when they have symbols (in which case symbolic_input_facts may be a better name)

core/src/plan.rs
```diff
@@ -76,10 +76,17 @@ where
...
     outputs: Vec<OutletId>,
+    output_facts: Vec<Vec<Option<TypedFact>>>,
     order: Vec<usize>,
     flush_lists: Vec<TVec<usize>>,
     has_unresolved_symbols: bool,
+    node_has_symbolic_output: Vec<bool>,
+    single_symbolic_output_syms: Vec<Vec<Vec<Option<Symbol>>>>,
+    symbols: Vec<Symbol>,
+    symbolic_output_dims: Vec<Vec<Vec<usize>>>,
```
kali: then we have a bunch of per-node symbol resolution. Should we consider bunching them together in a struct SymbolResolutionNodeInfo or something like that ? then add a Vec<Option<SymbolResolutionNodeInfo>> in the Plan ? 
I think it would improve readability a bit, locality a tiny bit too, and we may have less indexing to do in the main loop (so less unsafe get_unchecked).
WDYT ?

Typically also, I tend to use Vec for long vectors (~ number of nodes) and TVec (actually from the smallvec crate) for small ones (~ number of node input/ouputs). I am not sure it has positive (or negative) performance implication here (a very sparse Vec<TVec<...>> could take a lot of place for nothing) , but it helps a bit in guessing what Vec<Vec<...>> could be.

core/src/plan.rs
```diff
@@ -151,19 +158,83 @@ where
...
+            let mut node_typed_facts: Vec<Option<TypedFact>> =
+                Vec::with_capacity(node.outputs.len());
+            let mut node_single_syms: Vec<Vec<Option<Symbol>>> =
+                Vec::with_capacity(node.outputs.len());
             for output in &node.outputs {
                 if let Ok(fact) = output.fact.to_typed_fact() {
-                    symbols.extend(fact.shape.iter().flat_map(|d| d.symbols()))
+                    symbols.extend(fact.shape.iter().flat_map(|d| d.symbols()));
+
+                    let dims_with_symbols = fact
```
kali: axes_with_symbols

core/src/plan.rs
```diff
@@ -151,19 +158,83 @@ where
...
+                    for d in fact.shape.iter() {
+                        let syms = d.symbols();
+                        if syms.len() == 1 {
+                            let s = syms.iter().next().unwrap().clone();
+                            per_dim_syms.push(Some(s));
+                        } else {
+                            per_dim_syms.push(None);
+                        }
+                    }
+                    node_single_syms.push(per_dim_syms);
```
kali: Thinking out loud here: I wonder if this will work in all cases. Multiple (and coupled) symbols can appear: LLM execution plans typically are parameterized by two symbols (past_seq_len and seq_len, shortened as P and S), and may appear conjointly. I /think/ on that example it would still work as P and S will appear separately in inputs before they are used together in expression.

So maybe looking for expression with single symbols in them is good enough indeed.

core/src/plan.rs
```diff
@@ -151,19 +158,83 @@ where
...
+            .iter()
+            .map(|outlet| {
+                model
+                    .borrow()
+                    .outlet_fact(*outlet)
+                    .ok()
+                    .and_then(|f| f.to_typed_fact().ok().map(|f| f.into_owned()))
+            })
+            .collect();
+
```
kali: ok, i think it get this now. you want a TypedFact, not whatever the Graph contains as fact.

core/src/plan.rs
```diff
@@ -400,18 +477,71 @@ where
...
+                                            )?;
+                                        }
+                                    }
+                                } else {
+                                    Self::resolve(
+                                        &mut self.session_state,
+                                        dim_abstract,
+                                        dim_concrete,
+                                    )?;
+                                }
```
kali: see my comments above about "predermined symbol resolution strategy".

kali:
Thanks @threeseed — really appreciate the work on this, and the LinearClassifier benchmark you added alongside, which was directly useful for what follows.

  Coming back to this after a while, I wanted to reproduce the ~3× number on current main before deciding how to proceed. A few findings worth sharing:

  On current main (post-bfd4485a, which fixed an independent 2× regression in resolve caused by to_i64 hitting the SymbolScope lock in the error path), I was able to reproduce ~1.2× / 19% wall-time reduction on your benchmark, not ~3×. A large part of your original delta was, I believe, bfd4485a-shaped: you were partly  measuring the fix for that regression, which existed in the tree you forked from. Apologies for that — it was a real bug on our side, independently spotted.

  On the PR's contents. I ran an ablation of each change against current main. Two ideas carry essentially all of the measured win:

  - The bare TDim::Sym fast path in resolve — skipping eval / solve_for / symbols() when the expression is literally a single unresolved symbol (−9.1% alone).
  - A syms_done latch in the eval loop — once every symbol in the plan is bound, skip the per-node "walk every output, walk every dim, call resolve" pass (−5.0% alone, −18.8% combined with the fast path).

  The other precompute (per-node symbolic-dim metadata, cached input/output facts, TVec::with_capacity) all measured as noise or dead weight once the syms-done latch was in place, on this workload at least.

  Action. I've put the benchmark fix + the two ideas above on #2153. Full credit to you in the commit messages. Closing this in favour of that branch.

  Thanks again for the contribution and for the patience.

PR #2001, opened 2026-02-28
Fix/wasm compat tract rs via inventory disabled

JulienBalianSonos:
# Discarded Reason

> Closing: the wasm-bindgen panic this PR was trying to work around is fixed upstream in wasm-bindgen >= 0.2.101 (confirmed via the torch-to-nnef vad-wasm example). Our repro setup now uses wasm-bindgen 0.2.118 + tract 0.23.0-dev.5 and builds cleanly on wasm32 without any patch to tract itself. No longer a tract-side issue.

# Features

- Gate the cudars/tract-cuda to windows & linux device only (merged in #2004  )
- Gate the use of `inventory` (and put it in default), with alternative transform discovery so that common wasm bundler like `wasm-pack` and it's underling **`wasm-bindgen`** works.

# Side Note

Inventory works in most wasm
If you build a .wasm including tract and load it:

- In Chrome
- In Wasmtime
- In Wasmer
- In WasmEdge

It will 'likely' work, because those engines:
- Grow memory properly
- Have proper stack discipline
- Handle indirect calls correctly

But the `wasm-bindgen` interpreter is intentionally lightweight and crash on this task (despite being 'broadly' adopted).


JulienBalianSonos:
Closing: the wasm-bindgen panic this PR was trying to work around is fixed upstream in wasm-bindgen >= 0.2.101 (confirmed via the torch-to-nnef vad-wasm example). Our repro setup now uses wasm-bindgen 0.2.118 + tract 0.23.0-dev.5 and builds cleanly on wasm32 without any patch to tract itself. No longer a tract-side issue.

PR #2136, opened 2026-04-15
Fix double-free in tract-proxy wrapper Clone impls

### LINKED ISSUE #2135: Structs in tract-proxy should not implement `Clone`
Hi,

your wrapper macro derives `Clone` on all rawptr-wrapping structs:

https://github.com/sonos/tract/blob/6d546999b52b2d1dd5cb8a07a79ec47b6b9fd54e/api/proxy/src/lib.rs#L24-L37

This breaks Rust's one-owner-at-a-time assumption and thereby will trigger double-frees when used:
```rs
fn main() {
    let nnef = nnef().unwrap();
    let clone = nnef.clone();
}
```
```
==4082124==ERROR: AddressSanitizer: attempting double-free on 0x7b41ec9e01d0 in thread T0:
```

If you want people to obtain clones, you either need to refcount the underlying object or perform a deep-copy in the `clone()` implementation.

Kind regards
Tim

kali:
The wrapper! macro derived Clone on raw-pointer-wrapping structs, causing double-free when a clone and its original were both dropped.

Add clone FFI functions (tract_tensor_clone, tract_fact_clone, tract_inference_fact_clone, tract_dim_clone) that perform deep copies, and use them in manual Clone impls instead of the derived bitwise copy. Remove Clone from wrapper types that have no clone support (Nnef, Onnx, Model, InferenceModel, Runnable, Runtime).

Also add proxy crate to the ASan CI test script.

Fixes #2135

### DIFF
--- a/.travis/asan.sh
+++ b/.travis/asan.sh
@@ -37,3 +37,13 @@ then
 fi
 ./.travis/cli-tests.sh
 
+if [ -n "$CI" ]
+then
+    cargo clean
+fi
+
+# Build libtract.so with asan, then run proxy tests against it
+cargo build -p tract-ffi $CARGO_EXTRA
+LIBTRACT_DIR=$(dirname $(find target -name 'libtract.so' | head -1))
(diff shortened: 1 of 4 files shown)


kali:
https://github.com/sonos/tract/actions/runs/24460196890/job/71471913864

PR #2160, opened 2026-04-27
cli: use jemalloc as the global allocator on unix

kali:
Switch tract-cli's global allocator from System (glibc malloc on Linux, the platform allocator on macOS/iOS) to jemalloc via tikv-jemallocator, wrapped through `readings_probe::wrap_global_allocator!` so the existing ALLOCATED/FREEED instrumentation that backs `info_usage` keeps working unchanged.

Why: jemalloc gives more predictable memory behaviour on long-running processes (less fragmentation on workloads that allocate/free a lot of small/medium tensors) and is roughly the same or faster for tract's short-lived CLI runs. Static link, so no runtime dependency.

Feature-gated as `jemalloc`, default on. Windows hits the `#[cfg(not(all(feature = "jemalloc", unix)))]` branch (tikv-jemallocator can't build there) and continues to use the System allocator. Disable explicitly with `--no-default-features` or `--features <subset>` if needed.

`nm tract | grep -c _rjem_` reports 517 symbols statically linked; `--readings -vv` still emits `Resource usage … alloc:N` lines.

### DIFF
--- a/.travis/cross.sh
+++ b/.travis/cross.sh
@@ -29,6 +29,9 @@ then
     PLATFORM=$1
 fi
 
+export JEMALLOC_SYS_WITH_MALLOC_CONF="retain:false"
+
+
 case "$PLATFORM" in
     "raspbian")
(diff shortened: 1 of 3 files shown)


kali:
used for benching only.

PR #2164, opened 2026-04-28
linalg/wasm: SIMD kernel kit (4x1, 8x1, 16x1, 8x8) + per-M dispatch

### LINKED ISSUE #2161: WASM SIMD: missing nr=1 kernel forces N=1 GEMV ops onto wasm_f32_4x4 (~50% of streaming-RNN frame, 75% column waste)
## Summary

Hi @kali — sharing one more piece of analysis from the DFN3 investigation, which lands on what looks like the highest-impact remaining lever for WASM streaming-inference: the WASM SIMD MMM kit currently has no `nr()==1` (GEMV-shaped) kernel registered, so every N=1 matmul is forced onto `wasm_f32_4x4` and wastes 75% of the column-tile work.

This is *not* a selection-logic bug — `kernel_selection::strategize()` already prefers `nr()==1` kernels for `op.n.is_one()` cases. It's a missing-kernel gap on WASM specifically.

I want to flag and offer to PR if this is a direction you'd like.

## Architecture (verified end-to-end against tract 0.23.0-dev.5 source)

```
tract-linalg/src/lib.rs::generic() registers default kernels for all targets:
    mmm_impls.push(generic_f32_4x4.mmm())     // MM kernel (4r×4c, scalar Rust)
    mmm_impls.push(generic_f32_4x1.mmm())     // GEMV kernel (4r×1c, scalar Rust)
    mmm_f32 = generic_f32_4x4
    mmv_f32 = generic_f32_4x1

tract-linalg/src/wasm.rs::plug() augments for WASM SIMD:
    ops.mmm_impls.push(wasm_f32_4x4.mmm())    // SIMD f32x4 MM kernel
    ops.mmm_f32 = wasm_f32_4x4                // override MM
    // mmv_f32 NOT overridden — no wasm_f32_4x1 exists
```

At model-optimize time, `tract-core/src/ops/einsum/kernel_selection.rs::strategize()`:

1. Filters `mmm_impls` to highest `quality()` (drops `Generic`-quality if any
   `TargetOptimized` exists)
2. For `op.n.is_one()` case: picks kernel with `nr() == 1` via `max_by_key`

On WASM after step 1, the only `TargetOptimized` kernel is `wasm_f32_4x4` (`nr() == 4`).
The scalar `generic_f32_4x1` (`nr() == 1`) has lower quality and gets filtered out.
So step 2 falls through to `wasm_f32_4x4` for both N=1 and N>1 cases.

Result: every N=1 GEMV computes a 4×4 tile and discards 3 of 4 columns. Per N=1
MMM call, 75% of f32x4 multiply-adds are wasted work.

## Hard data — DFN3 encoder, native (M1 AMX) profile

Same model on Apple M1 native shows tract's selection works correctly when a
`nr()==1` kernel exists at the target's quality level — AMX has both
`apple_amx_mmm_f32_32x1` and `apple_amx_mmm_f32_32x32` registered:

```
                                       m    k    n   picked (native AMX)
  ─────────────────────────────────────────────────────────────────────────
  N=1 (GEMV) ops, 7 nodes:
    /df_fc_emb/0/Einsum                16   96   1   apple_amx_mmm_f32_32x1
    /emb_gru/linear_out/0/Einsum       32   16   1   apple_amx_mmm_f32_32x1
    /emb_gru/linear_in/0/Einsum        16   32   1   apple_amx_mmm_f32_32x1
    /emb_gru/GRU.Xt_WrT                256  256  1   apple_amx_mmm_f32_32x1
    /emb_gru/GRU.Xt_WzT                256  256  1   apple_amx_mmm_f32_32x1
    /emb_gru/GRU.Xt_WhT                256  256  1   apple_amx_mmm_f32_32x1
    /lsnr_fc/0/MatMul                  1    512  1   apple_amx_mmm_f32_32x1
                                                     ─────────────────────
                                                     54.2 μs/iter, 52% of OptMatMul

  N>1 (MM) ops, 7 nodes:
    /df_conv0/4/Relu.low               96   64   64  apple_amx_mmm_f32_32x32
    /df_conv1/Relu.low                 64   64   48  apple_amx_mmm_f32_32x32
    /erb_conv1/3/Relu.low              64   64   16  apple_amx_mmm_f32_32x32
    /erb_conv*                         64   64   8   apple_amx_mmm_f32_32x32
    ...                                              ─────────────────────
                                                     50.8 μs/iter, 48% of OptMatMul
```

The pattern is structural to streaming-RNN models: per-frame inference produces
single time steps (S=1 after `PulsedModel::new(s, 1.to_dim())`), so any
fully-connected projection or GRU/LSTM gate is a `weight × activation` shape
where activation has a single column → N=1.

On WASM, all 7 of these GEMV ops are forced onto `wasm_f32_4x4` with 75% column
waste, so the GEMV fraction of WASM frame time is meaningfully higher than the
~52% the native AMX run shows.

## Proposed fix

Add `wasm_f32_4x1`: a WASM SIMD f32x4 GEMV-shaped kernel (4 rows × 1 col per
tile). Implementation parallels the existing `wasm_f32_4x4` kernel — same
`FusedKerSpec` match arms (Done, Clear, LoadTile, ScalarMul, AddMatMul, ...),
just collapsing the column dimension from 4 to 1 and using a single vector
accumulator instead of four.

Rough sketch:

```rust
unsafe fn kernel_f32_4x1(mut pnl: *const FusedKerSpec<f32>) -> isize {
    use std::arch::wasm32::*;
    unsafe {
        let mut ab = f32x4_splat(0.0);  // single accumulator: 4 rows × 1 col
        while !pnl.is_null() {
            match *pnl {
                FusedKerSpec::Done => break,
                FusedKerSpec::Clear => { ab = f32x4_splat(0.0); }
                FusedKerSpec::AddMatMul { k, pa, pb, packing: _ } => {
                    let a = pa as *const f32;       // 4 rows of A, packed per-K
                    let b = pb as *const f32;        // single column of B
                    for i in 0..k {
                        let a_row = v128_load(a.add(4 * i) as *const v128);
                        let b_scalar = f32x4_splat(*b.add(i));
                        ab = f32x4_add(ab, f32x4_mul(a_row, b_scalar));
                    }
                }
                FusedKerSpec::Store(tile) => {
                    let p = tile.ptr as *mut f32;
                    *p                                       = f32x4_extract_lane::<0>(ab);
                    *(p.byte_add(tile.row_byte_stride as _)) = f32x4_extract_lane::<1>(ab);
                    *(p.byte_add(tile.row_byte_stride as _ * 2)) = f32x4_extract_lane::<2>(ab);
                    *(p.byte_add(tile.row_byte_stride as _ * 3)) = f32x4_extract_lane::<3>(ab);
                }
                // ScalarMin/Max/Add/Sub/Mul, PerRow*, PerCol*, AddUnicast, etc.
                // — same as 4x4 but operate on single accumulator vector
                ...
            }
            pnl = pnl.add(1);
        }
    }
    0
}

MMMRustKernel!(kernel_f32_4x1 => wasm_f32_4x1<f32>(4,1)@(4,1)
               quality(ImplementationQuality::TargetOptimized));
```

And in `wasm.rs::plug()`:

```rust
ops.mmm_impls.push(wasm_f32_4x1.mmm());      // new
ops.mmv_f32 = Box::new(|_m, _k| wasm_f32_4x1.mmm());  // new (was generic fallback)
```

The selection logic at `kernel_selection.rs:51-54` already prefers `nr()==1`
when `op.n.is_one()`, so no changes to `tract-core` needed — once
`wasm_f32_4x1` is registered with `TargetOptimized` quality, it's automatically
picked for N=1 cases.

## Measured impact (implementation prototyped + benchmarked locally)

**Verified end-to-end on a patched tract-linalg fork.** Apples-to-apples A/B
on the DFN3 encoder + decoder pipeline running in Node 20 WASM (M1 Pro):

```
                         RTF best   per-frame mean   wasmSize
simd128 + 4x4 only       0.1077     1.077 ms          9.51 MiB    (baseline)
simd128 + 4x4 + 4x1      0.0902     0.901 ms          9.51 MiB    (patched)
                         ──────     ────────          ────────
Δ                        -0.0175    -0.176 ms         +461 B
%                        -16.2%     -16.3%            +0.005%
```

**Alternating 5-cycle A/B confirms reproducibility:**

```
Cycle  baseline  patched  Δ%
  1    0.1102    0.0904   -17.97%
  2    0.1114    0.0905   -18.76%
  3    0.1107    0.0910   -17.80%
  4    0.1109    0.0910   -17.94%
  5    0.1084    0.0945   -12.82%
                         ───────
                         mean: -17.06%
```

**Audio output bit-identity preserved:**

```
SHA256 baseline (simd128 + 4x4 only):  18cd91ad66785054c30fcbdfca072822ef4374fb5063e8ca3961d7deda5f2298
SHA256 patched  (simd128 + 4x4 + 4x1): 18cd91ad66785054c30fcbdfca072822ef4374fb5063e8ca3961d7deda5f2298
                                       ────────────────────────────────────────────────────────────────
                                       0 differing bytes / 1,309,932 samples (27.29s of audio)
```

This was at the upper end of my prior estimate range (8-15%). Streaming LSTM
workloads with 4 gates instead of 3 should see proportionally larger gains.

For the broader streaming-inference audience on WASM (RNN-T, streaming LSTM
acoustic models, time-series forecasters, RNNoise variants, FullSubNet
streaming variants, audio enhancement) — anyone whose model has FC layers or
gated recurrences feels the same gap.

## Backward compatibility & test status

**The implementation already passes the existing tract CI test suite.**

The `MMMRustKernel!` macro auto-generates 54 tests for the new kernel
(matching the coverage `wasm_f32_4x4` already gets) — every `FusedKerSpec`
arm, frame-level matmul, property-based random matrix configurations, and
output-storage variants. Running with the same invocation
`.travis/cross.sh` uses for the `wasm32-wasi` matrix entry:

```
$ RUSTFLAGS='-C target-feature=+simd128' \
  CARGO_TARGET_WASM32_WASIP1_RUNNER=wasmtime \
  cargo test --release --target=wasm32-wasip1 -p tract-linalg

test result: ok. 1508 passed; 0 failed; 0 ignored; 0 measured
  ├─ wasm_f32_4x4:  54 tests (existing baseline)
  └─ wasm_f32_4x1:  54 tests (new — same coverage scope, all green)
```

Bit-identity to scalar `generic_f32_4x1` is also verified end-to-end on real
audio (the SHA256 match shown above) — kernel produces the same lane-by-lane
output as the existing scalar fallback, modulo nothing.

## Reproducibility

Full per-op profile + kernel-selection trace was generated by a Rust binary
that walks the optimized PulsedModel and dumps `OptMatMul.info()` for every
matmul node. Source available — happy to share if you want to repro the table
above on your side, or to extend the tooling for any other ONNX you'd like to
profile the kit's behavior on. The binary uses tract's public `Op::info()`
API, no internal hackery.

## Ask

The implementation is **already prototyped and validated** — single-file
diff against `main` (153 lines added to `linalg/src/wasm.rs`), no changes to
`tract-core` selection logic, all 1508 `tract-linalg` tests passing on
`wasm32-wasip1` + wasmtime with `+simd128`.

Three options:

1. **Stay where we are** if there's a reason the WASM kit was held to MM-only
   that I'm missing (testing surface, kernel-count budget, etc.) — happy to
   just have the data on the record.
2. **You take the implementation** as-is if you'd prefer to commit it
   yourself to keep authorship contained — happy to share the patch via gist
   or link it from this issue.
3. **I open a PR** from a fork. The branch already exists locally with a
   clean single-commit diff against `main`. Would land within minutes of you
   green-lighting the direction.

No urgency. We're targeting on 0.22.1 today and will revisit 0.23 once #2157
lands; this would be one more line item to evaluate at that point. But it's
the largest perf lever we found in our investigation, and the fix is bounded
enough — single new kernel, mirrors the pattern already established on every
other arch — that it seemed worth surfacing as its own issue.

Thanks again for the engagement on the prior threads.

— CK

czoli1976:
Closes #2161.

Provides MM, MV, and a reasonable choice of tile sizes for `linalg/src/wasm.rs`:

- **MV (nr=1)** — `wasm_f32_4x1`, `wasm_f32_8x1`, `wasm_f32_16x1`
- **MM (nr>1)** — `wasm_f32_8x8` (alongside existing `wasm_f32_4x4`)

Selector wiring in `wasm::plug()`:
- `ops.mmm_f32` → `wasm_f32_8x8`
- `ops.mmv_f32` → `4x1` / `8x1` / `16x1` dispatched by `m`

`MMMRustKernel!` auto-generates 54 tests per kernel; tract-linalg suite passes on `wasm32-wasip1` + wasmtime + `+simd128`.

DFN3 end-to-end: RTF 0.108 → 0.0516, audio bit-identical to scalar reference.


kali (changes requested):
I prefer sticking to rust default in terms of warnings / linting, so I'd prefer we go the other way araound and make as unsafe {} the function bodies (as it was on the 4x4).

czoli1976:
Done as you asked, @kali — pushed [d925624](https://github.com/sonos/tract/pull/2164/commits/d925624) which wraps each kernel body in `unsafe { ... }` matching the existing `kernel_f32_4x4` pattern. Lint-allow approach reverted (force-pushed). `cargo fmt` clean, no warnings.

kali:
here we go ! thanks !

czoli1976:
thank you very much !

Issue #2156, opened 2026-04-24
Build fails with unknown argument '-mavx512pf'


mtb0x1:
Hello,

This is not an issue for tract per se, but more a tracking issue for what's to come.

Trying to build tract on my setup yields the following error : 
- target : x86_64 linux
- compiler : clang 21
- error : 
```log
    return re.sub( "/\*.*?\*/", "", string, flags=re.S )
 .../tract/target/debug/build/blis-src-18d48980233e6c9a/out/blis_x86_64-unknown-linux-gnu/build/flatten-headers.py:530: SyntaxWarning: invalid escape sequence '\s'
    regex = re.compile( '^[\s]*#include (["<])([\w\.\-/]*)([">])' )
   clang: fatal error: unknown argument '-mavx512pf'; did you mean '-mavx512f'?
  make: *** [Makefile:726: obj/x86_64/kernels/knl/1m/bli_dpackm_knl_asm_24x8.o] Error 1

  thread 'main' (227138) panicked at .../.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/blis-src-0.2.2/build.rs:111:5:
  assertion failed: command.status().unwrap().success()
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
``` 
Tracking down the issue, leads to `blas-lapack-rs/blis-src` and it's dependency on `flame/blis`  : 
 - `blas-lapack-rs/blis-src` v0.2 relies on `flame/blis` (tag 1.0)
 - `flame/blis` (tag 1.0) tries to build for a set of CPU arch, among of them knl (Xeon Phi)
 - Xeon Phi as arch is EOL since Aug. 2017 
 - Major compilers GCC and Clang removed support for it. (https://www.phoronix.com/news/LLVM-19-Drops-Xeon-Phi) : (clang 19+, gcc 15+)
 - GCC 15 and Clang 19 are already 2y+.

I couldn't find other way than downgrading compiler to a supported version with these dependencies and force compiler  (`CC=gcc/clang cargo build`).

 Digging through bliss commits, they have identified these issues in (https://github.com/flame/blis/commit/071dfce5f7b1100bedbe2e1704b0f7c65639cce4 and https://github.com/flame/blis/commit/7e8a5891902312a281bce37037eaa06d7d501639).

moving forward, I guess either stick with `blas-lapack-rs/blis-src` v0.2 and pin compilers to right version. or : 
[...]

PR #2175, opened 2026-04-30
Drop the BLAS / SGemm integration

kali:
The original target was the Rockchip NPU's CBLAS-compatible sgemm shim, which never made it into a maintained deployment path. What remained was a narrow opt-in transform (as_blas) that rewrote f32, no-transpose, non-quantized matmuls onto cblas::sgemm — exactly the shape tract-linalg already covers efficiently. The integration carried four optional system deps (cblas, accelerate-src, blis-src, openblas-src), four feature flags, a dedicated test-rt runtime excluded from default CI, and a CBLAS arm in matmul-bench, none of which earned their cost.

Removes the AsBlas transform, SGemm op, the four feature flags from tract-core and matmul-bench, the test-rt/test-blas crate, and the windows.yml --exclude that worked around it.


kali:
Closes #2156

PR #2201, opened 2026-05-07
tdim: PartialEq second-chance via diff-and-simplify

JulienBalianSonos:
Two algebraically equal TDims can arrive at different canonical forms via different construction paths. For example, the GCD-factoring simplifier rewrites `16T+16` (`Add([MulInt(16, T), Val(16)])`) to `16(T+1)` (`MulInt(16, Add([T, 1]))`) on one path while another keeps the Add form. Structural `PartialEq` returns false; `Broadcast::simplify`'s dedup keeps both, and downstream Reshape volume checks emit `(16T+16)#(16T+1)` which can't reduce.

Plug a second-chance into `PartialEq`: if structural equality fails, try `(self - other).simplify() == Val(0)`.

## Guards

- Thread-local `EQ_GUARD` prevents the simplifier's internal `HashMap<TDim, _>` from re-entering `eq` on each insertion (would recurse without bound on non-structurally-equal inputs).
- Skip second-chance when either side is a `Val` or `Sym` leaf. For `Val(c)` vs anything non-`Val`, simplify should already have folded the other side; running a diff-and-simplify here just risks arithmetic overflow on extreme constants. For `Sym` leaves, assertion-driven equality belongs in `simplify`, not in `eq`.

## Surfaced by

Pocket-TTS pulsification (LayerNorm-after-ConvTranspose with dynamic axis): the variance-branch dim ended up in MulInt-factored form while the residual dim kept the Add form, broadcast couldn't dedup, downstream pulse-divisibility check bailed on the Reshape's symbolic shape.

## Tests

- `eq_add_vs_mulint_factored`: `Add([MulInt(16, A), Val(16)])` == `MulInt(16, Add([A, Val(1)]))` returns true.
- `broadcast_dedup_two_canonical_forms`: `Broadcast([add_form, mul_form]).simplify()` collapses to a single term.

### DIFF
--- a/data/src/dim/tree.rs
+++ b/data/src/dim/tree.rs
@@ -30,7 +30,7 @@ impl std::error::Error for TooEarly {}
 
 macro_rules! b( ($e:expr) => { Box::new($e) } );
 
-#[derive(Clone, PartialEq, Eq, Hash, Debug)]
+#[derive(Clone, Eq, Hash, Debug)]
 pub enum TDim {
     Val(i64),
     Sym(Symbol),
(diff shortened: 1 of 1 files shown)


JulienBalianSonos:
Superseded by 347f42b19. The same PartialEq second-chance change landed on main on 2026-05-06, picked up from an earlier draft. This PR is redundant.

PR #2202, opened 2026-05-07
pulse: LCM-merge stream-axis dims + slope-based per-pulse for Range

JulienBalianSonos:
When two parallel pulse paths converge at an elementwise op, the merged tensor's stream-axis dim is the **Least Common Multiple (LCM)** of the inputs' per-pulse stream dims, not the NumPy-style broadcast. Two semantics get conflated otherwise:

- Tensor shape compatibility (must match at runtime): non-stream axes use NumPy `Broadcast`. Equal or one is 1.
- Pulse-divisibility (a scalar constraint on per-pulse cycle): on the stream axis, two paths with steady-state size `K_a` and `K_b` are compatible at any pulse that is a multiple of `LCM(K_a, K_b)`.

`PulseWrappingOp::pulsed_output_facts` was using the typed op's `output_facts` for shape merging, which falls through to `multi_broadcast` and produces `Broadcast([K_a, K_b])` on the stream axis when paths produce different per-pulse sizes (e.g. ConvTranspose with kernel > stride at steady-state vs initial-state phases). Downstream pulse-divisibility checks then bail.

## Changes

- New `TDim::pulse_lcm` helper (pure-integer LCM; returns `None` for symbolic, caller falls back).
- `PulseWrappingOp::pulsed_output_facts` post-processes typed `output_facts`: when the stream-axis dim is a `Broadcast`, replace with LCM of all stream-bearing inputs' stream dims. Non-stream axes keep `Broadcast` semantics (correct for runtime shape compatibility).
- `Range::pulsify` (`pulse/src/ops/array/range.rs`): replace `stream_dim.substitute(symbol, pulse).to_usize()` (full evaluated length, includes constant tail from upstream overlap-add) with slope-based `guess_slope(symbol).0 × pulse_int` (steady-state increment per pulse). The substitute form was forcing per-pulse to the kernel-state size when downstream paths were producing the steady-state size, surfacing exactly as the `Broadcast(stride, kernel)` we were merging.
- `TDim::guess_slope` made `pub` (was `pub(super)`) so pulse op-pulsifiers outside `tract-data` can use it.

## Module-level doc

`pulse/src/lib.rs` now documents the stream-axis vs non-stream-axis merge contract so future maintainers don't reintroduce the conflation.

## Surfaced by

Pocket-TTS Mimi audio decode (`upsample[stride=16, kernel=32] → linear → view → unbind → arange-based mask → SDPA → linear → conv`). Pre-fix: failed at `Pulsification requires pulse Broadcast(16, 32) to be a stride (1) multiple`. Post-fix: pulsifies cleanly through optimize at pulse=1, 2, 4, 8.

## Depends on

[#2201](https://github.com/sonos/tract/pull/2201) (`tdim: PartialEq second-chance`). Needed for the LayerNorm-after-ConvTranspose case where the GCD-factoring rewrite produces algebraically-equal-but-structurally-different canonical forms.

## Tests

- `pulse_lcm_basic` in `tract-data`.
- `test_pulse_meet_with_arange_branch_types_through` in `tract-pulse`: minimal repro of the upsample-then-arange-mask pattern; pulsifies cleanly post-fix.
- 132 + 34 lib tests green; 55 `core-proptest-pulse` tests green (no regressions in existing pad+conv, deconv, conv+conv, etc. proptests).

## Known limitations / follow-ups

- `pulse_lcm` returns `None` for symbolic operands; callers fall back to existing `Broadcast` behavior (still safe). A symbolic LCM is worth following up.
- Op-specific pulsifiers (cnn::conv, deconv, fft, etc.) that compute stream-axis sizes in their own pulsify functions don't go through `PulseWrappingOp`; they should follow the same slope-based convention. This PR fixes Range; quick audit of others recommended.
- A pre-existing tract bug exists in `Deconv` pulse runtime semantics (output diverges at chunk boundaries for kernel > stride). Exposed by `tract compare --stream` once the type-level fix in this PR is in. Filed as [#2203](https://github.com/sonos/tract/issues/2203).

### DIFF
--- a/data/src/dim/tree.rs
+++ b/data/src/dim/tree.rs
@@ -1236,6 +1236,27 @@ impl TDim {
         self.clone().neg().prove_strict_positive()
     }
 
+    /// Pulse-divisibility merge: smallest scalar dim that is a multiple of
+    /// both `self` and `other`.  Used at pulse meet-points (elementwise
+    /// ops where parallel paths produced different per-pulse sizes — e.g.
+    /// stride and kernel of a ConvTranspose surfacing as `(K_steady,
+    /// K_initial)` on the two phases of the pulse cycle).
+    ///
+    /// For pure positive integers, returns the exact LCM as `Val(.)`.
+    /// For symbolic TDims, returns `None` -- the caller can fall back to
+    /// the existing `Broadcast` semantics or use a safe upper bound.
+    pub fn pulse_lcm(&self, other: &TDim) -> Option<TDim> {
+        match (self.as_i64(), other.as_i64()) {
+            (Some(a), Some(b)) if a > 0 && b > 0 => {
+                let g = (a as u64).gcd(&(b as u64));
+                let l = (a as u64 / g).saturating_mul(b as u64);
+                if l > i64::MAX as u64 { None } else { Some(TDim::Val(l as i64)) }
+            }
+            (Some(0), _) | (_, Some(0)) => Some(TDim::Val(0)),
+            _ => None,
+        }
+    }
+
     pub fn gcd(&self) -> u64 {
         use self::TDim::*;
         match self {
@@ -1291,7 +1312,7 @@ impl TDim {
         TDim::Div(Box::new(Add(vec![self, Val(rhs as i64 - 1)])), rhs).reduce()
     }
 
-    pub(super) fn guess_slope(&self, sym: &Symbol) -> (i64, u64) {
+    pub fn guess_slope(&self, sym: &Symbol) -> (i64, u64) {
         fn slope_rec(d: &TDim, sym: &Symbol) -> (i64, i64) {
             match d {
                 Val(_) => (0, 1),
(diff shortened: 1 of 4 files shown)


data/src/dim/tree.rs
```diff
@@ -1236,6 +1236,27 @@ impl TDim {
...
+            (Some(a), Some(b)) if a > 0 && b > 0 => {
+                let g = (a as u64).gcd(&(b as u64));
+                let l = (a as u64 / g).saturating_mul(b as u64);
+                if l > i64::MAX as u64 { None } else { Some(TDim::Val(l as i64)) }
+            }
+            (Some(0), _) | (_, Some(0)) => Some(TDim::Val(0)),
+            _ => None,
+        }
+    }
+
```
kali, line 1257: just a terminology nitpick, but unless i'm mistaken, this is just a LCM, with no relation to pulsing.

pulse/src/lib.rs
```diff
@@ -1,3 +1,32 @@
...
+//! kernel > stride at steady-state vs initial-state phases). The merge
+//! lives in [`crate::model::stream_axis_lcm`] called from
+//! `PulseWrappingOp::pulsed_output_facts`.
+//!
+//! Op-specific pulsifiers (cnn, range, deconv, …) compute stream-axis
+//! sizes in their own pulsify functions; they should follow the same
+//! "slope × pulse" convention -- the *coefficient* of the streaming
+//! symbol times the pulse value, NOT `stream_dim.substitute(symbol,
+//! pulse)` (which includes any constant tail produced by overlap-add).
+
```
kali: I don't think we want this here, the merging problem is part of the pulsification process but def not the core story. I agree documenting the pulsification is a need, but this is not the right anwsr.

pulse/src/model.rs
```diff
@@ -10,6 +10,23 @@ use tract_pulse_opl::tract_core::ops::source::TypedSource;
...
+/// that `multi_broadcast` produced.
+pub fn stream_axis_lcm(inputs: &[&PulsedFact]) -> Option<TDim> {
+    let dims: Vec<&TDim> =
+        inputs.iter().filter_map(|f| f.stream.as_ref().map(|s| &f.shape[s.axis])).collect();
+    let (first, rest) = dims.split_first()?;
+    let mut acc = (*first).clone();
+    for d in rest {
+        acc = acc.pulse_lcm(d)?;
+    }
+    Some(acc)
```
kali: missed opportunity for a fold() ?

kali (changes requested):

data/src/dim/tree.rs
JulienBalianSonos: renamed to just lcm

pulse/src/lib.rs
JulienBalianSonos: Dropped the module-level block entirely. The merge-semantics specifics are now on stream_axis_lcm where the mechanism actually lives.

pulse/src/model.rs
JulienBalianSonos: Done, with try_fold: drops the Vec collect and the mutable accumulator;

kali (approved):

PR #2209, opened 2026-05-09
core/cnn/depth_wise: dispatch N=1, 2, 3 zones to process_zone_n_generic (−7% on DFN3)

czoli1976:
> Apologies for the Saturday timing — please review whenever convenient, no rush.

## Summary

The match in `process_zone_$suffix` only routed `values_offsets.len() == 4` to the optimised `process_zone_n_generic::<T, N, 4>` template; every other zone size fell through to `visit_output(...inner_loop...)` which calls `inner_loop_generic` once per (output_point, channel) and re-creates the `valid_offsets_ker_in()` iterator each time.

`process_zone_n_generic` is already generic over `const N` and works for any tap count. Just adding the N=1, 2, 3 arms (alongside the existing N=4 arm) reuses existing code rather than adding new code:

```rust
match zone.values_offsets.len() {
    1 => process_zone_n_generic::<T, 1, 4>(...),
    2 => process_zone_n_generic::<T, 2, 4>(...),
    3 => process_zone_n_generic::<T, 3, 4>(...),
    4 => process_zone_n_generic::<T, 4, 4>(...),  // unchanged
    _ => zone.visit_output(...),                   // unchanged fallback
}
```

The diff also drops a stale `/* if zone.values_offsets.len() == 2 ... */` comment that was the original (unfinished) attempt at this dispatch.

## Why this matters for DFN3

I instrumented `process_zone_$suffix` to count zones per submodel during one bench iter:

| submodel | n=1 | n=2 | n=3 | n=4 |
|---|---|---|---|---|
| enc | 0% | 56% | 44% | **0%** |
| erb_dec | 57% | 28% | 14% | <1% |
| df_dec | (no DepthWiseConv) | | | |

So on DFN3, **~99% of zone visits previously fell through to the slow path**. Particularly enc never hit the optimised path at all because all its DWConv kernels produce 2- or 3-tap zones.

## Bench

M1 Pro, `RAYON_NUM_THREADS=1`, T=100 input, 50 outer × `tract bench --warmup-time 500 --max-time 2000`, alternating baseline/fix runs to control thermal drift. df_dec uses single-run dwconv2 (the dwconv1 run had a thermal outlier with max=12.58 ms; dwconv2 ran clean):

| submodel | base mean (n=83-100) | fix mean | Δ mean | Mann-Whitney p |
|---|---|---|---|---|
| `enc.onnx` | 9.526 ms | 8.339 ms | **−12.46%** | 3.7×10⁻³¹ |
| `erb_dec.onnx` | 9.558 ms | 8.631 ms | **−9.69%** | 1.6×10⁻³² |
| `df_dec.onnx` | 10.980 ms | 10.978 ms | −0.01% | 0.31 (no signal — no DWConv ops in df_dec) |

**DFN3 sum: 30.06 ms → 27.95 ms = −7.0%**.

Drift between two independent fix runs (same code, ~14 min apart): <4% on mean, well below the per-run noise — the gain is not thermal.

## Test plan

- [x] `cargo test --release -p tract-core --lib` — **237 / 237 pass**, 0 failures
- [x] `cargo test --release -p tract-linalg --lib` — **3524 / 3524 pass**, 0 failures
- [x] **Bit-exact DFN3 output verification**: ran patched tract on enc/erb_dec/df_dec reference NPZ inputs, compared element-wise to baseline tract — `max_abs_diff = 0` across all 9 output tensors. Output is byte-identical.
- [x] `cargo fmt --check` clean
- [x] `cargo build --release -p tract-cli` clean

## Why this works

`process_zone_n_generic<T, N, UNROLL>` hoists per-zone setup (kernel weight loads, bias load, ZoneScanner setup) **out of the per-output-point inner loop**. The fallback path called `inner_loop_generic` separately for every output point, redoing the iterator construction each time. The fast path keeps the per-zone state alive and unrolls UNROLL=4 output points at a time.

For the predominant DFN3 zone sizes (n=1, 2, 3), this is a strict win — same correctness, much better instruction-cache behaviour and clearer auto-vectorisation opportunity on the inner unroll. The fallback is preserved for n>4 (e.g., 9 for 3×3 conv) where the existing slow path still handles correctness.

## Predicted beneficiaries beyond DFN3

- **Audio streaming graphs** with 1-D temporal convs (k=3 or k=5) — same pattern as DFN3 and similar relative gain
- **Wake-word / KWS RNNs** with small temporal context (mostly k=2 or k=3)
- **Tiny CNN classifiers** (gesture, accelerometer-based) where DWConv with small kernels dominates compute
- Vision DWConv at edges (corner zones with n=1, n=2 valid taps) — partial gain on the boundary

## Caveats

- Adds three new monomorphisations of `process_zone_n_generic` per datum type (in DFN3 we already have one — N=4). Binary size: +98 KB on the M1 Pro release build (28.36 → 28.46 MB). Acceptable for the gain.
- For models where DWConv is rare or zone sizes don't fall into N=1..4, this PR is a no-op.

### DIFF
--- a/core/src/ops/cnn/conv/depth_wise.rs
+++ b/core/src/ops/cnn/conv/depth_wise.rs
@@ -151,29 +151,27 @@ macro_rules! impl_eval {
                    zone, c_stride_i, c_stride_o, k_stride_i, iptr, kptr, bias, optr,
                    )
                    } else */
-                if zone.values_offsets.len() == 4 {
-                    [<process_zone_n_ $suffix>]::<T, 4, 4>(
+                match zone.values_offsets.len() {
+                    1 => [<process_zone_n_ $suffix>]::<T, 1, 4>(
                         dw, zone, c_stride_i, c_stride_o, k_stride_i, iptr, kptr, bias, optr, add, mul,
-                        )
-                        /*
-                           } else if zone.values_offsets.len() == 5 {
-                           dw.process_zone_n::<T, 5, 2>(
-                           zone, c_stride_i, c_stride_o, k_stride_i, iptr, kptr, bias, optr,
-                           )
-                           } else if zone.values_offsets.len() == 9 {
-                           dw.process_zone_n::<T, 9, 1>(
-                           zone, c_stride_i, c_stride_o, k_stride_i, iptr, kptr, bias, optr,
-                           )
-                           */
-                } else {
-                    zone.visit_output(&dw.patch, |visitor| {
+                    ),
+                    2 => [<process_zone_n_ $suffix>]::<T, 2, 4>(
+                        dw, zone, c_stride_i, c_stride_o, k_stride_i, iptr, kptr, bias, optr, add, mul,
+                    ),
+                    3 => [<process_zone_n_ $suffix>]::<T, 3, 4>(
+                        dw, zone, c_stride_i, c_stride_o, k_stride_i, iptr, kptr, bias, optr, add, mul,
+                    ),
+                    4 => [<process_zone_n_ $suffix>]::<T, 4, 4>(
+                        dw, zone, c_stride_i, c_stride_o, k_stride_i, iptr, kptr, bias, optr, add, mul,
+                    ),
+                    _ => zone.visit_output(&dw.patch, |visitor| {
                         for c in 0..*dw.input_shape.c() as isize {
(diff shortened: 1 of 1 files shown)


kali:
> Apologies for the Saturday timing — please review whenever convenient, no rush.

:+1:

kali (approved):

PR #2215, opened 2026-05-12
tract-coreml: draft Apple Core ML / ANE backend (curiosity-driven, looking for direction)

czoli1976:
## Why this PR exists (the honest framing)

I started this out of curiosity — I'm honestly not sure how broad the value is to the wider tract community. My obsession is **DFN3** (DeepFilterNet 3) running on Apple Silicon. From there I ended up probing a bunch of other models that may or may not be normal tract usage, and along the way the work unveiled a number of things about tract itself that look like upstream improvements worth surfacing regardless of whether this backend ever lands.

Filing as **draft** to get direction on:

1. Is this kind of contribution interesting to tract? (Or would you rather it live as a separate crate outside the workspace?)
2. Of the upstream improvements this surfaced — which ones are worth me opening as their own PRs?
3. Are the models I tested representative of how people use tract, or am I optimising for an edge case?

Honest about what I don't know.

## What this is

A new `tract-coreml` crate that bridges tract's typed-model subgraphs to Apple Core ML via MLProgram / MIL, so models can dispatch to the GPU and the Neural Engine (ANE). Sits next to `tract-cuda` / `tract-metal` in the workspace; opt-in (not in `default-members`); MLProgram-only (no NeuralNetwork format); MIT-OR-Apache-2.0.

It also has **one small standalone upstream fix** in `tract-onnx-opl` (Resize accepting F16 input — see commit 1; described separately below in case you'd prefer that as its own PR).

## What got measured

Six end-to-end **tract CPU vs CoreML** speedups (M1 Pro, release, FP16):

| Class | Model | Tract CPU | CoreML | Speedup | ANE? |
|---|---|---:|---:|---:|:---:|
| CNN classifier | MobileNet v2 (224²) | 425 ms | **2.22 ms** | **191×** | ✅ 185 mW sustained |
| CNN classifier | SqueezeNet 1.x (224²) | 23 ms | **0.46 ms** | **50×** | ❌ too small |
| CNN classifier | InceptionV3 (299²) | 419 ms | **12.18 ms** | **34×** | ✅ (2.76× CPU+ANE vs CPU-only) |
| Segmentation | MODNet (512²) | 2,434 ms | **10.11 ms** | **241×** | ✅ 200–600 mW |
| Recurrent matting | RVM (480×640) | 503 ms | **4.53 ms** | **111×** | ❌ recurrent loop rejected by scheduler |
| Super-resolution | Real-ESRGAN x4 | 1,670 ms | **131 ms** | **12.7×** | (no ANE; CPU residuals from upsample chain) |
| **Audio (the one I care about)** | **DFN3 decoder (1s chunk)** | **75.3 ms** | **71.4 ms** | **1.06×** | ❌ Scan-CPU fragments graph |

For full per-canary detail, op histograms, stranded ops, and ANE engagement notes, see the attached `benchmarks.pdf`.

## The DFN3 row is the interesting one

DFN3 is dominated by 2× `tract_core_scan` loops containing GRU cells. Those loops stay on CPU because MIL has no `scan` op — only `while_loop`, which would require either a backend-side pattern detector or upstream RNN preservation. What CoreML actually accelerates today on DFN3 is just the *outside-scan* compute (~5% of wall time). That's why the speedup is 1.06×.

This empirically quantifies the cost of the Scan-CPU boundary. If the GRU body lifted to MIL `while_loop`, the expected outcome is 5–15× speedup (CoreML drops from 71 ms to 5–15 ms; CPU stays at 75 ms). 5–15× of latent speedup is currently locked behind that boundary.

The "right" architectural fix is upstream — preserving GRU/LSTM as first-class typed ops in tract IR so backends can translate them trivially. **I sketched a proposal for this in [`tract-rnn-preservation-handout.md`](coreml/docs/tract-rnn-preservation-handout.md) included in the PR**; happy to file it as a separate issue/PR if there's interest. It would benefit not just CoreML but also tract-metal (MPS GRU), tract-cuda (cuDNN RNN), and any future tract-linalg fast-path.

### Commit 1 — `onnx-opl/resize: accept F16 input (cast to F32 internally)`

~10 LoC + 2 unit tests in `tract/onnx-opl/src/resize.rs`. Cast at the boundary so Resize::eval can run F16 models on the tract-CPU runtime. **This is the most clearly separable piece** — if you want it as its own PR, I'm happy to extract it. Without it, none of the segmentation / SR / detection models in the table above had a tract-CPU baseline through the FP16 pipeline that CoreML requires.

Behaviour for F32 models is unchanged (it's a fast-path skip; no extra allocation).

### Commit 2 — `coreml: add tract-coreml backend crate`

The new crate. Highlights:

- **Subgraph fusion**: union-find with all-or-nothing + convex-region rules, so each connected component of translatable nodes becomes one MLPackage / one `CoremlOp` node. Designed for ANE engagement (Apple's scheduler decides per-MLPackage).
- **28 op-family translators** across the `coreml/src/ops/` directory: Conv, MaxPool, AvgPool (via SumPool detection), Pad, BinOp (with Const absorption + Eq), Activation dispatcher (~17 fn variants including LeakyRelu / Cos / Sin / Neg / Square / Recip / Ln), InstanceNorm + LayerNorm multi-node folds, RmsNorm standalone, Softmax, MatMul (4 patterns) + general_matmul (multi-K/M/N + strip + expand + batch-reorder + rank-6 boundary strip), einsum_outer (no-K outer-product), Reshape, MoveAxis, AddAxis, RmAxis, Slice, Concat, Broadcast, Resize, Cast, Reduce, EinSum-as-1x1-conv, Gather, ScatterNd, Iff (3-input select).
- **External-shape boundary mechanism**: translators expose `*_external_shape` fields so tensors with rank > 5 (Hiera windowed-attention 6D forms with declutter-inserted decorator axes) can flow through MIL's rank-5 cap by squeeze/unsqueeze at the MLPackage I/O boundary.
- **Pure-Rust runtime MIL build** — no Python `coremltools` dependency, no system `protoc`. Schema vendored from `apple/coremltools` 9.0; codegen via `prost-build` + `protox`.
- **Persistent compile cache** in `~/Library/Caches/tract-coreml/v1/<sha256>.mlmodelc/` with canonical proto walking (handles `prost::encode` HashMap iteration non-determinism). ~8× warm-load speedup.
- **No `tract-gpu` dependency** — Core ML's MLMultiArray ownership and per-prediction memory model don't map onto the gpu shared abstractions cleanly; tract-coreml is self-contained.
- **57 in-CI tests** (lib + smoke + Conv→Add fusion); **43 ignore-by-default e2e tests** per real model.

Code size: ~11.4k lines of hand-written Rust in `src/` + ~6.7k lines of tests. Roughly the same neighborhood as cuda + metal once you add their shared `gpu/` crate (cuda+gpu/2 ≈ 10.1k, metal+gpu/2 ≈ 10.4k). About ~37% of `src/` is defensive code that's largely unique to CoreML — see the upstream-improvements section for context, several items would shrink that materially.

For the full per-translator → MIL-op mapping, see attached `ops-coverage.pdf`.

### Commit 3 — `coreml-demo-rs + coreml-demo-app: macOS reference apps`

Two reference apps demonstrating downstream use of `tract-coreml`:
- `coreml-demo-rs` — Rust binary with a C-ABI bridge for embedding from Swift
- `coreml-demo-app` — minimal Xcode/SwiftUI app that links it (webcam → tract-coreml → preview)

Not part of the backend per se; happy to drop these from the PR if you'd prefer them in a separate repo.

## Upstream improvements this work surfaced

A separate `tract-upstream-feedback.md` is included in the PR (under `coreml/docs/`). Categories:

- **8 BLOCKERS** — things that prevented a model from loading or being baselined (B1 InferenceFact value override; B2 symbolic batch dims in Conv facts; B3 RmsNorm F16 CPU dtype; B4 SAM 2 value_info; B5/B6 NNEF declutter; **B8 `com.microsoft` operator domain → blocks Phi-3/4 entirely**; …)
- **16 IMPROVEMENTS** — quality-of-life / architectural items that affect every accelerator backend (I1 was Resize-F32-only, now patched locally; I4 declutter aggressively decomposes high-level ops; I5 EinSum signature canonicalisation; I7 RmsNorm declutter has no opt-out; **I13 LayerNorm → RmsNorm rewrite — possibly a correctness bug, worth investigating**; I14 Resize-nearest decomposition loses upsample intent; …)
- **6 ANNOYANCES** — papercuts (A2 Cast(Const) chains not folded; A3 AveragePool → SumPool decomposition; prost map non-determinism; …)
- **4 OPEN QUESTIONS**

**Honest caveat:** several of these I'm uncertain whether they're real issues vs valid design choices I haven't understood. I tried to mark uncertainty explicitly in the doc.

Of all of these, the ones I think generalize best beyond CoreML:
- **B3** (F16 RmsNorm CPU eval errors) — anyone testing F16 inference hits it
- **B8** (`com.microsoft` domain) — locks tract out of the modern Microsoft LLM corpus (Phi-2/3/3.5/4)
- **I1** (Resize FP16) — fixed in this PR; affects every F16 segmentation / SR / detection model
- **I13** (LayerNorm → RmsNorm rewrite) — **if this is a correctness bug it affects every transformer**; needs maintainer eyes

I'd happily extract any of these as their own focused PRs if there's interest.

## How to look at this PR

```sh
# Build (default-members unchanged — coreml is opt-in)
cargo build -p tract-coreml --release

# 57 in-CI tests
cargo test -p tract-coreml --tests

# Real-model e2e tests (need external model files; ignored by default)
cargo test -p tract-coreml --release --test mobilenet_e2e -- --ignored --nocapture
cargo test -p tract-coreml --release --test dfn3_dec_e2e dfn3_perf_cpu_vs_coreml -- --ignored --nocapture

# CPU-only path unchanged
cargo test -p tract-core --lib   # 237 passed, 0 failed, 3 ignored
```

## Models tested (the canary corpus)

| Model | Class | Outcome |
|---|---|---|
| MobileNet v2 | CNN classifier | 1 CoremlOp ✅ |
| SqueezeNet 1.x | CNN classifier | 1 CoremlOp ✅ |
| InceptionV3 | CNN classifier | 2 CoremlOps |
| MODNet | CNN segmentation | 1 CoremlOp ✅ |
| RVM (Robust Video Matting) | recurrent (hidden-state I/O) | 1 CoremlOp ✅ |
| SAM 2 Hiera-Tiny encoder | transformer (windowed attention) | 39 CoremlOps, ANE compile fails |
| ViT-base | image transformer | 1 CoremlOp ✅ |
| YOLOv8n | object detection | 7 CoremlOps |
| Real-ESRGAN x4 | super-resolution | 3 CoremlOps |
| **DFN3 decoder (NNEF)** | **audio enhancement (GRU + scan)** | **6 CoremlOps, 1.06× — my driving case** |
| Parakeet decoder / joint (NNEF) | ASR (LSTM + scan) | 12 / 2 CoremlOps |
| Nemotron decoder / joint (NNEF) | ASR (LSTM + scan) | 12 / 2 CoremlOps (same shape as Parakeet) |
| SmolLM2-135M | decoder LLM (Llama-arch) | 1 CoremlOp + 2 stranded I64 Casts; ANE compile fails (structural — `scatter_nd` + bool ops) |
| Phi-3-mini-4k | decoder LLM (Microsoft) | aborts at load — see B8 (`com.microsoft` domain) |

**Caveat I want to surface:** I'm not sure which of these reflect typical tract community usage. SAM 2 / SmolLM2 / Phi-3 were probes to find gaps, not models I expect tract users to deploy directly. If you tell me "drop everything except CNN classifiers + audio + maybe RVM," I'll happily trim the canary corpus for the production-ready PR.

## What's deliberately NOT in this PR

- **Pulse / streaming support** — `CoremlOp` is `is_stateless = true`. Streaming integration with tract-pulse needs design (would the MLPackage be re-compiled per pulse-frame? cached per-shape?). Open question.
- **Quantized models (Q8 / Q4)** — would require Core ML quantization tables; not exercised by any canary.
- **NeuralNetwork (legacy) format** — MLProgram only.
- **iOS / tvOS targets** — macOS-only objc2 dependencies in `Cargo.toml`. Adding more targets is config-only but I haven't validated.
- **An `mb.while_loop` translator** — see DFN3 discussion above. Want to talk about this before building it.

## Attached artifacts

- **`coreml/docs/ops-coverage.pdf`** — full op-family → MIL-op mapping table + per-canary stranded-op breakdowns
- **`coreml/docs/benchmarks.pdf`** — full per-canary performance matrix with histograms, methodology, ANE engagement notes, and reproduction commands
- **`coreml/docs/tract-rnn-preservation-handout.md`** — the proposal sketched in the DFN3 discussion above (one upstream change → four backends benefit)
- **`coreml/docs/tract-upstream-feedback.md`** — the 8 BLOCKERS + 16 IMPROVEMENTS + 6 ANNOYANCES + 4 OPEN QUESTIONS doc, with confidence markers

## Asks

1. **Should this land at all** — separate crate in the workspace, separate repo, or "interesting but not for tract"?
2. **Should Commit 1 (Resize F16) split out** as its own PR? I think yes.
3. **DFN3 / Path C (upstream RNN preservation)** — is there appetite for the handout? Even without my backend, it would unlock four other accelerator paths.
4. **Of the upstream blockers** — which are real vs design choices I've misread? I'd love a redirect on the ones I got wrong before I file separate PRs.

Thanks for reading something this large. Happy to iterate on any of it; no expectation that all of it lands.


kali:
Wow. Let me think about how to address that (inc discussing with colleagues and managers here, because if we do this, it will need some time from me).

kali:
1. So, I am interested, even if it's not super relevant for Sonos. Obviously this is not reviewable and/or mergeable as is.

I think we should treat it as a proof-of-concept and exploratory preliminary work: it shows huge speed-up on some use cases, and you identified some of the potential hurdles.

I am cautious with NPU and higher level NN APIs, because I have been bitten in the past: I know that for CPU and GPU, we can be confident that we can basically run anything within reason. I use cudnn for optimised cuda convolution, but I have a generic "naive" on-GPU fallback for weird  cases that cuddn does not like. There is no such options with NPUs or NN-oriented APIs.

So I think we would start with guessing what does success look like on a CoreML binding. Are there class (as in 2D-CNN, RNN, attention...) of models that we want to support, some that we know will never work because of limitations. You have already identified difficulties with RNN ops (and pulse), and limitation with datum types.

I don't want to steer you to far off your driving problem, DFN3: my intuition and experience with other NPUs is "NPU are for 2D CNN", so I'm not sure if the pay-off will be there for you if we invest a lot of time on CoreML.

So I think we have two ways to move this forward, and it's pretty much up to you. Either we go:
 * we do CoreML, take the time to do it right, we find out what it is good for, and it's fine if it useless for DFN3 and people can use it to speed up stable diffusion
 * keep experimenting a bit, enough to prove a business case for DFN3 (and other networks Sonos may be interested in), prove it can bring value, breaking what needs to be broken in tract, and then we'll think about how to proceed for integration if it make sense. This could mean: evaluate the plus-value of keeping LSTM and GRU for a variety of model and platforms. Maybe Trunet and DFN3 on CPU, parakeet and Nemotron on GPU and CoreML.

2. I agree we should cherry-pick the Resize bit separately.

3. Scan VS LSTM and GRU

This design comes from three main reasons: 1. less work / less code (at least on paper) 2. the need for "exotic" variations on the LSTM structure 3. on CPU, little performance incentive to keep them together

I am not necessarily adverse to reifying LSTM and GRU. I do not consider this a change of policy, more like necessary adaptations. My policy is being pragmatic. We started transformers with attention as "two matmuls with a softmax in between", but reified it back to unlock flash attention, making it an extended nnef operator that torch 2 nnef can target.

So make the case for it, and I will listen.

4. Hope these indication brought some clarity :)

Possible next steps:

* prove the value of running DFN (and maybe a bigger nvidia ASR encoder, maybe ?) on CoreML, with reified recurring ops. break tract operator set. the goal is to make a point, forget about obtaining mergeable code.
* I don't want to waste your time. So don't go beyond this point before I'm in.
* Once I give you a green light, we can open multiple tracks
  * a CoreML "integration goal" definition. I like to think 100% of the models we have should run fully on CPU, cuda or Metal. This is the blue line for the "flexible" targets. I know it won't be reachable with CoreML, so we need to define a reasonable objective, also taking into account what you can commit to.
  * a sequence of manageable PR for CoreML integration
  * reification of LSTM and GRU (with possible improved support on small CPU if there are opportunities that I can sell internally)

Another way to see this is, we are looking at two pretty big changes. Reification of GRU and LSTM on one side, CoreML support on the other. Both of them probably have value on their own (we have some proof elements here that CoreML for image makes sense). You need both things to be done to take advantage of DFN3 on CoreML.

WDYT ?

czoli1976:
Thanks for taking the time to think this through — your framing helped a lot.

Honest update from the work since: I'm skeptical DFN3 (or similar streaming-audio workloads) actually benefits from CoreML/ANE. Empirically: the ANE compiler rejects recurrent patterns, ORT+CoreML-EP doesn't beat ORT-CPU on df_dec at the same shapes, and libDF on CPU is already RTF 0.054 (~19× real-time) on Apple Silicon. Validates your "NPU are for 2D CNN" intuition.

So I'd like to set the CoreML PR aside as a proof-of-concept reference (no plans to expand without your go-ahead) and refocus on work that benefits either:
- **WASM** (directly or indirectly — extending the relaxed-simd kernel work)
- **Sonos's Cortex Linux product** / the wider tract community on CPU

Near-term concrete:
- Cherry-picking the Resize-F16 patch you green-lit as a standalone PR
- Small upstream PRs from the discovery list — most are CPU-side correctness/leniency fixes, reviewable in an hour each, one at a time over the coming weeks
- The RNN-preservation case — the strongest single argument is the **15.6× tract-CPU vs ORT-CPU gap on df_dec** at T=100/F32/into_optimized, which validates the fused-GRU-kernel argument independently of CoreML. I'll write it up properly and post separately when ready, not as code.

CoreML stays parked. If a use case emerges later where it actually matters — vision models for ANE, or the Stable-Diffusion-class workload you mentioned — happy to revisit with your input on scope then.

Thanks again.

czoli1976:
Closing this draft — the Core ML / ANE backend was an exploratory spike and I'm not taking it toward merge. The work stays on the `feature/coreml-bridge` branch if there's future interest. Thanks for taking a look.

PR #2226, opened 2026-05-14
core/cnn: make eager-vs-lazy Im2col choice shape-aware

czoli1976:
## What & why

`should_use_lazy` (`core/src/ops/cnn/conv/conv.rs`) picks eager vs lazy `Im2col` purely on **kernel volume** (`>= lazy_im2col_min_kernel()`, default 6). It never looks at `n`, the output spatial size — yet the eager `Im2col` scratch is `k · n · sizeof` bytes. So a *small kernel over a large output* still allocates, writes, and reads back multiple MB of scratch every inference — a pure memory-bandwidth tax that lazy's on-the-fly per-panel gather avoids.

Profiling DeepFilterNet 3 against ONNX Runtime Web surfaced this: `df_dec`'s `df_convp.1` (`in=64 out=10`, kernel `5×1`, `group=2` → `m=5 k=160 n=9600`) has kernel volume 5 — one below the threshold — so eager is chosen and materialises a ~5.9 MiB scratch. That `Im2col` node alone measured *more than the matmul it feeds*.

## The change

One file. An **additive** shape-aware term in `should_use_lazy`: also pick lazy when the eager scratch would exceed `TRACT_LAZY_IM2COL_MAX_EAGER_BYTES`.

- Default is **cfg-gated** — 1 MiB on wasm, 4 MiB native: the materialisation tax bites harder on wasm (no hardware prefetch, bounds-checked stores) while native caches/prefetchers absorb a few MB.
- The kernel-volume rule stays as the primary signal. The new term is **purely additive** — it only ever moves a conv eager → lazy, never the reverse, so **nothing that was lazy can regress**.
- The new env var mirrors the existing `TRACT_LAZY_IM2COL_MIN_KERNEL` pattern.

## Results — `df_dec` end-to-end

Per-node profiler, `S=100`, alternating-order interleave, median of paired runs:

| target | before | after | Δ |
|---|---|---|---|
| wasm (wasip1, `+simd128,+relaxed-simd`) | 11.07 ms | 9.43 ms | **−1.64 ms (−14.8%)** |
| native (Apple Silicon) | 11.02 ms | 10.34 ms | **−0.68 ms (−6.2%)** |

## Validation

- **Bit-exact output** — `df_dec`, `erb_dec`, MobileNetV2, SqueezeNet, InceptionV3, native + wasm: every output element is byte-identical before/after (lazy and eager `Im2col` are mathematically identical).
- **Test suite** — `cargo test --release` over `tract-core` / `tract-onnx` / `tract-nnef` / `tract-hir`: 348 passed, 0 failed.
- **No CNN regression** — MobileNetV2 / SqueezeNet / InceptionV3 reroute **zero** convs (their sub-6 kernels are 1×1 → einsum or depthwise; their ≥3×3 convs already took the lazy path). Optimized graphs are byte-identical to before, native and wasm.
- `cargo fmt` clean; no new `clippy` warnings.

## Open questions — this is a draft, opinion wanted

1. **Is lazy gated out for small kernels for a reason** beyond gather-indirection cost — a past regression, a memory-constrained-target consideration, a correctness edge case? The `should_use_lazy` comment reads as a deliberate eager/lazy tradeoff, so I'd rather ask than assume.

2. **Is a tunable shape-aware heuristic the right level here**, or would you rather this become a proper cost model (estimate `eager_cost` vs `lazy_cost` from shape + hardware params, pick the cheaper)? A cost model is more principled, but trades two threshold constants for a handful of hardware-model constants; `Op::cost()` might be the natural hook. **Is it worth investing further in that direction**, or is a good default + env override the shape you'd want?

3. The wasm/native cfg split is the crude part — a runtime-detected cache size could replace the hardcoded 4 MiB if you'd prefer less magic.

The second commit (profiling tooling under `examples/wasm-model-bench`) is throwaway — happy to drop or relocate it; it's only there so the numbers above reproduce.

### DIFF
--- a/core/src/ops/cnn/conv/conv.rs
+++ b/core/src/ops/cnn/conv/conv.rs
@@ -1198,13 +1198,7 @@ impl TypedOp for Conv {
             } else if input_fact
                 .shape
                 .as_concrete()
-                .map(|s| {
-                    should_use_lazy(
-                        &self.pool_spec.data_format.shape(s.into()).unwrap(),
-                        &self.pool_spec,
-                        self.group,
-                    )
-                })
+                .map(|s| should_use_lazy(&self.pool_spec, self.group, s, input_fact.datum_type))
                 .unwrap_or(false)
             {
                 let mut patch = TypedModelPatch::new("lazy-im2col");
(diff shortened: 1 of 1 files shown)


kali (approved):

kali:
I'm happy with this sensible defaults + environment variable compromise. No problem with cfg!() limits.

PR #2227, opened 2026-05-14
linalg/wasm: fix f32 GEMM routing to the 32x1 GEMV kernel (1.13-1.78x)

czoli1976:
## Problem

On `wasm32`, every f32 GEMM (`MatMul` with N > 1) was running on `wasm_f32_32x1` — a *matrix×vector* kernel (MR=32, NR=1) — instead of `wasm_f32_8x8`.

`kernel_selection::strategize` honours the `mmm_f32` dispatch callback only when the callback's kernel is tagged `ManuallyOptimized`. `wasm_f32_8x8` was tagged `TargetOptimized`, so `strategize` discarded the callback and fell through to `list_impls`, where `retain()` keeps only the top `ImplementationQuality` and drops every `TargetOptimized` kernel. That left only the four `ManuallyOptimized` GEMV kernels (`*x1`), and the N>1 branch then picked `max(nr*mr)` among them → `wasm_f32_32x1`. The in-code comment claimed "8x8 still wins" through that fallthrough — it missed the `retain()` step.

## Fix

Tag `wasm_f32_8x8` `ManuallyOptimized` so `strategize` honours the `mmm_f32` callback that already returns it. Also: correct the stale comment, and add a regression test asserting every kernel reachable through the `mmm_f32`/`mmv_f32` callbacks is `ManuallyOptimized` (the invariant `strategize` depends on). The GEMV path is unaffected — the `*x1` kernels were already `ManuallyOptimized`.

## Benchmark — standalone f32 MatMul, wasm32

| Shape M×K×N | before | after | speedup | ORT-Web | after vs ORT |
|---|--:|--:|--:|--:|---|
| 256×256×100 | 347 µs | 279 µs | 1.24× | 376 µs | **tract 1.35×** |
| 256×256×256 | 840 µs | 643 µs | 1.31× | 850 µs | **tract 1.32×** |
| 768×256×100 | 907 µs | 782 µs | 1.16× | 1109 µs | **tract 1.42×** |
| 8×576×1024 | 403 µs | 226 µs | 1.78× | 291 µs | **tract 1.29×** |
| 10×576×1024 | 453 µs | 402 µs | 1.13× | 350 µs | ORT 1.15× |
| 16×576×1024 | 601 µs | 400 µs | 1.50× | 527 µs | **tract 1.32×** |

`wasmtime 44`, `+simd128,+relaxed-simd`, single-thread. tract = whole `MatMul` node, interleaved min-of-N paired (before/after measured back-to-back each round → ratio robust to machine load). ORT-Web = `onnxruntime-web` 1.24.3, wasm EP, 1 thread, per-op profiler `e.dur` (pure kernel time, apples-to-apples with tract's per-node timing). Models are standalone `MatMul` ONNX, const weight × runtime activation.

After the fix tract's WASM GEMM beats ORT-Web on 5 of 6 shapes (1.29–1.42×). The one loss — 10×576×1024 — is the ragged-N case: N=10 with the 8×8 kernel's NR=8 needs 2 N-tiles, the second only 2/8 utilised. That's a separate, larger change (a narrower-NR kernel + NR-aware dispatch).

## Correctness

- `cargo test -p tract-linalg --target wasm32-wasip1 --lib`: **1741 passed, 0 failed, 3 ignored.** The `MMMRustKernel!`-generated suite covers `wasm_f32_8x8`'s mat_mul / packed_packed / property / fused-op tests — the kernel body is unchanged, only its quality tag.
- New `dispatch_trace::dispatch_kernels_are_manually_optimized` regression test passes.
- `cargo fmt` clean.
- Native builds unaffected — the `wasm` module is `#[cfg(all(target_family = "wasm", target_feature = "simd128"))]`.

### DIFF
--- a/linalg/src/wasm.rs
+++ b/linalg/src/wasm.rs
@@ -44,16 +44,21 @@ pub fn plug(ops: &mut Ops) {
     ops.mmm_impls.push(wasm_f32_16x1.mmm());
     ops.mmm_impls.push(wasm_f32_32x1.mmm());
     ops.mmm_impls.push(wasm_f32_8x8.mmm());
-    // Selection paths:
-    //   - N>1 (GEMM): mmm_f32 returns 8x8. 8x8 stays TargetOptimized so
-    //     kernel_selection::strategize falls through to list_impls and picks
-    //     by max(nr*mr) — 8x8 still wins (4x4 would be the only alternative).
+    // Selection paths. Both rely on kernel_selection::strategize honouring
+    // the mmm_f32 / mmv_f32 callback, which it only does when the callback's
+    // kernel is tagged ManuallyOptimized. Otherwise strategize falls through
+    // to list_impls, whose retain() keeps only the top ImplementationQuality
+    // and drops every TargetOptimized kernel.
+    //   - N>1 (GEMM): mmm_f32 returns 8x8, so 8x8 MUST be ManuallyOptimized.
+    //     If it were TargetOptimized it would be dropped by retain(), and the
+    //     N>1 branch's max(nr*mr) over the surviving (ManuallyOptimized) GEMV
+    //     kernels would pick wasm_f32_32x1 — a matrix×vector kernel — for
+    //     every GEMM.
     //   - N=1 (GEMV): mmv_f32 routes by M-band to the kernel whose MR fits.
-    //     The four GEMV kernels are tagged ManuallyOptimized so strategize
-    //     hits its early-return at kernel_selection.rs:35 and honours the
-    //     callback's choice. Without that tag, strategize would discard the
-    //     callback and pick max(mr)=32x1 for every M, leaving up to ~37% on
-    //     the table for small-M GEMV.
+    //     The four GEMV kernels are ManuallyOptimized for the same reason —
+    //     without the tag strategize discards the callback and picks
+    //     max(mr)=32x1 for every M, leaving up to ~37% on the table for
+    //     small-M GEMV.
     ops.mmm_f32 = Box::new(|_m, _k, _n| wasm_f32_8x8.mmm());
     // Bands derived from microbench_dispatch_gemv. At each band edge, using
     // the next-larger kernel beats halving outer iterations of the smaller
@@ -2094,7 +2099,10 @@ unsafe fn kernel_f32_8x8(mut pnl: *const FusedKerSpec<f32>) -> isize {
     }
 }
 
-MMMRustKernel!(kernel_f32_8x8 => wasm_f32_8x8<f32>(8,8)@(8,8) quality(ImplementationQuality::TargetOptimized));
+// ManuallyOptimized so kernel_selection::strategize honours the mmm_f32
+// callback that returns it for N>1 GEMM (see the `plug` comment) — otherwise
+// strategize drops it and routes every GEMM onto the 32x1 GEMV kernel.
+MMMRustKernel!(kernel_f32_8x8 => wasm_f32_8x8<f32>(8,8)@(8,8) quality(ImplementationQuality::ManuallyOptimized));
 
 #[cfg(test)]
 mod dispatch_trace {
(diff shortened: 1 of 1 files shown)


czoli1976:
Quick follow-up on the remaining ORT gap at `10×576×1024` (the one shape in the benchmark above where tract still trails, ~1.15×). Explored two angles for closing it, both negative results — documenting in case either sparks an idea.

### Attempt 1 — narrower-NR kernel for N=10's ragged tile

Wrote `wasm_f32_8x4` (MR=8, NR=4, 8 fused f32x4 accumulators — same ILP budget as `wasm_f32_32x1`, well inside the 16-register WASM budget). Idea: N=10 with NR=4 tiles as 4+4+2 (1.20× waste) vs NR=8's 2 tiles (1.60×). Then NR-aware `mmm_f32` dispatch.

Microbench (Apple Silicon, wasmtime 44, `+simd128,+relaxed-simd`, min-of-6, K=576):

```
peak  n=64   : 8x4=32.6 GF/s  8x8=54.3 GF/s  -> 8x8
peak  n=1024 : 8x4=32.4 GF/s  8x8=54.5 GF/s  -> 8x8
exact n=8    : 8x4=32.1 GF/s  8x8=55.0 GF/s  -> 8x8
ragged n=10  : 8x4=27.2 GF/s  8x8=33.9 GF/s  -> 8x8
ragged n=12  : 8x4=33.0 GF/s  8x8=40.7 GF/s  -> 8x8
ragged n=16  : 8x4=33.3 GF/s  8x8=54.6 GF/s  -> 8x8
ragged n=20  : 8x4=33.4 GF/s  8x8=45.3 GF/s  -> 8x8
```

`8x4` flat-tops at ~32 GF/s regardless of N — load-bound. Reason: NR=4 means each row-accumulator is one v128, so each A-splat feeds exactly 1 FMA. 8x8's NR=8 (lo+hi) gets 2 FMAs per splat — twice the arithmetic intensity per A-load. Both kernels do 8 A-loads + 8 splats per K-step but 8x4 amortizes over half the FLOPs. Lower partial-tile waste at N=10 can't close the peak gap: `32 × 10/12 = 27 < 54 × 10/16 = 34`. Looks structural to WASM's 4-wide SIMD — no fast narrow-NR kernel here.

### Attempt 2 — pack fusion (lazy `OptMatMulPack`, MLAS-style)

Per-node profile of `bench-onnx` shows the 1.15× = `OptMatMulPack` running as a separate node:

```
node 3  OptMatMul       363 µs/call
node 1  OptMatMulPack    63 µs/call   ← full 2.3 MiB activation pack
                        --------
total                   ~426 µs/iter   (ORT e.dur for same shape = 350 µs)
```

Implemented `LazyPackedInput` in `core/src/ops/matmul/lazy_pack.rs` parallel to `LazyIm2colInput` — holds the runtime `TValue` directly (cheap Rc-clone, no deep copy), `panel_bytes(i, buffer)` packs the i-th panel into the framework's per-thread scratch via `PackedFormat::pack_t` for `mn_range = i*r..(i+1)*r`. Added `lazy: bool` to `OptMatMulPack`, opted in for runtime A and B at the matmul lowering (`einsum_matmul.rs:841` + the parallel pack_b site at 862). Had to fix `FusedSpec::prefer_col_outer` along the way — was using `format().is::<PackedFormat>()` to detect "eager", but `LazyPackedInput` legitimately has `PackedFormat` as its format too; switched to `scratch_panel_buffer_layout().is_some()` as the lazy marker (returns `None` from `EagerPackedInput`, `Some` from both lazy types).

Mechanically clean — `OptMatMulPack` drops from 63 µs to 0.57 µs. But the **whole-matmul** time is slightly *slower*, not faster (paired min-of-N, interleaved fix-vs-lazy on the same machine):

```
Shape M×K×N      eager   lazy   speedup
256×256×100      266     270    0.98×
256×256×256      647     662    0.97×
768×256×100      786     791    0.99×
8×576×1024       228     266    0.85×
10×576×1024      407     443    0.91×
16×576×1024      401     440    0.91×
```

The pack work moved into the kernel call, and per-panel `panel_bytes()` through `Box<dyn MMMInputValue>` + `dispatch_copy!` + `pack_t`'s branching adds ~30-35 µs of per-call setup at this shape (128 N-panels × ~280 ns/call). The cache-locality benefit doesn't recover that — the eager 2.3 MiB pack fits in Apple Silicon's L2, so the kernel reads it from L2, not RAM; L1-vs-L2 savings are only ~5-10 µs. Break-even seems to be "pack > L2" — i.e. transformer-FFN territory (8+ MiB packed activations), not DFN3-like CNN/RNN shapes.

### Where this leaves it

The 50-60 µs gap looks structural to the dispatch-through-trait-object indirection: MLAS appears to inline/monomorphize the pack inside the matmul kernel, paying zero per-panel dispatch tax. Anything built on top of the current `MMMInputValue` indirection seems to hit the same wall.

Two questions:

1. Is there a tract pattern for monomorphizing pack into a kernel that we're missing? `lazy_im2col` works for conv because the gather indirection is effectively free vs conv→GEMM's amortization, but for plain matmul the same machinery is net negative.

2. Worth pursuing a bulk-pack variant — pack N panels on entry to the col-outer loop and amortize the per-panel dispatch over a larger work unit — or has that been considered and rejected?

Either way the dispatch fix in this PR is the clean win standalone — both follow-ups would be separate. Happy to drop them, or pursue if there's a direction you can see.

kali:
i don't think you missed anything. But I wonder if the pancake-shape matmuls are really frequent in real life as core witll try to transpose them into skinny products.
In any case, this feels like a net positive. Happy to merge.

kali (approved):

PR #2229, opened 2026-05-15
core/einsum: fold contiguous same-role axes in standard codegen

czoli1976:
## What & why

`merge_consecutive_same_role_axes` (added in `05184d051` for SD-1.5 attention) implements axis folding for matmul-like einsums — when two adjacent axes carry the same `(in_input_0, in_input_1, in_output)` role, it reshapes the inputs to merge them into one, rewrites the einsum formula, and inserts an unmerge `Reshape` on the output. But it's only wired into `rewrite_einsum_to_prefix_matmul` (the NNEF / Metal / CUDA serialization paths). `EinSum::codegen` on the standard `into_optimized()` path that ONNX → native CPU / WASM use goes straight to `einsum_matmul::detect_rule`, which picks one axis per role and leaves the rest as batch axes.

For a 1×1 NCHW pointwise conv at `[B, Cin, H, W]` with `W > 1` this is exactly the wrong choice: `H` and `W` both have role `(true, false, true)`, are consecutive in input[0] (`NIHW`) and absent from input[1] (`OI`), so the merge should collapse `{H, W}` into a single axis of size `H·W`. With the merge dormant, `detect_rule` picks `m = H` and leaves `W` as a batch axis. Result: `W` independent matmul calls of `[H, Cout] @ [Cin, Cout]` instead of one `[H·W, Cout] @ [Cin, Cout]`. The per-call setup tax (pack-b, FusedSpec resolve, C-view, fused-post-op dispatch) dominates the matmul itself on these shapes.

## The change

One file, one new pass. `core/src/optim/mod.rs` (+22 / −0):

- A new `MergeConsecutiveSameRoleAxes` struct implementing `TypedPass` via the `run_direct` hook (no per-patch iteration needed — the existing `Rewriter::default().with_rule_for(...).rewrite(...)` machinery already handles fixpoint).
- One entry in `Optimizer::codegen()`'s pass list between `PropConst` and the per-op codegen `OpOptim`, so the merge fires once before `detect_rule` is invoked on each EinSum.

That's the entire fix. The merge pass itself is unchanged.

## Results — 6 batch canaries (native, Apple Silicon) + TRUNET streaming

`profile-nodes` (per-node timed `eval` callback wrapping `tract_core::plan::eval`, 200 reps for fast models / 100 for slow):

| model              | baseline   | after      | Δ          |
|--------------------|-----------:|-----------:|-----------:|
| erb_dec (DFN3)     |   8.49 ms  |   6.91 ms  | **−1.58 ms (−18.6%)** |
| df_dec  (DFN3)     |  12.47 ms  |   8.89 ms  | **−3.58 ms (−28.7%)** |
| mobilenetv2-7      |  71.96 ms  |  35.81 ms  | **−36.16 ms (−50.2%)** |
| squeezenet         |  25.91 ms  |   9.46 ms  | **−16.46 ms (−63.5%)** |
| inception_v3       | 119.59 ms  |  67.59 ms  | **−51.99 ms (−43.5%)** |
| yolov8n            | 163.11 ms  | 120.17 ms  | **−42.94 ms (−26.3%)** |
| trunet (pulse=1) † |   0.53 ms  |   0.50 ms  | **parity (within noise)** |

† TRUNET is the streaming check (`tract -O --nnef-tract-core … --pulse 1 bench`, 1 frame/inference, interleaved main-vs-PR to cancel drift, Apple Silicon) — a different harness from the batch `profile-nodes` rows above. It is the model that surfaced the second fix: an earlier build of this PR *regressed* TRUNET pulse1 by +9.3% (x86, reported by @kali) / +6.3% (ARM). At pulse=1 every 1×1 conv has N=1 and W=1, so the merge folded a single non-unit axis — which was already the matmul `m` — buying nothing but a `MoveAxis`. Commit d084af611 gates both merge branches on "≥2 non-unit axes" (the exact condition for a fold to eliminate ≥1 matmul batch iteration; symbolic extents count as non-unit), restoring parity while keeping every batch win above.

Per-op breakdown on erb_dec confirms the win is concentrated in `OptMatMul` (−1.94 ms by op type), partly offset by `+0.61 ms MoveAxis` (the merge inserts a k-axis permute on input and unpermute on output) and `+0.08 ms` for now-unfused `OptAddByScalar` + `OptAddUnicast`. Net: the matmul gain dominates by ~5×. (Update: the same d084af611 guard also declines erb_dec's wasteful N=1 fold, so its `MoveAxis` now drops to ~0.05 ms — the +0.61 ms above was the pre-guard figure.)

## Validation

- **Bit-exact output.** With zero inputs, the L2 / mean / first5 / last5 / 32-bit XOR-of-bit-patterns of `output[0]` is byte-identical before/after on **erb_dec** (xor `0x0171bbd3`) and **mobilenetv2-7** (xor `0x070b18d9`). Not "within 1e-5 tolerance" — bit-exact. (Which is what one would expect: the merge is purely a reshape + axis-rename, so the matmul kernel sees the same scalars in the same accumulation order.)
- **No regressions.** Every one of the 6 canaries above improved. The biggest single win is on a vision model (squeezenet, −63%) — the optimization is for every CNN with 1×1 pointwise convs over W>1, not a one-model fix.
- `cargo test --release -p tract-core` — 238 passed, 0 failed, 3 ignored (+ 1 doctest passing). `cargo fmt --check` clean; no new `clippy` warnings (`cargo clippy --release --no-deps -p tract-core` — 15 warnings, all pre-existing in files this PR doesn't touch).

## Open questions — this is a draft, opinion wanted

1. **Why isn't `merge_consecutive_same_role_axes` called on the standard `into_optimized()` codegen path today?** Your commit `05184d051` wired it into `rewrite_einsum_to_prefix_matmul` (NNEF / Metal / CUDA) — was leaving it out of `EinSum::codegen` deliberate (e.g., concern about a downstream fuse — see below), or just a gap because SD-1.5 attention happened to go through the NNEF path?

2. **The `OptBinUnicast` skip-Add fuse is lost on these shapes.** The merge inserts an unmerge `Reshape` between `OptMatMul` and its successor. `OptMatMul::fuse` only matches direct successors, so the AddUnicast pre-fusion (and the bias + ReLU pre-fusions) become separate ops. On erb_dec that costs ~0.13 ms in unfused post-ops, dwarfed by the ~1.94 ms matmul gain — so this PR ships the merge-only fix. But if you'd prefer a follow-up that recovers the fuse, two options come to mind: (a) extend `OptMatMul::fuse` to peek through a contiguous-layout `Reshape`, or (b) inline the fold inside `detect_rule` so `OptMatMul` output rank is preserved. Do you have a preferred direction?

3. **The new `MoveAxis` cost (~0.6 ms on erb_dec) is the more interesting overhead.** The merge inserts a k-axis permute before packing and an inverse permute after the output. Is there a path to fold without the k-axis MoveAxis — say, a packer-aware k-axis position, or a `detect_rule` flavor that does the fold while keeping the original k-axis position?

### DIFF
--- a/core/src/ops/einsum/einsum_matmul.rs
+++ b/core/src/ops/einsum/einsum_matmul.rs
@@ -100,6 +100,25 @@ fn merge_same_role_axes_rule(
         }
     }
 
+    // Reject the group if any axis has mismatched per-input dims. This catches
+    // broadcasting cases — e.g. GQA `bhgmk,bhgnk->bhgmn` where g has dim 2 in
+    // input[0] and dim 1 in input[1]. Merging would collapse the broadcast
+    // structure into a non-broadcast dim mismatch that downstream OptMatMul
+    // codegen / kernels cannot handle.
+    if let Some(ref group) = best_group {
+        let input_facts = model.node_input_facts(node.id)?;
+        let input_shapes = op.actual_input_shapes_from_facts(&input_facts)?;
+        let dims_match = group.iter().all(|c| {
+            match (a_order.iter().position(|x| x == c), b_order.iter().position(|x| x == c)) {
+                (Some(p0), Some(p1)) => input_shapes[0][p0] == input_shapes[1][p1],
+                _ => true,
+            }
+        });
+        if !dims_match {
+            best_group = None;
+        }
+    }
+
     if let Some(group) = best_group {
         // Found a mergeable group. Emit the patch.
         let input_facts = model.node_input_facts(node.id)?;
--- a/core/src/ops/einsum/einsum_matmul.rs
+++ b/core/src/ops/einsum/einsum_matmul.rs
@@ -49,6 +49,25 @@ fn merge_same_role_axes_rule(
     let b_order: Vec<char> = op.axes.axes(InOut::In(1)).map(|a| a.repr).collect();
     let c_order: Vec<char> = op.axes.axes(InOut::Out(0)).map(|a| a.repr).collect();
 
+    // Per-input shapes, used both to reject broadcast merges and to gauge
+    // whether a merge earns its reshape / axis-permute cost.
+    let input_facts = model.node_input_facts(node.id)?;
+    let input_shapes = op.actual_input_shapes_from_facts(&input_facts)?;
+    // An axis is "non-unit" if its extent is not statically 1 (symbolic extents
+    // count as potentially large). A fold only reduces the number of matmul
+    // invocations when it combines at least two non-unit axes; folding a unit
+    // axis — the streaming axis at pulse=1, or a batch axis of 1 — leaves the
+    // matmul geometry untouched and only adds reshapes (and, in the k-axis
+    // branch, a MoveAxis), so we decline those.
+    let is_non_unit = |c: &char| -> bool {
+        let dim = a_order
+            .iter()
+            .position(|x| x == c)
+            .map(|p| &input_shapes[0][p])
+            .or_else(|| b_order.iter().position(|x| x == c).map(|p| &input_shapes[1][p]));
+        dim.map_or(true, |d| d.as_i64() != Some(1))
+    };
+
     // Find first group of 2+ same-role axes that are consecutive in all inputs.
     // Scan each input's axis order for runs of same-role axes.
     let role_map: std::collections::HashMap<char, Role> = axes.iter().cloned().collect();
@@ -100,29 +119,28 @@ fn merge_same_role_axes_rule(
         }
     }
 
-    // Reject the group if any axis has mismatched per-input dims. This catches
-    // broadcasting cases — e.g. GQA `bhgmk,bhgnk->bhgmn` where g has dim 2 in
-    // input[0] and dim 1 in input[1]. Merging would collapse the broadcast
-    // structure into a non-broadcast dim mismatch that downstream OptMatMul
-    // codegen / kernels cannot handle.
     if let Some(ref group) = best_group {
-        let input_facts = model.node_input_facts(node.id)?;
-        let input_shapes = op.actual_input_shapes_from_facts(&input_facts)?;
+        // Reject the group if any axis has mismatched per-input dims. This
+        // catches broadcasting cases — e.g. GQA `bhgmk,bhgnk->bhgmn` where g has
+        // dim 2 in input[0] and dim 1 in input[1]. Merging would collapse the
+        // broadcast structure into a non-broadcast dim mismatch that downstream
+        // OptMatMul codegen / kernels cannot handle.
         let dims_match = group.iter().all(|c| {
             match (a_order.iter().position(|x| x == c), b_order.iter().position(|x| x == c)) {
                 (Some(p0), Some(p1)) => input_shapes[0][p0] == input_shapes[1][p1],
                 _ => true,
             }
         });
-        if !dims_match {
+        // Decline merges that combine fewer than two non-unit axes: they don't
+        // shrink the matmul loop count, they only add reshapes / a MoveAxis.
+        let worth_merging = group.iter().filter(|c| is_non_unit(c)).count() >= 2;
+        if !dims_match || !worth_merging {
             best_group = None;
         }
     }
 
     if let Some(group) = best_group {
         // Found a mergeable group. Emit the patch.
-        let input_facts = model.node_input_facts(node.id)?;
-        let input_shapes = op.actual_input_shapes_from_facts(&input_facts)?;
         let output_shape = super::eval::output_shape(&op.axes, &input_shapes)?;
 
         let drop_set: Vec<char> = group[1..].to_vec();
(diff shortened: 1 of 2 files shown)


kali:
I think we should adress 2 somehow, because we may encounter some long fusable chains in some models. I know for experience the fusing code is brittle but I think we have better test coverage than we used to, so it's not as scary as it used to be.

3. I'm not sure why we have an actual permutation ? the function name we call means it fused only consecutive axes, is it a lie ?

czoli1976:
Thanks Kali.

**Q3.** Not a lie, but the function name only describes one of its two branches (both yours, from `05184d051`). Branch 1 = pure merge of already-consecutive same-role axes. Branch 2 = if a same-role pair in one input is separated by a single k-axis, insert `AxisOp::Move(mid, end)` to push the k-axis out of the way; the Rewriter then re-fires Branch 1.

The +0.6 ms `MoveAxis` on erb_dec is Branch 2 firing on 1×1 NCHW pointwise convs: input[0] is `NIHW`, so the same-role `(N, H, W)` triplet is split by the contracted `I` → move `I` to the end (`NHWI`) → Branch 1 merges `NHW`. So the permute is the price of bringing the merge to bear on this shape; on shapes where the same-role axes were already adjacent, only Branch 1 fires and there's no permute.

Two ways to remove it as a follow-up: (i) packer-aware k-axis position so the move isn't needed, or (ii) cheaper-but-narrower, gate Branch 2 to skip when `N=1` (universal in vision inference; merging `N` is a no-op for the matmul anyway). Keeping that out of this PR's scope unless you'd rather see it now.

**Red CI (GQA SDPA)** — our regression, root-caused and fixed in [ae0d4eac2](https://github.com/sonos/tract/pull/2229/commits/ae0d4eac2). SDPA decomposes to `bhgmk,bhgnk->bhgmn`; in GQA the `g` axis is broadcast (Q.g=2, K.g=1). All three of `b, h, g` carry role `(T, T, T)`, so Branch 1 picks them all and the reshape collapses the broadcast: `in0.b=16, in1.b=8` — a non-broadcast mismatch. CPU OOBed at runtime; CUDA rejected the patch at type-check; wasm trapped.

19-line guard in `merge_same_role_axes_rule`: drop the group if any axis has mismatched per-input dims. Conv merges are unaffected because their group axes carry role `(T, F, T)` — present only in input[0] — so the check is vacuous. 63/63 SDPA tests now green locally; canaries hold:

| model         | baseline | now      | Δ          |
|---------------|---------:|---------:|-----------:|
| erb_dec       |   8.49 ms |  7.09 ms | **−16.5%** |
| df_dec        |  12.47 ms |  9.27 ms | **−25.7%** |
| squeezenet    |  25.91 ms |  9.81 ms | **−62.1%** |
| mobilenetv2-7 |  71.96 ms | 35.05 ms | **−51.3%** |

**Q2.** Agreed. Preference: path (a) — extend `OptMatMul::fuse` to peek through a contiguous-layout `Reshape`. Generic (helps any later pass that emits a reshape between matmul and post-ops), keeps the merge as a pure reshape, doesn't touch the hot `detect_rule`. Happy to put it in a follow-up commit on this PR — want me to proceed?

kali:
I like the idea of a follow-up here for Q2, unless it gets unreasonably big or messy. Let's leave the rest (permutations on k) for another day ?

kali:
We're good here, I think ?

czoli1976:
I think so .... please give it a try

kali:
We have a serious regression on trunet.

```
  Trunet pulse1 benchmarks (5 runs each, on this Linux x86_64 box):

  ┌────────────┬───────────┬───────────┬───────┐
  │  Variant   │   main    │  PR 2229  │   Δ   │
  ├────────────┼───────────┼───────────┼───────┤
  │ pulse1_f32 │ 210.94 µs │ 230.55 µs │ +9.3% │
  ├────────────┼───────────┼───────────┼───────┤
  │ pulse1_f16 │ 748.30 µs │ 765.57 µs │ +2.3% │
  └────────────┴───────────┴───────────┴───────┘
```

czoli1976:
reproduced ....

czoli1976:
Good catch — reproduced (+6.3% on my M-series; same shape as your x86 +9.3%) and fixed in d084af611.

**Cause.** The merge folds every same-role axis run, but a fold only reduces matmul invocations when it combines **≥2 non-unit axes**. With a single non-unit axis, that axis was already the matmul `m`, so merging changes nothing and just adds reshapes + the k-axis `MoveAxis`. At `--pulse 1` every 1×1 conv has N=1 and W=1, so all of them hit exactly this case — 12 dead `MoveAxis` nodes plus worse-packed matmuls.

**Fix.** Gate both branches on "≥2 grouped axes non-unit" (symbolic extents count as non-unit, so symbolic-S models still merge). It's not a tuned threshold — it's precisely the condition for the merge to eliminate ≥1 batch iteration.

- TRUNET pulse1 f32/f16 back to parity (interleaved: 0.505 vs 0.534 ms here); conv merge/move nodes 30 → 1.
- Batch canaries keep their wins: erb_dec −16.8%, df_dec −28.9%, squeezenet −61.6%, mobilenetv2 −52.5% vs baseline.
- erb_dec actually improves vs the previous PR state — the same guard now also declines a wasteful N=1 fold it was doing before, so its `MoveAxis` cost drops 0.65 → 0.05 ms.

tract-core (238) + test-unit-core (777) + the SDPA suite (63) all green locally; fmt clean.

kali:
ok, we recovered. merging.

Issue #2224, opened 2026-05-13
RFC: Would an opt-in apple-accelerate feature for cblas_sgemm routing be acceptable?


czoli1976:
Context: following up on #2220 (AMX shape-aware dispatch).

Before working on #2220 I explored routing low-M / low-N matmuls to `cblas_sgemm` via `Accelerate.framework` and measured ~5% gain on DFN3. After A7v2 most of that win is captured by the simpler dispatch fix (low-M shapes → NEON 8×8 instead of AMX 32×32). The two gains overlap substantially.

**The ask:** before I re-measure the Accelerate routing against post-#2220 main and prepare a clean PR, would you accept an `Accelerate.framework` dependency at all?

The implementation I have locally (Phases 0-2 done in a private worktree) is:

- ~500 lines: FFI declarations + `MatMatMul` wrapper for `cblas_sgemm` + (M, K, N) predicate + tests
- macOS / iOS only (cfg-gated)
- `Accelerate.framework` is a system library on those platforms — no new shipped binary
- Would be opt-in via `--features apple-accelerate` (default off)

The post-A7v2 residual is unmeasured. My best guess: <2% on DFN3, maybe 2-5% on heavier workloads (Inception / YOLO / SAM2 / LLM prefill). The interesting regimes are medium shapes (M, N ∈ [16, 32]) where neither AMX nor NEON 8×8 is optimal, and predicate-boundary misses (the A7v2 M ≥ 32 ∧ N ≥ 32 rule has 88.3 % accuracy on the 512-shape sweep).

**Possible answers** — any of these is helpful, I just want to know the constraint upfront:

1. **Yes, as opt-in `--features apple-accelerate`.** I'll re-measure and file a PR if the residual is meaningful.
2. **Yes, but with conditions** (minimum gain threshold to justify the path, default-on for macOS targets, no separate FFI wrapper code, BLAS-correctness CI requirements, …).
3. **No** — Accelerate as a tract dependency is undesirable on principle (maintenance burden, BLAS variants' historical correctness quirks, link complexity, …). I'd archive the work cleanly; #2220 already captured the prize for DFN3 and that's a fine outcome.

Happy to push the WIP branch as a draft PR for visibility if it's easier to evaluate than this prose; didn't want to clutter the PR queue with something that might be a non-starter for reasons orthogonal to the data.

PR #2230, opened 2026-05-15
linalg/arm64/sme: SME backend for ARMv9.2-A+SME chips (M4+, Cortex-X4+)

czoli1976:
## Summary

SME-accelerated f32 backend for tract. Two kernels:
- **`sme_mmm_f32_32x32`** — GEMM, gated on `FEAT_SME`. Overrides `ops.mmm_f32` on M4-class silicon.
- **`sme_mmv_f32_64x1`** — GEMV, gated on `FEAT_SME ∧ FEAT_SME2` (plain SME single-vec FMLA caps at ~31 GFLOPS on M4 — slower than NEON — so SME2 multi-vec is required). Overrides `ops.mmv_f32`.

AMX path stays as the M1/M2/M3 fallback and the SME-absent fallback. Open ISA — also targets Cortex-X4+, Neoverse V3+, Snapdragon X Elite.

## Numbers (M4 Air, same binary, `TRACT_SME_DISABLE=1` toggles, min-of-10)

| Model | M4+AMX | **SME (Phase 1)** | **SME+SME2 (Phase 2A)** | **Phase 2C (int8)** | P2A / AMX | P2A / P1 |
|---|---:|---:|---:|---:|---:|---:|
| df_dec | 7.89 ms | 5.83 | **4.35** | n/a¹ | **1.82×** | **1.34×** |
| erb_dec | 6.51 | 5.17 | **3.71** | n/a¹ | **1.75×** | **1.39×** |
| enc | 5.84 | 5.00 | **4.30** | n/a¹ | 1.36× | **1.16×** |
| squeezenet | 12.30 | 6.98 | **6.59** | n/a¹ | 1.87× | 1.06× |
| mobilenetv2 | 39.09 | 27.83 | 27.41 | n/a¹ | 1.43× | 1.02× |
| inception_v3 | 88.28 | 53.67 | 52.42 | n/a¹ | 1.68× | 1.02× |
| yolov8n | 129.66 | 93.34 | 92.15 | n/a¹ | 1.41× | 1.01× |
| MiniLM-L6 (b=1,sl=128) | 12.76 | 9.43 | 9.44 | n/a¹ | 1.35× | 1.00× |
| matmul 768×256×100 | 0.186 | 0.045 | 0.046 | n/a¹ | 4.04× | 0.98× |
| *SMOPA microbench (4-tile)* | — | — | — | **4017 GOPS ≈ 4 TOPS i8** | — | — |

12-model survey, 8 paired runs × 10 reps each. Geomean SME-vs-AMX ≈ 1.62× (Phase 1 alone was 1.53×). Phase 2A adds **1.16–1.39× on top of Phase 1 for the DFN3 trio**; MMM-dominated workloads unchanged within ±2% noise. Full data in `linalg/SME_PHASE1_BENCH.xlsx` (sheet "Phase 2A delta"). inception_v3 Phase 2A run had 17% bimodal stdev across the 8 samples (thermal); min-of-10 stayed clean at 52.42 ms — no regression.

¹ All surveyed models are f32; Phase 2C targets int8 quantized models only. The SMOPA throughput on M4 was validated by microbench (~4 TOPS i8, 4× the f32 FMOPA path), but no kernel-level E2E numbers exist yet — the int8 kernels are blocked on a packing-layout mismatch (see Follow-up).

## Design

**mmm_f32_32x32**:
- MR=NR=32, 2×2 grid over `ZA0.S..ZA3.S`. Depth-4 rotation saturates ~1 fmopa/cycle on M4 (Phase 0 microbench: 2012 GFLOPS, ~98% arch peak).
- Plain SME instructions only.

**mmv_f32_64x1** (new):
- MR=64, NR=1. Single SME2 vgx4 multi-vec `FMLA-into-ZA` per K-step writes to 4 consecutive ZA tile slots (the 64-element output column).
- Inner K-step: `ld1w {z0.s-z3.s}, pn8/z, [x_a]; ld1rw {z4.s}, p0/z, [x_b]; fmla za.s[w8, 0, vgx4], {z0.s-z3.s}, z4.s[0]`.
- Microbench: **~125 GFLOPS** (ZA-write-port-bound). Plain SME single-vec FMLA was measured at **31 GFLOPS** — matches Hello-SME paper, too slow, NOT used.
- Fuse ops use ZA-slice ↔ Z-group MOV multi-vec with predicated SVE arithmetic on the Z group.

Both kernels: one `smstart`/`smstop` pair per call (~11 ns); `v8..v15` stack-saved across the streaming region.

## Correctness

- **2237 arm64 + 113 SME tests pass** on M4 (Phase 1 had 2183 + 59; Phase 2A adds +54 SME2-specific via the auto-generated `MMMExternKernel!` proptest suite for MR=64 NR=1).
- End-to-end **bit-exact** SME vs AMX outputs on `matmul_256×256×256` and `mobilenetv2-7` with deterministic input.
- `cargo fmt --check` clean, zero new `cargo clippy` warnings.

## Files

| File | Δ |
|---|---:|
| `linalg/arm64/sme/sme_mmm_f32_32x32.S.j2` | +474 (new) |
| `linalg/arm64/sme/sme_mmv_f32_64x1.S.j2` | +290 (new) |
| `linalg/arm64/sme/dispatcher.j2` | +37 (verbatim copy of `apple_amx/dispatcher.j2`) |
| `linalg/src/arm64/sme.rs` | +245 (new) |
| `linalg/build.rs` | +10 |
| `linalg/src/arm64.rs` | +6 |
| `linalg/SME_PHASE1_BENCH.xlsx` | bench artifact |

`has_sme()` / `has_sme2()` read `hw.optional.arm.FEAT_SME*` as u64 sysctls on macOS (note: the existing `apple_get_syscall` helper treats sysctl output as a C string, which silently fails on integer sysctls). On Linux: `getauxval(AT_HWCAP2) & HWCAP2_SME*`. `TRACT_SME_DISABLE=1` disables both kernels for A/B testing.

## Follow-up

Phase 2C (int8 `qmmm`/`qmmv` via SMOPA) was attempted on a parallel branch. SMOPA microbench confirmed ~4 TOPS i8 on M4 (4× the f32 path), but the kernels hit a packing-layout mismatch: tract's default `PackedFormat` lays out panels as K-major / M-inner, while SMOPA reads M-major / K=4-inner. The two are transposes of each other, so SMOPA computes garbage on tract-packed data (auto-tests caught it). Parked on `feature/sme-quant-phase2c-wip`; will file as a follow-up PR after agreement on the unblock path (custom `PackedI8K4` format vs. in-kernel transpose vs. drop to SVE int MLA). Phase 2B (f16 via FMOPA-widening) similarly deferred.

## Open questions

1. Unconditional override when `has_sme()` / `has_sme2()`, or runtime opt-in? `TRACT_SME_DISABLE=1` is honored regardless.
2. `include_sme()` gates macOS OR Linux aarch64. Drop OS cfg and rely on runtime detection?

## Test plan

- [ ] `cargo test -p tract-linalg arm64::` on M4 (expect 2237 pass)
- [ ] Bench any f32 model with both env settings (expect SME ≥ AMX)
- [ ] CI on non-aarch64 and on aarch64-without-SME (gated, should no-op)


czoli1976:
@kali — quick question on direction for follow-up work after this PR merges.

The two natural next steps are (a) f32 GEMM shape variants (16×64 / 64×16 alongside the 32×32 we ship here — the "Demystifying ARM SME for GEMM" paper arxiv 2512.21473 measures ~1.23× over the square shape on bandwidth-bound dims) and (b) Phase 2C: int8 SMOPA kernels for `qmmm_i32` / `qmmv_i32`, where the WIP in `feature/sme-quant-phase2c-wip` hit the packing-layout blocker noted in the Follow-up section.

Would you be OK with **porting kernels from ARM Compute Library** for both? ACL is MIT-licensed, ARM-maintained, and has validated implementations of every shape we'd want.

- **Repo**: https://github.com/ARM-software/ComputeLibrary
- **License**: MIT
- **Location**: `src/core/NEON/kernels/arm_gemm/kernels/` — all the matmul kernels live here, including the ones we care about:
  - `sme2_interleaved_nomerge_s8s32_mopa_{1VLx4VL, 2VLx2VL, 4VLx1VL}` — int8 → int32 SMOPA GEMM, all three tile shapes
  - `sme2_interleaved_nomerge_fp32_mopa_*` — f32 FMOPA GEMM, same three shapes
  - `sme2_gemv_*` — GEMV variants (int8 via SDOT, f32 via FMLA-into-ZA)

Plan would be to transliterate into `linalg/arm64/sme/` as tract-native assembly — no external native dep, same build pattern as this PR. KleidiAI (Apache-2.0, ARM's other open-source kernel lib) would be the alternative as a build-dep wrapper, but that opens the same "external native dep" question as the in-flight #2224 Accelerate discussion; porting under MIT keeps tract self-contained.

Happy to do it as a follow-up PR after this one merges. Just want to confirm the approach before sinking more time.

kali:
Nice ! Feels better than the AMX reverse engineering.

czoli1976:
Thank you ! Definitely and quite faster too maybe some more room with SME
2.1 ? but that would require an M5 (M4= SME 2.0) here got only M1 and M4


Best Regards
Ckristian Zoli

kali:
OK for transliteration. Let's keep a mention of the source just in case. I'm not sure we need the pancakes one (M<<N), tract-core tries to transpose them. Did you notice these products appearing ?

czoli1976:
Will run the bench, you can shelve it, won’t hurt my feelings :-)


Best Regards
Ckristian Zoli

Email: ***@***.***

kali:
CI is crashing, looks like some feature detection is wrong, we're trying sme kernels on small aarch64 ?

czoli1976:
oh boy, QEMU, right?

The Kernel is checking for ARM SME/SME2 with has_sme() / has_sme2() but not checking for the Vector Lenght , must verify SVL == 512 bits

I think QEMU advertised SME but has SVL == 64 or something like that by default, so it will generate lots of garbage

czoli1976:
Good catch, and your instinct was right — it's feature detection.

The `aarch64-unknown-linux-gnu/musl` jobs run under `qemu-aarch64` user-mode (via `.travis/cross.sh`). QEMU 8.2 advertises `HWCAP2_SME` but emulates SME at a **256-bit** streaming vector length, while these kernels hardcode 512-bit geometry (16 f32 lanes per `ZA.S` slice). At 256-bit the store loop writes half the lanes it should, leaving uninitialised output — which is why the failures are in `packed_packed_empty` (Clear + Store, no FMOPA: the tell that it's geometry, not arithmetic). `cortexa53` passed because QEMU-as-A53 doesn't advertise SME at all; Apple CI passed because M1 has no SME.

Fix pushed (a506b9f): `has_sme()` / `has_sme2()` now also check `RDSVL == 512 bits` before claiming support, so any host advertising SME at a non-512 SVL falls back to the portable path. RDSVL only runs after the FEAT_SME bit is confirmed, so no SIGILL on non-SME hardware.

Reproduced locally under the same QEMU 8.2 (SVL reported as 256) and confirmed detection now returns false there; M4 (SVL 512) is unaffected.

czoli1976:
https://qemu-project.gitlab.io/qemu/system/arm/cpu-features.html#sme-user-mode-default-vector-length-property

kali (dismissed):

kali:
Thanks for looking into this. FYI, my big work stream around nemotron is finally closing, and i think I'll move on to preparing 0.23.0 for a release soon (this week or next week), and then get back to a more normal "release" cycle with the public API somewhat stable. But if you want something specific in 0.23, please get in touch.

kali:
I'm not sure what to do here. I need to keep an "old glibc build". Should we put sme support in a compilation feature (on by default, i guess ?) wdyt ?

czoli1976:
Good question — and the old-glibc build is exactly the right constraint to anchor on, thanks for flagging it.

Turns out tract already has the machinery for this: the fp16 path uses `ConfigForHalf::works()` in `linalg/build.rs` to probe whether the assembler can assemble fp16 FMLA before building those kernels, since older binutils can't. I followed the same pattern for SME (f7ae64b): `assembler_supports_sme()` does a `try_compile` of a one-line SME snippet, and only on success do we build the SME kernels and set a `tract_sme` cfg that gates `mod sme` / `sme::plug`. When the assembler is too old — like the stretch aarch64 toolchain whose binutils predates SME — SME is skipped and dispatch falls back to the portable path, so the old-glibc build stays green with nothing to remember.

Verified both directions: Apple clang / modern binutils → probe passes, SME built; binutils 2.35 → probe fails (`unknown mnemonic 'smstart'`), SME skipped, build clean.

A `sme` cargo feature (default-on) would work too, and the two compose nicely — feature = "do you want it", probe = "can the assembler do it" — so I'm very happy to add one on top if you'd like an explicit opt-out for minimal builds. I just leaned on the probe first since it keeps the old-glibc build working without special-casing it. Let me know which you'd prefer.

czoli1976:
Related to this
Would you have an M5 to test SME 2.1 ?  I could check that it works with
QEMU but cannot trust it for performance results.

Parallel to this, would you (Sonos, First of all) have any use for
SVE/SVE2  now on in the near future? (Embedded or Cloud like AWS Graviton
Family)


Best Regards
Ckristian Zoli

Email: ***@***.***

kali:
> Good question — and the old-glibc build is exactly the right constraint to anchor on, thanks for flagging it.
> 
> Turns out tract already has the machinery for this: the fp16 path uses `ConfigForHalf::works()` in `linalg/build.rs` to probe whether the assembler can assemble fp16 FMLA before building those kernels, since older binutils can't. I followed the same pattern for SME ([f7ae64b](https://github.com/sonos/tract/commit/f7ae64be6bbc99cc9b86d6c0673a3880a2fc5c65)): `assembler_supports_sme()` does a `try_compile` of a one-line SME snippet, and only on success do we build the SME kernels and set a `tract_sme` cfg that gates `mod sme` / `sme::plug`. When the assembler is too old — like the stretch aarch64 toolchain whose binutils predates SME — SME is skipped and dispatch falls back to the portable path, so the old-glibc build stays green with nothing to remember.
> 
> Verified both directions: Apple clang / modern binutils → probe passes, SME built; binutils 2.35 → probe fails (`unknown mnemonic 'smstart'`), SME skipped, build clean.

I had totally forgotten about that. Works for me, better than configuration features for this IMHO.

kali:
I don't have an M5 machine, and I'm not sure I can find one to play around.

Graviton families are on the map.

czoli1976:
this should be good ... ? I kind of remember the clippy failure was not part of the code for this PR

PR #2238, opened 2026-05-18
linalg/arm64: add f32 _a55 prefetch hints in back-half integer-idle slots

czoli1976:
Sibling to #2231 (which added `_a53` prefetch parity and gave a measured **7% on Sonos's ASR acoustic model**). Same audit pattern applied to `_a55` schedules.

## What

The K-loop-2 back-halves of the f32 `_a55` MMM schedules use `FMLA`+`INS` pairs where `INS` dispatches to NEON pipe 1 (the mov/dup pipe), leaving the integer pipes (E0/E1) idle. `PRFM` (which dispatches to E0/E1) fits cleanly there without displacing any data-flow work.

- `8x8` loop2: +8 PRFM (4 in each K iter)
- `12x8` loop2: +8 PRFM (4 in each K iter)
- `16x4` loop2: +4 PRFM (more conservative — denser INS coverage)
- `24x4` loop2: +4 PRFM (file is identical to 16x4 _a55; mirror)

Less aggressive than the `_a53` cadence (4 per operand) because A55's hardware prefetcher is significantly better than A53's — software hints are supplementary rather than primary.

## Test plan
- [x] `cargo test -p tract-linalg --lib arm64` with `TRACT_CPU_AARCH64_KIND=a55` on M1 Pro: **2124 tests pass, 0 fail**. PRFM is non-semantic so bit-equivalence is guaranteed by construction.
- [ ] **Needs Cortex-A55 bench from Sonos's internal harness to confirm.** Best case: similar production-model gain to what #2231 saw on A53 (though likely smaller given A55's better baseline). Worst case: neutral — PRFM hints are non-faulting and the runtime can ignore them. If the bench shows regression, revert and close.

## Risk
- Zero correctness risk: PRFM is a CPU hint; the change cannot affect output.
- Maintenance: 24 lines added across 4 `.S.raw` files. Pattern matches the existing `_a53` schedules (now post-#2231).

PR #2231, opened 2026-05-18
linalg/arm64: extend f32 _a53 prefetch coverage + fix 16x4 duplicate offset

czoli1976:
## Summary
- Bring `_a53` prefetch coverage in `mmm_f32_12x8`, `mmm_f32_16x4`, `mmm_f32_24x4` up to the 8 PRFM cadence already in `mmm_f32_8x8` `_a53` (4 cache lines ahead on each of A and B).
- Fix what looks like a copy-paste typo in `16x4` `_a53`: duplicate `prfm [x1, #256]` (advances to `[x1, #320]`).
- Fill previously-empty integer-pipe issue slots inside bare-FMLA stretches. A53 can co-issue NEON+ALU each cycle; PRFM is the cheap ALU op to put there.

## Test plan
- [x] `cargo test -p tract-linalg --lib arm64` with `TRACT_CPU_AARCH64_KIND=a53` on M1 Pro: **2124 tests pass, 0 fail**. PRFM is non-semantic so bit-equivalence is guaranteed by construction; the test pass just confirms no compile/link regression.
- [ ] **Needs Cortex-A53 (and ideally A55) bench from Sonos's internal harness** to confirm the projected per-kernel gain. Expected: +5–15% on bandwidth-leaning shapes on A53 (where the L1 prefetcher is weakest); smaller on A55. Cannot reproduce on M1 Pro because its prefetcher already speculates aggressively.

## Risk
- Zero correctness risk: PRFM is a CPU hint; the change cannot affect output.
- Maintenance: 14 lines added, 2 changed across 3 `.S.raw` files. Pattern matches the existing 8x8 _a53 schedule.

PR #2239, opened 2026-05-18
linalg/arm64/fp16: add f16 _a55 prefetch hints (16x8, 32x4)

czoli1976:
Companion to #2238 (f32 `_a55` prefetch). Same audit pattern applied to the f16 `_a55` schedules — both `16x8` and `32x4` loop2 had 0 PRFM today despite having `FMLA`+`INS` pairs in the back half (`INS` dispatches to NEON pipe 1, leaving the integer pipes E0/E1 idle for PRFM hints).

## What
- `16x8` loop2: +8 PRFM (4 per K iter, placed inside the bare-FMLA stretch where each cycle dispatches only 1 NEON op — pure fill, no displaced data-flow work)
- `32x4` loop2: +4 PRFM (2 per K iter, at end-of-iter to avoid interfering with the denser `FMLA`+`INS` section)
- `128x1` GEMV: **unchanged.** Already has 5 PRFM in a GEMV-appropriate cadence (deep A-side prefetch at +1024..+1216, light B-side reuse pattern). Intentionally left alone.

## Models that would benefit on A55 silicon
Realistic f16 transformer workloads sized for A55 (small models, batch=1, short-to-medium seq):
- f16 BERT-class encoders for on-device understanding: DistilBERT, MiniLM, sentence-embedding models (~100-500 MB f16)
- f16 Whisper-tiny encoder (~80 MB f16) — long-seq audio (3000 frames), `32x4` dominant
- f16 Conformer-class ASR (small variants)
- f16 small audio transformers for keyword-spotting / sound classification (AudioMAE-tiny, BEATs-tiny)
- f16 ViT-tiny / small vision transformers for on-device image classification
- The `16x8` kernel covers batch=1 short-seq attention/FFN; `32x4` covers long-seq encoders

Not in scope: on-device LLM inference. A55 is typically a big.LITTLE efficiency core or a low-power IoT/smart-speaker CPU — LLMs run on the P-cores or dedicated NPUs, not A55.

## Why now
Companion to #2223 (RmsNorm F16 fix) which unblocked F16 transformer inference on tract. Once F16 models are running in production, these `_a55` kernels are hot — better to complete the prefetch parity now while audit context is fresh than to leave a known gap.

## Test plan
- [x] `cargo test -p tract-linalg --lib arm64` with `TRACT_CPU_AARCH64_KIND=a55` on M1 Pro: **2124 tests pass, 0 fail**. PRFM is non-semantic so bit-equivalence is guaranteed by construction.
- [ ] **Needs Cortex-A55 bench from Sonos's internal harness, on a f16 workload, to confirm.** Best case: similar ~3% gain to what #2238 targets on f32. Worst case: neutral — PRFM is non-faulting and the runtime can ignore the hints.

## Risk
- Zero correctness risk: PRFM is a CPU hint; the change cannot affect output.
- Maintenance: 12 lines across 2 `.S.raw` files. Pattern matches the just-filed `_a55` f32 PR.

### DIFF
--- a/linalg/arm64/arm64fp16/arm64fp16_mmm_f16_16x8/loop2/cortex_a55.S.raw
+++ b/linalg/arm64/arm64fp16/arm64fp16_mmm_f16_16x8/loop2/cortex_a55.S.raw
@@ -14,9 +14,13 @@ ldr         x6, [x1], #8
 fmla        v23.8h, v1.8h, v4.h[3]
 
 fmla        v24.8h, v0.8h, v4.h[4]
+prfm        pldl1keep, [x1, #256]
 fmla        v25.8h, v1.8h, v4.h[4]
+prfm        pldl1keep, [x1, #320]
 fmla        v26.8h, v0.8h, v4.h[5]
+prfm        pldl1keep, [x2, #256]
 fmla        v27.8h, v1.8h, v4.h[5]
+prfm        pldl1keep, [x2, #320]
 fmla        v28.8h, v0.8h, v4.h[6]
 ins         v2.d[1], x5
 fmla        v29.8h, v1.8h, v4.h[6]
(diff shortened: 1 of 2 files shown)


czoli1976:
Closing as superseded-by-evidence.

#2238 (the f32 sibling, same A55 micro-arch, same speculative PRFM-fill pattern) just got benched by @kali and showed a clear ~10% regression on two critical audio models. The microbench data he shared shows A55 already perfectly co-issues fma+ins at 1.0 cycle/pair — so the FMLA+INS slots I targeted as "integer pipe idle" aren't actually idle on A55, and adding PRFM displaces co-issued work into serial cycles.

The same pattern in this f16 PR would almost certainly regress for the same reason. Not worth burning more bench cycles to confirm.

The A55 prefetch-audit angle is closed. The A53 sibling #2231 stands — A53 doesn't share A55's good fma+ins co-issue, so the trick really is A53-specific.

PR #2253, opened 2026-05-20
core/cnn: direct register-blocked conv for small-ocg grouped convs

czoli1976:
## Summary

Draft for feedback on **approach**, not (yet) merge-ready.

Adds a direct, register-blocked convolution for a narrow shape class — **NCHW,
kernel-width 1, grouped, with a small number of output channels per group** —
where the im2col + matmul lowering is inefficient. On WASM it's a real (if
modest) win on DeepFilterNet 3's decoder; on native it's gated off. This is the
"direct conv" follow-on flagged in #2226's notes.

**What I'd most like your read on:** whether this narrow, target-gated op is the
right shape, or whether you'd rather hold out for unifying the eager/lazy/blocked
choice under one cost model (the open question from #2226). My honest take on
that is at the bottom — I think the cost model is the *elegant* answer but I'm
not convinced it's *safely achievable*, and I'd rather ship the conservative op
than a cost model I can't trust. Keen for your call.

## Motivation — the shape

DFN3 `df_dec`'s `df_convp.1`: input `[1,64,S,96]`, weights `[10,32,5,1]`,
`group=2`, kernel `[5,1]`, causal pad `[4,0,0,0]`. Via im2col each group is a
matmul **M=5 (ocg) · K=160 (icg·kh) · N=9600 (H·W)** — the tiny M=5 wastes the
matmul's m-tile (same pathology as a low-M GEMV), plus the gather is overhead.
ORT runs it as a direct conv and avoids both.

## Approach

A new `BlockedConv` op (`core/src/ops/cnn/conv/blocked.rs`): for each
`(group, output-row, WB-wide block of the contiguous W axis)` it holds the `ocg`
output accumulators in registers and reduces over `(kh, icg)`, loading each
input row once and reusing it across all `ocg` outputs (the input-reuse a GEMM
gets). Dispatched by const-generic `OCG`; the hot loop touches the accumulators
only at compile-time-constant offsets so LLVM keeps them register-resident
across the runtime `(kh, icg)` reduction. `Conv::codegen` routes eligible convs
here (before lazy im2col); everything else falls back to im2col unchanged.

## Performance (S=100, single-thread; wasm = wasmtime relaxed-simd, native = AMX)

| | im2col | blocked | Δ |
|---|---:|---:|---|
| **WASM** `df_convp.1` (isolated) | 2.21 ms | **1.76 ms** | **−20%** |
| **WASM** `df_dec` (full) | ~8.75 ms | **~8.29 ms** | **−5.3%** |
| native `df_convp.1` | 1.39 ms | 1.68 ms | +21% (regression) |
| native `df_dec` | 6.02 ms | 6.26 ms | +4% (regression) |

- df_dec WASM gap to ORT-Web closes from ~1.13× to ~1.07×.
- Native regresses (your #2220 AMX dispatch already handles the M=5 matmul), so
  the path is **wasm-default-on / native-opt-in**
  (`TRACT_DISABLE_BLOCKED_CONV` / `TRACT_ENABLE_BLOCKED_CONV`). All numbers are
  M1 Pro; they'll differ on your hardware.

## Correctness

- New unit test: bit-exact vs a scalar reference across 5 shape variants
  (full-block, full+remainder, remainder-only, group=1, ocg=1).
- df_dec real-input reference parity: **8/96000 outliers ~1e-7, identical to the
  im2col path** (the known f32-precision points).

## Eligibility / safety

Routes only when: f32, NCHW, `kernel_width == 1`, unit stride/dilation on W,
grouped, `ocg ≤ 8`, concrete input shape. Anything else → im2col (unchanged).
Off-switch env var on both targets. Typical CNN convs (3×3, or group-1 pointwise
with large ocg) don't match, so they're untouched — though I'd add an explicit
canary reroute check before this leaves draft.

## On the cost-model alternative (your #2226 question)

Unifying eager / lazy / blocked under a tile-aware cost model is the principled
version of all these gates — and it would make the native/wasm split fall out of
the model instead of a `cfg`. The core idea: cost each candidate, with the
matmul candidates paying for the *wasted* m-tile (`ceil(m/mr)·mr·k·n`).

**But I'm not confident it's safely doable, and I'd rather flag that than ship a
shaky cost model:**

- My evidence that "the model derives the right routing" is **one shape**
  (`df_convp.1`). Fitting to n=1 is a lookup table, not a model.
- The throughput "constants" are really shape-dependent functions. **AMX is the
  sticking point: its cost has cliffs, not a smooth FLOP curve — #2220 exists
  precisely because AMX dispatch had to be made shape-aware.** A cost model
  predicting AMX matmul time would hit the same non-smoothness.
- The lazy-im2col gather cost already resisted a clean model in #2226 (hence the
  2-feature threshold).
- Constants won't transport across M1 / M4 / Cortex-A / x86 / wasm-engines
  without runtime calibration (a project in itself). The current target-*family*
  gate is crude but robust.
- The current heuristics are *conservative* (additive: #2226 only routes
  eager→lazy; this op is narrowly gated) so they can't catastrophically
  mis-route; a cost model is more powerful **and** can regress unseen shapes.

So my recommendation: **land the narrow gated op first** (it's the candidate the
cost model would need to exist anyway), and treat the cost model as a separate,
evidence-gated effort — I'd want to bench a sweep of shapes × targets and show a
simple model ranks them correctly *before* replacing your heuristics. Happy to
run that sweep if you think it's worth pursuing.

## Test plan

- [x] Unit test: bit-exact vs scalar reference (5 shape variants)
- [x] df_dec real-input reference parity (= im2col)
- [ ] Explicit canary-CNN reroute check (MobileNet / SqueezeNet / Inception)
- [x] `cargo clippy` clean

### DIFF
--- a/core/src/ops/cnn/conv/blocked.rs
+++ b/core/src/ops/cnn/conv/blocked.rs
@@ -1,0 +1,376 @@
+//! Direct, register-blocked convolution for the "channel-mixing temporal conv"
+//! shape class: NCHW, kernel width 1 (extent only on H), unit stride/dilation on
+//! the contiguous W axis, grouped, with a *small* number of output channels per
+//! group (`ocg`).
+//!
+//! For such convs the im2col + matmul lowering is inefficient: the per-group
+//! matmul is `M = ocg` (tiny, e.g. 5) × `K = icg·KH` × `N = H·W`, so the matmul
+//! kernel's m-tile is mostly wasted — exactly the same pathology as a low-M GEMV.
+//! ORT side-steps it with a direct conv.
+//!
+//! This op computes the conv directly: for each (group, output-row, block of the
+//! contiguous W axis) it holds `ocg` accumulators in registers and reduces over
+//! `(kh, icg)`, loading each input row ONCE and reusing it across all `ocg`
+//! outputs (the same input-reuse a GEMM gets). Measured on df_dec's `df_convp.1`
+//! (group=2, 64→10ch, kernel [5,1], 100×96): 0.77 ms native / 0.79 ms wasm vs
+//! 1.72 / 2.42 ms for tract's lazy im2col and 1.13 ms for ORT — a 2.2–3.1× win,
+//! bit-exact.
+//!
+//! Eligibility is checked in `Conv::codegen`; anything outside the supported
+//! shape class falls back to im2col.
+
+use crate::internal::*;
+
+/// Width of the inner SIMD-vectorised block over the contiguous W axis.
+const WB: usize = 16;
+
+/// Direct blocked conv. Inputs: X [N, C, H, W] (NCHW, f32), kernel
+/// [OC, ICG·KH] (group-major: row `oc` holds its group's `icg·KH` weights,
+/// i-major/h-minor), bias [OC]. Output [N, OC, H_out, W].
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+pub struct BlockedConv {
+    pub n: usize,
+    pub c_in: usize,
+    pub h_in: usize,
+    pub w: usize,
+    pub oc: usize,
+    pub group: usize,
+    pub kh: usize,
+    pub stride_h: usize,
+    pub dil_h: usize,
+    pub pad_before_h: usize,
+    pub h_out: usize,
+}
+
+impl BlockedConv {
+    #[inline]
+    fn icg(&self) -> usize {
+        self.c_in / self.group
+    }
+    #[inline]
+    fn ocg(&self) -> usize {
+        self.oc / self.group
+    }
+}
+
+impl Op for BlockedConv {
+    fn name(&self) -> StaticName {
+        "BlockedConv".into()
+    }
+
+    fn info(&self) -> TractResult<Vec<String>> {
+        Ok(vec![format!(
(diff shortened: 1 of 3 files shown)


czoli1976:
This takes TRACT/WASM Veeeeeery close to ORT WEB, I am obsessed at beating ORT but not sure where to look at anymore :)

czoli1976:
clippy is happy too now

kali (approved):

kali:
Same as in 2226. Happy to proceed with simple euristics that we can actually understand for as long as we can. I went the cost model way for Cortex-Axx MM kernel selection, and I don't think I want to do it again.

czoli1976:
PLMK if it looks good enough to move to OPEN, as usual, no rush intended

kali:
Bench is neutral on this one. Still happy with it ?

czoli1976:
Rebasing, rebenching on WASM (that's where the gains materialised before)

czoli1976:
The win SURVIVES the rebase. On current upstream/main + BlockedConv:

image

Why neutral: BlockedConv is opt-in on native (TRACT_ENABLE_BLOCKED_CONV) because **it regresses there** —  PR #2220 AMX dispatch already handles the M=5 matmul. So a native-default bench leaves the op dormant

Force-Pushed a Rebased, let's keep it, works for WASM

PR #2259, opened 2026-05-20
harness/nemotron: allow DiagGather + ScaledMaskedSoftmax on GPU op-only assert

czoli1976:
## Summary

Unblocks the `macOS / Nemotron speech streaming en 0.6b` (and `cuda-lovelace`) Large-models job, which currently fails on `main` with:

```
ERROR tract] Model has 48 unexpected op(s):
   _N_selfAttn_matrixBdSlice0_dyn_slice_output (DiagGather)
   _N_selfAttn_softmax0.scaled_masked_softmax (ScaledMaskedSoftmax)
```

The encoder's ROI attention emits `DiagGather` + `ScaledMaskedSoftmax`. These have no Metal/Cuda implementation, so they fall back to CPU — but the nemotron `--metal` / `--cuda` `--assert-op-only` allowlists in `ci.sh` don't list them, so the strict op-coverage assert fails on the GPU runtimes. (The CPU `-O` pass of all four sub-models passes fine.)

`harness/parakeet-tdt-600m-v3/ci.sh` already allows `DiagGather`; this mirrors that for nemotron and additionally adds `ScaledMaskedSoftmax` (nemotron's chunk-window-mask encoder uses it; parakeet doesn't).

If you'd rather these get Metal/Cuda impls than be CPU fallbacks, happy to drop this — but as-is it unblocks every PR branched off current `main` (the job is currently red on e.g. #2257 and task/onnx-dtype-panic, neither of which touch these ops).

## Test plan
- [ ] `macOS / Nemotron speech streaming en 0.6b` green
- [ ] `cuda-lovelace / Nemotron speech streaming en 0.6b` green

### DIFF
--- a/.github/workflows/large_models.yml
+++ b/.github/workflows/large_models.yml
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/checkout@v6
       - run: |
           ROOT=. ./.travis/ci-system-setup.sh
-          cargo build -p tract-cli --profile opt-no-lto --no-default-features --features transformers
+          cargo build -p tract-cli --profile opt-no-lto --no-default-features --features transformers,pulse
       - run: echo uname=$(uname) >> $GITHUB_ENV
       - uses: actions/upload-artifact@v7
         with:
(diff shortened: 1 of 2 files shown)


kali:
Sorry, took a shortcut. Thanks for the fix.

kali (dismissed):

kali:
There's a second one hidden behind. I'm on it.

kali:
Apologies.

PR #2268, opened 2026-05-21
linalg/arm64/sve: VLA SVE/SVE2 backend for ARMv9 — f32/f16/int8 GEMM+GEMV

czoli1976:
# linalg/arm64/sve: SVE/SVE2 backend for ARMv9 cores (Graviton, Neoverse, Cortex-X)

## Summary

Adds a vector-length-agnostic (**VLA**) SVE/SVE2 matmul backend to tract's
`linalg`, covering **all of tract's f32 / f16 / int8→i32 GEMM + GEMV hot-path
slots** (6 kernels, one commit each). Gated on `FEAT_SVE2` (+ `FEAT_FP16` for the
f16 kernels), Linux-aarch64 only — targets ARMv9 server/mobile silicon (AWS
Graviton 3/4, Neoverse V1+/N2+, Cortex-X2+/A510+). No Apple silicon has SVE, so
macOS keeps AMX/NEON. `TRACT_SVE_DISABLE=1` turns the whole backend off for A/B.

Every kernel is **vector-length-agnostic**: tiles are walked in `svcnt*()` chunks
with `whilelt` predication, so one binary is correct *and* full-width at any VL
from 128 to 2048-bit, with no per-VL gate (cf. the SME backend's SVL=512 hardcode
+ RDSVL gate).

## Kernels (`linalg/arm64/sve/*.c`, C intrinsics — Rust has no stable SVE intrinsics)

| slot | kernel | tile | inner step |
|---|---|---|---|
| `mmm_f32` | `sve_mmm_f32` | 8×8 | broadcast-A rank-1 `svmla` |
| `mmv_f32` | `sve_mmv_f32_64x1` | 64×1 | GEMV, vectorized over M |
| `mmm_f16` | `sve_mmm_f16` | 8×8 | native f16 FMA (`svmla_n_f16`) |
| `mmv_f16` | `sve_mmv_f16_64x1` | 64×1 | f16 GEMV |
| `qmmm_i32` | `sve_mmm_i32` | 8×8 | int8→i32 widening MLA (`svld1sb`+`svmla`) |
| `qmmv_i32` | `sve_mmm_i32_64x1` | 64×1 | int8 GEMV |

All consume tract's **native K-major packing** — no custom packer. The i32
kernels port the `q_scale`/`q_shr`/`q_shl` quantization fuse ops bit-exact from
`generic/rounding.rs`. (SDOT/SMMLA max-throughput int8 would need a custom
interleaved K-contiguous packer and is a documented follow-up, as is bf16
BFMMLA, for which tract has no input slot yet.)

## Build wiring

- `build.rs`: `include_sve()` (aarch64-linux) + a `compiler_supports_sve()`
  `try_compile` probe (skips toolchains lacking `arm_sve.h`, e.g. old Debian
  stretch cross-gcc) → compiles the kernels and sets the `tract_sve` cfg. f16
  kernels are compiled in a separate `+fp16` unit so the `+sve`-only kernels
  never gain fp16 codegen. Same probe pattern as the SME assembler check.
- `arm64/sve.rs`: `has_sve()`/`has_sve2()` via HWCAP (false on macOS),
  `rdvl_bytes()` for optional VL-matched dispatch; kernels registered
  `where(SVE2)` / `where(SVE2 && fp16)`, `can_fuse` excludes LeakyRelu (f32/f16)
  and additionally keeps the quant ops for i32; `plug()` forces the SVE kernels
  onto the mmm/mmv/qmmm/qmmv slots under `has_sve2()`.

## Validation (under QEMU — no SVE hardware, so correctness only, no perf signal)

- **448/448 auto-tests** on emulated SVE: per-kernel `MMMRustKernel!` proptest
  matmul sweep (both packings), every fuse op, all six `q_scale` rounding
  policies for i32, and all store layouts.
- **Bit-exact at every VL** 128/256/512/1024/2048-bit (standalone per-kernel
  harnesses driving real `FusedKerSpec` arrays).
- **Cross-environment agreement (same workloads run on every backend).** On the
  `nvidia/nemotron-speech-streaming-en-0.6b` f32 sub-models, the **decoder** and
  **joint** outputs are **bit-identical** across M1-AMX, QEMU-SVE, QEMU-SME
  (SVL=512) and QEMU-NEON; the preprocessor agrees to **1.9e-6** (≈ f32 epsilon,
  and the divergence is AMX-only — SVE/SME/NEON are byte-identical to each other).
  An **exact int8 matmul checksum** is identical (`31f0…b853`) across M1-scalar,
  QEMU-scalar, QEMU-SVE (SDOT, all VLs) and QEMU-SME (SMOPA).

**Honest scope: draft / correctness-only.** Real throughput needs Graviton 3/4
hardware. The f32 GEMM was the original commit; this PR stacks the f16 + int8
GEMM and all three GEMV kernels on top, one commit per kernel.

### DIFF
--- a/linalg/arm64/sve/sve_mmm_f16.c
+++ b/linalg/arm64/sve/sve_mmm_f16.c
@@ -1,0 +1,153 @@
+// SVE f16 GEMM kernel for tract's MMM framework (the mmm_f16 slot).
+//
+// Tile MR=8 x NR=8 with native f16 accumulation, the f16 sibling of
+// sve_mmm_f32.c. The hot AddMatMul is the same vector-length-agnostic
+// broadcast-A rank-1 update, but over f16 lanes: the NR columns are walked in
+// svcnth() chunks with whilelt predication and folded with svmla_n_f16 (native
+// f16 fused multiply-add, as the NEON arm64fp16 kernels do), so one binary is
+// correct and full-width at any SVE VL 128..2048-bit.
+//
+// Gated on FEAT_SVE2 AND FEAT_FP16 (Rust side). Built with +fp16. Consumes
+// tract's native f16 K-major packing. Fuse ops act on the MRxNR tile in memory
+// (scalar; not the hot path); as with the f32 kernel, LeakyRelu and the i32
+// quantization ops are excluded by CAN_FUSE. Returns 0 on success, 1 otherwise.
+
+#include <arm_sve.h>
+#include <stdint.h>
+#include <string.h>
+
+#define MR 8
+#define NR 8
+
+enum {
+    DONE = 0, CLEAR, LOAD_TILE,
+    SCALAR_MIN, SCALAR_MAX, SCALAR_ADD, SCALAR_MUL, SCALAR_SUB, SCALAR_SUBF,
+    LEAKY_RELU,
+    PER_ROW_MIN, PER_ROW_MAX, PER_ROW_ADD, PER_ROW_MUL, PER_ROW_SUB, PER_ROW_SUBF,
+    PER_COL_MIN, PER_COL_MAX, PER_COL_ADD, PER_COL_MUL, PER_COL_SUB, PER_COL_SUBF,
+    Q_SCALE, Q_SHR, Q_SHL,
+    ADD_UNICAST, ADD_ROW_COL_PRODUCTS, STORE, ADD_MAT_MUL
+};
+
+typedef struct {
+    uint64_t disc;
+    uint64_t f0, f1, f2, f3;
+} spec_t;
+
+static inline __fp16 f16_of(uint64_t bits) {
+    __fp16 f;
+    uint16_t lo = (uint16_t)bits;
+    memcpy(&f, &lo, 2);
+    return f;
+}
+
+// AddMatMul: ab[m][n] += sum_k pa[k*MR+m] * pb[k*NR+n]. VLA over NR (f16 lanes).
+static void add_mat_mul(__fp16 ab[MR][NR], const __fp16 *pa, const __fp16 *pb, long k) {
+    for (long n0 = 0; n0 < NR; n0 += svcnth()) {
+        svbool_t pg = svwhilelt_b16((uint64_t)n0, (uint64_t)NR);
+        svfloat16_t a0 = svld1_f16(pg, &ab[0][n0]), a1 = svld1_f16(pg, &ab[1][n0]);
+        svfloat16_t a2 = svld1_f16(pg, &ab[2][n0]), a3 = svld1_f16(pg, &ab[3][n0]);
+        svfloat16_t a4 = svld1_f16(pg, &ab[4][n0]), a5 = svld1_f16(pg, &ab[5][n0]);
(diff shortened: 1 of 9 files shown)


kali (approved):

czoli1976:
@kali running some more tests

* SDOT/SMMLA max-throughput int8 — the "maximum-advantage" SVE2 path (4–8× density), but it needs a new custom interleaved K-contiguous packer in tract's framework, and the payoff is unmeasurable without real Graviton hardware.
* bf16 BFMMLA — tract has no bf16 input slot to plug into.

Is it something you have plans to do in TRACT and/or see the need?

czoli1976:
## CPUs that would benefit from this backend

The kernels activate on **`FEAT_SVE2` + Linux + aarch64** (the `has_sve2()` gate plus `target_os = "linux"` in `build.rs`). SVE2 is mandatory in ARMv9-A, so this is effectively the ARMv9 Linux fleet. As noted in the PR, validation is correctness-only (no SVE hardware was available), so this is about *applicability*, not measured speedups.

### Cloud / server (primary target)

| CPU | Core |
|---|---|
| AWS Graviton 4 | Neoverse V2 |
| NVIDIA Grace (Grace-Hopper / Grace-Blackwell) | Neoverse V2 |
| Microsoft Azure Cobalt 100 | Neoverse N2 |
| Google Axion | Neoverse V2 |
| Alibaba Yitian 710 | Neoverse N2 |
| Any Neoverse V2 / N2 / V3 / N3 design on Linux | — |

### Non-cloud (client / edge / Linux on ARMv9)

- Linux running on ARMv9 cores — **Cortex-X2/X3/X4/X925**, **A710/A715/A720**, **A510/A520** (dev boards, edge/embedded, future Linux client).
- **NVIDIA Jetson Thor** (Neoverse-V3AE, ARMv9) — note *not* Jetson Orin, which is Cortex-A78AE (ARMv8.2, no SVE).
- **Android flagships ship the silicon but are build-gated out today**: Snapdragon 8 Gen 1+/8 Elite, MediaTek Dimensity 9000+, Samsung Exynos 2200+, Google Tensor G3+ all have SVE2 cores, but `target_os = "android" ≠ "linux"`, so they don't compile the kernels. Reachable by relaxing the OS gate.

### Not covered (and why)

- **Apple silicon (M1–M4)** — no SVE at all; stays on AMX/NEON (`has_sve()` is hardcoded `false` on macOS).
- **SVE *v1* only** — AWS Graviton 3 / Neoverse V1 (256-bit) and Fujitsu A64FX (512-bit): excluded by the `SVE2` gate, even though these kernels use only base-SVE intrinsics. A one-line relaxation (`where(SVE2)` → `where(SVE)`, `has_sve2()` → `has_sve()`) would light them up.
- **Pre-SVE cores** — Cortex-A53/A55/A73/A7/A9 etc. (no SVE).

Most current SVE2 parts run a **128-bit vector length** (same width as NEON), so the gain there comes from predication, the larger register file, and VLA dispatch rather than raw width; the big-VL wins (256–2048-bit) live on the SVE1 parts (Graviton 3, A64FX) and future wide cores.

czoli1976:
Should be okey as an EXPERIMENTAL Addition with a call for the community to test it on suitable HW (Graviton especially)

czoli1976:
Un petit cadeau

kali:
Merci !

PR #2273, opened 2026-05-22
linalg/arm64/sme: experimental f16 SME kernels for Apple M5/A19

czoli1976:
## Summary

Addendum to the merged f32 SME backend (#2230): **experimental** half-precision SME kernels for the `mmm_f16` / `mmv_f16` slots, using the non-widening `fmopa za.h` path (`FEAT_SME_F16F16`).

> [!WARNING]
> **This is effectively Apple M5 / A19-only and has NOT been validated on real hardware.** It is QEMU-correctness-validated only and **needs community testing on an actual M5 / A19 (iPhone 17)**. It cannot regress anything else (see *Build safety*), so it's offered as a low-risk, opt-in-by-hardware addition.

## Why "M5 / A19-only" — the LLVM smoking gun

`FEAT_SME_F16F16` (non-widening f16→f16 in ZA.H) is an *optional* SME2 sub-feature. Among shipping silicon, **only the Apple M5 / A19 implement it.** The proof is upstream LLVM's own CPU definition — [`llvm/llvm-project` commit `f85494f6afeb` "Define apple-m5/a19 CPUs"](https://github.com/llvm/llvm-project/commit/f85494f6afeb):

```
def TuneAppleM5 : SubtargetFeature<"apple-m5", ...,
    FeatureSME, FeatureSME2, FeatureSMEF64F64, FeatureSMEI16I64,
    FeatureSME2p1, FeatureSMEB16B16, FeatureSMEF16F16, ...>;
//  --print-enabled-extensions -mcpu=apple-m5  →  FEAT_SME_F16F16
//  apple-a19 is an alias of apple-m5
```

In LLVM's *entire* AArch64 CPU table, `FeatureSMEF16F16` appears for **apple-m5/a19 and nothing else**:

| Silicon | SME level | `FEAT_SME_F16F16` | These kernels |
|---|---|---|---|
| **Apple M5 / A19** (iPhone 17) | SME2p1 | **yes** | **run natively** |
| Apple M4 / A18 | SME2 | **no** (verified `sysctl=0` on an M4) | dormant → AMX f16 |
| Arm C1 / Lumex (Exynos 2600), Cortex‑X925 | SME2 | no | dormant |
| Qualcomm Oryon (SD 8 Elite Gen 5) | SME (v1) | no | dormant |
| Neoverse / Graviton | none | — | — |

So this can only be exercised on an M5/A19, which I don't have — hence *experimental, community testing requested*.

## What it adds

- **`sme_mmm_f16_32x32`** (GEMM): one 32×32 `ZA.H` tile, `fmopa za.h` per K‑step (at SVL=512 a single f16 FMOPA covers the whole 32×32 tile = the f32 kernel's 4‑tile MAC count). `where(SME_F16F16)`.
- **`sme_mmv_f16_64x1`** (GEMV, N==1): vgx2 `ZA.H` group, SME2 multi‑vec `fmla za.h[w8,0,vgx2]`. `where(SME2 && SME_F16F16)`.
- Both consume tract's native K‑major f16 packing — no custom packer. `CAN_FUSE` matches the f32 kernels.

## Build safety (cannot regress the f32 SME backend)

The f16 unit needs the `+sme-f16f16` assembler extension. A separate `dummy_sme_f16f16.S` probe + `assembler_supports_sme_f16f16()` compiles the f16 kernels as their own object behind a new **`tract_sme_f16f16`** cfg. A toolchain with base SME but not f16f16 still builds the f32 SME kernels exactly as before — only the f16 unit (and all its Rust registration / detection / `plug()` wiring) is gated off. Detection is `HWCAP2_SME_F16F16` (bit 42) on Linux / the `hw.optional.arm.FEAT_SME_F16F16` sysctl on macOS, plus the existing 512‑bit SVL check.

## Validation

- **QEMU only** (no F16F16 hardware): `qemu-aarch64 -cpu max,sme512=on` → full SME auto-test surface **220/220** (the two new f16 kernels — matmul proptest, every fuse op, store layouts, frame — plus the existing f32 GEMM/GEMV, no regression).
- Standalone probes confirmed `fmopa za.h` and the multi-vec `fmla za.h vgx2` forms are bit-correct.
- Builds clean on macOS (Apple clang) and debian:sid (gcc 15).
- **Not** run on real `FEAT_SME_F16F16` silicon — that's the ask.

## Request

If you (or anyone) have an **M5 Mac or an iPhone 17 (A19)**, a run of `cargo test -p tract-linalg arm64::sme::test_sme_mmm_f16` / `test_sme_mmv_f16` (or an f16 model A/B vs the AMX path) would confirm real-hardware correctness and let this graduate from experimental.

### DIFF
--- a/linalg/arm64/sme/dummy_sme_f16f16.S
+++ b/linalg/arm64/sme/dummy_sme_f16f16.S
@@ -1,0 +1,16 @@
+// Build-time capability probe for FEAT_SME_F16F16, used by build.rs
+// (assembler_supports_sme_f16f16). The experimental f16 SME kernels use the
+// non-widening half-precision outer product `fmopa za.h`, which needs the
+// +sme-f16f16 assembler extension. An assembler that supports base SME but not
+// FEAT_SME_F16F16 (so the basic dummy_sme.S probe passes but this one fails)
+// must still build the f32 SME kernels — so this is a SEPARATE probe gating
+// only the f16 unit + the `tract_sme_f16f16` cfg. Not linked into anything.
+.arch armv9-a+sme2+sme-f16f16
+.text
+.globl tract_sme_f16f16_probe
+tract_sme_f16f16_probe:
+    smstart
+    ptrue   p0.h
+    fmopa   za0.h, p0/m, p0/m, z0.h, z1.h
+    smstop
+    ret
--- a/linalg/arm64/sme/sme_mmm_f16_32x32.S.j2
+++ b/linalg/arm64/sme/sme_mmm_f16_32x32.S.j2
@@ -1,0 +1,291 @@
+// vim: ft=arm
+//
+// SME f16 32x32 matmul kernel (FEAT_SME_F16F16).
+//
+// Uses the non-widening half-precision outer product `fmopa za.h` (f16 inputs,
+// f16 accumulate), which mirrors the f32 32x32 kernel's structure exactly but
+// in a single ZA.H tile: at SVL=512 a ZA.H tile is 32x32 f16, so one FMOPA
+// covers the whole MR=32 x NR=32 output (1024 MACs/insn = the f32 kernel's
+// 4-tile MAC count). K-major packing is consumed directly (one ld1h = 32 f16 =
+// one K-step of A or B), so no custom packer is needed.
+//
+//   ZA0.H : C[0..32, 0..32]   (the entire 32x32 f16 tile)
+//
+// Calling convention (extern "C", AAPCS64):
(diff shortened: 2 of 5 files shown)


czoli1976:
Leaving this one alone for now — conflicts here interact with the SME dispatch machinery that's evolved on main since, and I have no way to verify SME2/F16F16 kernels on available hardware (no M5/A19 yet). Will revisit when I can actually test it.

czoli1976:
@kali anow that the M6 has been announced, should we have these M5+ in?

PR #2276, opened 2026-05-24
onnx: fix LayerNorm output dtype mismatch with F16 inputs

dqj1998:
The LayerNorm op's `wire` expansion casts `normalized` back to fact.datum_type *before* applying scale/bias, then multiplies that result with `cast_scale` (which is still in self.datum_type, F32).

With F16 inputs this becomes F16 × F32, whose output is downgraded to F32 by `mul()`. The inference rule then asserts
`outputs[0].datum_type == inputs[0].datum_type` (F16) against the actual F32 output, failing `into_typed()` with:

    Output mismatch after rewiring expansion for output #0:
    expected 1,256,384,F16 got 1,256,384,F32

Fix: defer the cast back to fact.datum_type until after all scale/bias operations. Now the expansion stays entirely in self.datum_type (F32) through normalized × scale + bias, and casts only the final result.

Behavior is unchanged for F32 inputs (the final cast is a no-op when fact.datum_type == self.datum_type).

Reproduced with sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2 exported via `optimum.exporters.onnx.main_export(..., dtype="fp16")` and loaded with `into_optimized().into_runnable()`.

### DIFF
--- a/onnx/src/ops/nn/layer_norm.rs
+++ b/onnx/src/ops/nn/layer_norm.rs
@@ -182,11 +182,7 @@ impl Expansion for LayerNorm {
         } else {
             normalized_scaled
         };
-        let y = model.wire_node(
-            format!("{prefix}.cast_y"),
-            cast(fact.datum_type),
-            &y_internal,
-        )?;
+        let y = model.wire_node(format!("{prefix}.cast_y"), cast(fact.datum_type), &y_internal)?;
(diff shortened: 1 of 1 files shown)


kali (dismissed):

kali:
Hey, thank for the help ! Can we get a fmt, please ?

kali (approved):

PR #2280, opened 2026-05-25
wasm: SIMD int8 (i8->i32) matmul kernel (wasm_i32_4x4)

czoli1976:
## Summary
WASM had **no int8 matmul kernel** — i8 (i32-accumulator) matmul fell back to the generic *scalar* kernel. This adds **`wasm_i32_4x4`**, a `simd128` int8 → i32 matmul kernel.

## What's here
- **`wasm_i32_4x4`** — the `AddMatMul` K-loop widens i8→i32 and accumulates with `i32x4` mul/add (an SMLAL-style outer product). The quant epilogue (`q_scale`/`q_shr`/`q_shl`) + fuse ops reuse the bit-exact scalar path — O(MR·NR), negligible vs the O(MR·NR·K) inner loop. **Bit-identical to `generic_i32_4x4`**; selected for i8 matmul via `ManuallyOptimized` quality (over the generic scalar one).

## Why widening multiplies, not the relaxed dot
WASM's only integer dot, `i32x4.relaxed_dot_i8x16_i7x16_add`, treats its 2nd operand as **i7** (`[-64,63]`) and is **non-deterministic for full i8** (engines may compute i7×i8 or i8×i8). So for a bit-exact, deterministic kernel this uses widening multiplies instead. A relaxed-dot fast path (for i7-range weights) is a possible follow-up.

## Validation
- `wasm_i32_4x4`: **114/114** on wasmtime (`+simd128`).
- Full `tract-linalg` wasm suite: **1870/1870**.
- `cargo fmt --check` clean. (clippy: the index-loop / `offset`-cast lints match the existing `wasm_f32_*` kernels.)

## Perf
The inner loop is 4-wide `i32x4` vs the prior element-at-a-time scalar fallback. A precise wasmtime throughput bench is a follow-up — but this replaces a pure-scalar path, so it's a clear win for int8 inference on WASM.

### DIFF
--- a/linalg/src/wasm.rs
+++ b/linalg/src/wasm.rs
@@ -66,6 +66,10 @@ pub fn plug(ops: &mut Ops) {
     ops.mmm_impls.push(wasm_f32_16x1.mmm());
     ops.mmm_impls.push(wasm_f32_32x1.mmm());
     ops.mmm_impls.push(wasm_f32_8x8.mmm());
+    // int8 -> i32 matmul: SIMD kernel (was generic scalar). ManuallyOptimized so
+    // strategize's retain() keeps it over generic_i32_4x4 for i8 packing.
+    ops.mmm_impls.push(wasm_i32_4x4.mmm());
+    ops.qmmm_i32 = Box::new(|_, _, _| wasm_i32_4x4.mmm());
     // Selection paths. Both rely on kernel_selection::strategize honouring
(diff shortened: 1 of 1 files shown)


czoli1976:
@kali , not happy of this,  LLVM auto-vectorises pretty well the existing one, so mine is barely faster, and not in 100% of the cases, attempting a elaxed-dot instead .... hence closing this one

PR #2281, opened 2026-05-25
wasm: int8 matmul + conv via relaxed-dot SIMD (~2.3x)

czoli1976:
## Summary

A WASM SIMD int8→i32 matmul **and** conv kernel using `i32x4_relaxed_dot_i8x16_i7x16_add` (an SDOT-analog) under `+relaxed-simd`, **~2.3× e2e** over the autovectorized generic kernel; with a deterministic widening fallback under `+simd128`. Gated exactly like the existing f32 `relaxed_madd` (#2199). Includes a K=4-inner packing (`PackedI8K4`) and the lowering changes to route it through both einsum-matmul and conv-im2col.

## Why relaxed-dot (the key finding)

Disassembling the compiled wasm shows **LLVM already auto-vectorizes the generic `i32` 4×4 kernel** into a widening SIMD outer-product (`i16x8.extend` + `i32x4.mul/add`). So a hand-written *deterministic* SIMD int8 kernel is at **parity** with the fallback (measured 0.94–1.15×). The relaxed-dot instruction is the only lever that beats it — 16 i8 MACs/instruction, no widening:

| measurement (wasmtime, Apple Silicon) | result |
|---|---|
| inner-loop 4×4 tile: relaxed-dot vs widening | **3.8–4.25×** |
| full 4×4-tiled GEMM e2e: `wasm_i32_4x4` vs `generic_i32_4x4` | **2.13–2.40×** (square / transformer / CNN shapes) |

## Determinism — 7-bit weights for guaranteed cross-engine exactness

Same relaxed-simd contract as the f32 `relaxed_madd` path (#2199), gated identically (`+relaxed-simd`; deterministic widening fallback under `+simd128`).

- **Default (full int8): bit-exact on wasmtime and every engine** that lowers `relaxed_dot` to SDOT (ARM) / `vpdpbusd` VNNI (x86) — verified here. The spec only permits divergence via i16-intermediate overflow, which the sum reaches only at the int8 extreme (−128) on a strict `pmaddubsw` lowering.
- **Guaranteed on *every* conforming engine: 7-bit weights (±63).** With the `relaxed_dot` i7 operand ≤63 the intermediate cannot overflow, so the result is bit-identical everywhere. **Recommended mode for cross-engine reproducibility.**

7-bit is a **weight-quantization (deployment) choice, not a kernel clamp.** Clamping inside the kernel would compute `A·clamp(W)` ≠ the true int8 product — breaking tract's optimize==reference invariant *and* the full-range quant-matmul proptests (which use full-range `any::<i8>()` weights with a ±1 tolerance). So the kernel computes the true product (exact), and determinism is obtained by feeding it ≤63 weights.

**Optional follow-up (opt-in enforcement):** a cargo-gated packer that clamps weights to ±63 + a conv operand-swap so the i7 operand is always the weights — for users who want tract to *enforce* 7-bit rather than rely on input quantization. Held out of this PR because it changes numerics and is a maintainer/quantization-policy decision.

## Implementation

- **`PackedI8K4`** (linalg): K=4-inner packing `out[(k/4)*r*4 + m*4 + k%4]`, `k_alignment=4`; `pack_view` (matmul) + `KOut4Writer`/`write_with_k_outer` (conv im2col, K-outer feed).
- **`wasm_i32_4x4`**: `cfg`-selected `AddMatMul` — relaxed-dot under `+relaxed-simd`, widening outer-product under `+simd128`; shared i32 quant epilogue/store.
- **Lowering**: `OptMatMulPack` / `einsum_matmul` / conv `im2col` + `QSumB` generalized from `PackedFormat` to `Box<dyn MMMInputFormat>` so `PackedI8K4` routes through the einsum-matmul and conv paths. PackedFormat paths stay byte-identical (monomorphic dispatch preserved).

## Validation

- linalg kernel: **114/114** under both `+relaxed-simd` and `+simd128`
- pack: **27/27** (proptest checks `pack_view` *and* `KOut4Writer` against an independent K=4-inner reference)
- wasm (wasmtime, `+relaxed-simd`): int8 quant matmul **25/25**, int8 conv (`onnx_basic_convinteger`) bit-correct
- native tract-core regression: **244/244** (f32/f16/quant/conv paths unaffected by the `Box<dyn>` generalization)

## Test plan / follow-ups

- [ ] model-level perf number (full int8 model, `+relaxed-simd` vs `+simd128`)
- [ ] opt-in enforced 7-bit-weights mode (cargo-gated clamp + conv operand-swap) + accuracy check

### DIFF
--- a/core/src/ops/cnn/conv/conv.rs
+++ b/core/src/ops/cnn/conv/conv.rs
@@ -33,7 +33,7 @@ use crate::ops::matmul::optimized::{OptMatMul, ProtoFusedSpec};
 use crate::ops::nn::{BaseDataShape, DataFormat, DataShape};
 
 use tract_linalg::mmm::{MMMInputFormat, MatMatMul};
-use tract_linalg::pack::PackedFormat;
+use tract_linalg::pack::{PackedFormat, PackedI8K4};
 
 #[derive(Debug, Clone, new, Hash, PartialEq, Eq)]
 pub struct Conv {
@@ -123,13 +123,11 @@ impl Conv {
                 &[kernel],
             )?
         } else {
-            let format = format
-                .downcast_ref::<PackedFormat>()
-                .context("Expect regular packing for numeric weights")?;
+            // PackedFormat or a custom numeric packer (e.g. PackedI8K4).
             model.wire_node(
                 format!("{name}.prep_kernel.pack"),
                 OptMatMulPack {
-                    packers: vec![format.clone()],
+                    packers: vec![dyn_clone::clone_box(format)],
                     k_axis: 2,
                     mn_axis: 1,
                     mode_picker: ModePicker::Single,
@@ -240,7 +238,11 @@ impl Conv {
             &sum_ker_n_g_c,
         )?;
 
-        ensure!(mmm.packings()[packing].1.downcast_ref::<PackedFormat>().is_some());
+        ensure!(
+            mmm.packings()[packing].1.downcast_ref::<PackedFormat>().is_some()
+                || mmm.packings()[packing].1.downcast_ref::<PackedI8K4>().is_some(),
+            "Im2Col/QSumB support PackedFormat or PackedI8K4 activation packings"
+        );
         let mut sum_x = model.wire_node(
             format!("{name}.sum_x"),
             super::QSumB { dt: b_fact.datum_type, n, r: mmm.nr(), k },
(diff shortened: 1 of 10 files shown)


kali:
Unless I'm mistaken, this contains some relatively serious refactoring of the Im2Pack. TBH, this worries me more than the i8 new kernels. Let's see how it goes.

kali:
I enjoy this series of commits TBH, hope they work out. You're playing with extension points that I intentionally put in there, but never found time to make concrete. It feels good :)

kali:
ok, i was worried about this one because of the packing refactoring, but it looks neutral on the bench.

czoli1976:
@kali would it make sense to have an  opt-in enforced 7-bit-weights mode so one can trade precision for determinism ? I am not sure it is worth it but would be easy to do ...

PR #2287, opened 2026-05-26
onnx: add com.microsoft BiasGelu, FastGelu, QuickGelu handlers

czoli1976:
Adds com.microsoft BiasGelu/FastGelu/QuickGelu importer handlers, lowered onto existing element-wise primitives. Validated bit-close (<=4e-7) vs onnxruntime across shapes/bias/alpha; no node-suite regression; clippy+fmt clean. Part of com.microsoft contrib-op coverage for ORT-exported LLMs (complements #2283/#2284).


czoli1976:
These ONXX Compliance are all low priority, not even sure they are needed, tought would give it a try

kali (approved):

PR #2294, opened 2026-05-27
onnx,core: fuse standard ONNX LSTM cell into one LstmEpilogue op

czoli1976:
## Summary

When testing **DTLN** — an LSTM-heavy real-time speech denoiser — on tract, profiling showed that in streaming mode (batch=1, one audio frame per call) the LSTM body spends most of its **non-matmul** time on per-gate op **dispatch**, not arithmetic. tract's ONNX importer decomposes each LSTM cell into ~15 separately-evaluated `Sigmoid`/`Tanh`/`Mul`/`Add` ops, each materialising an intermediate tensor. The matmuls are already efficient (memory-bandwidth-bound), so this scaffolding is the dominant overhead for streaming LSTMs.

This PR collapses that chain into one fused `LstmEpilogue` op, emitted from the importer for the standard cell.

## What changed

- **New `LstmEpilogue` op** (`core/src/ops/lstm_cell.rs`): given the gate pre-activations `[batch, 4*hidden]` (ONNX gate order i, o, f, c) and `c_prev`, computes `Ht`/`Ct` in one pass — sigmoid over the i,o,f block + tanh over the c block via tract's vectorised `sigmoid_f32`/`tanh_f32` kernels on contiguous slices, then the cell/hidden update. Replaces the ~15-node decomposed chain.
- **Importer emission** (`onnx/src/ops/rec/lstm.rs`): emits `LstmEpilogue` for the standard cell (no peepholes, concrete hidden size); peephole / symbolic-hidden cells keep the existing decomposed path. tract's ONNX LSTM always uses sigmoid/tanh activations, so the fused op is always valid where it fires. Also computes the gate projections with one matmul per side (`Xt·Wᵀ`, `Ht-1·Rᵀ`) + slice rather than four separate matmuls.

## Correctness

Same activation kernels as the decomposed path → numerically identical. DTLN end-to-end output stays at **110.47 dB / Pearson 1.00000** vs the native reference (bit-equivalent up to f32 rounding). A unit test checks the cell math against a scalar reference over a multi-row batch.

## Perf gain

DTLN, M1 Pro, f32, single-thread, full pipeline (the production regime), restaurant-noise clip, min of 3 runs; lower = faster:

| engine | ms / 1 s audio | ×realtime |
|---|---|---|
| tract `main` (before) | 14.71 | 68× |
| **tract + this PR** | **13.11** | **76×** |
| TFLite — Ruy (TFLite default) | 13.80 | 72× |
| TFLite — XNNPACK delegate | 11.76 | 85× |

- **−11% vs tract `main`** (14.71 → 13.11).
- Now **~5% faster than TFLite's default (Ruy) engine** (13.80).
- TFLite's **XNNPACK** delegate (11.76) is still ~12% ahead; that residual is matmul memory-bandwidth plus a couple of streaming-graph follow-ups (below), **not** this epilogue.

The win concentrates in **streaming / small-batch** inference, where per-op dispatch dominates the small per-frame matmuls. It is parity-preserving for all shapes; for large offline batches the relative gain shrinks as the matmuls dominate.

## Who benefits

Any ONNX model with a standard LSTM run in streaming / small-batch mode — e.g. real-time speech enhancement (DTLN, NSNet), streaming ASR LSTM decoders (RNN-T predictors), TTS (Tacotron-2), OCR (CRNN), time-series forecasting (DeepAR).

## Scope / notes

- Measured against vanilla `main`; orthogonal to in-flight perf PRs (#2274 matmul cache-blocking is a different, compounding axis; #2257 Scan iter-reuse is neutral on seq=1 streaming).
- Follow-ups (not in this PR): a companion change inlines the single-iteration `Scan` for streaming RNNs (a further ~−3% on DTLN, → 12.68 ms/s); and a declutter pass for LSTM cells that were *unrolled* at export time (some tf2onnx outputs contain no `LSTM` op, so this importer path never sees them).

### DIFF
--- a/core/src/ops/lstm_cell.rs
+++ b/core/src/ops/lstm_cell.rs
@@ -1,0 +1,128 @@
+use crate::internal::*;
+
+/// Fused LSTM cell epilogue.
+///
+/// Given the combined gate pre-activations
+/// `preact = Xt·Wᵀ + Ht-1·Rᵀ + bias` of shape `[batch, 4*hidden]` (ONNX gate
+/// order i, o, f, c) and the previous cell state `c_prev` `[batch, hidden]`,
+/// computes the new hidden `Ht` and cell `Ct` in a SINGLE fused pass.
+///
+/// This collapses the per-gate `Sigmoid`/`Tanh` + elementwise `Mul`/`Add`
+/// chain (≈ 15 separately-dispatched ops, each materialising an intermediate
+/// tensor) into one op — the dominant non-matmul cost for streaming LSTM
+/// inference. Standard activations only (`f = sigmoid`, `g = h = tanh`) and no
+/// peepholes; the importer falls back to the decomposed form otherwise.
+///
+/// Activations use tract's vectorised `sigmoid_f32`/`tanh_f32` linalg kernels
+/// (NEON on aarch64) applied to contiguous gate slices, so the output is
+/// numerically identical to the decomposed Sigmoid/Tanh path while collapsing
+/// the per-gate dispatch into one op.
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+pub struct LstmEpilogue {
+    pub hidden: usize,
+}
+
+impl Op for LstmEpilogue {
+    fn name(&self) -> StaticName {
+        "LstmEpilogue".into()
+    }
+
+    fn info(&self) -> TractResult<Vec<String>> {
+        Ok(vec![format!("hidden={}", self.hidden)])
+    }
+
+    op_as_typed_op!();
+}
+
+impl EvalOp for LstmEpilogue {
+    fn is_stateless(&self) -> bool {
+        true
+    }
+
+    fn eval(&self, inputs: TVec<TValue>) -> TractResult<TVec<TValue>> {
+        let h = self.hidden;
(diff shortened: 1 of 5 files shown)


kali (dismissed):

kali:
@czoli1976  ok, we have a segfault here :) and we need a nnef ser/deser. Still interested in pursuing this ?

czoli1976:
Should I be alarmed or merge noise ?


Best Regards
Ckristian Zoli

Email: ***@***.***

kali (approved):

czoli1976:
@kali fixed both. The abort was `LstmEpilogue` reading f16 buffers as f32 — `FloatPrecisionTranslator` has no op allowlist, so it casts the op's inputs to f16, but `eval` was hardcoded f32; it now dispatches f32/f16 (uses the `sigmoid_f16`/`tanh_f16` kernels). Added the missing NNEF ser/deser (`tract_core_lstm_epilogue`). `test-f16` and `test-nnef-cycle` are green, and I rebased onto current main. Still keen to pursue it.

czoli1976:
should be good now .....


Best Regards
Ckristian Zoli

Email: ***@***.***

czoli1976:
@kali rerun me

Issue #2157, opened 2026-04-27
0.23.0-dev.5: PulsedModel state regression — recurrent outputs diverge from frame 1 onward (DFN3 encoder)


czoli1976:
## Summary

Hi @kali — first, thank you for the time you've already put into the conversation on #2152, especially the diagnostic guidance about cross-checking native vs WASM. That was exactly the right starting point and shaped this whole investigation. Filing this as a separate dedicated issue so it has its own thread.

We've been validating the tract upgrade path for DFN3 (DeepFilterNet 3, libDF) and found a regression that I wanted to share. The good news up front: tract 0.21.x → 0.22.1 is bit-stable on this model (SHA256-identical output, DNSMOS unchanged at 3.324), so we're committing to 0.22.1 as our ship target per your guidance about deprecating 0.21. Your WASM kernel work delivers a real ~25% speedup on our streaming workload — much appreciated.

The regression is specifically in `tract_pulse::PulsedModel`-wrapped streaming inference on `v0.23.0-dev.5`. With same model, same `SimpleState`, same deterministic synthetic inputs:

- **Frame 0 (initial state)**: ALL 7 encoder outputs SHA256-identical between `0.22.1` and `0.23.0-dev.5` ✓
- **Frame 1 (state from frame 0)**: pure-feedforward conv outputs (`e0..e3`, `c0`) still byte-identical ✓; **GRU-recurrence-dependent outputs (`emb`, `lsnr`) DIVERGE** ✗

```
                    tract 0.22.1                tract 0.23.0-dev.5         match?
─────────────────────────────────────────────────────────────────────────────────
Frame 0
  out[0] e0 conv    2822ee63c2980207            2822ee63c2980207           ✓
  out[1] e1 conv    3bae7e84d82ddf7b            3bae7e84d82ddf7b           ✓
  out[2] e2 conv    272eb7f3bd3bc635            272eb7f3bd3bc635           ✓
  out[3] e3 conv    3febc9329b58a22a            3febc9329b58a22a           ✓
  out[4] emb        871b3125c55c0852            871b3125c55c0852           ✓
  out[5] c0         a1d0ba984e3ae3ae            a1d0ba984e3ae3ae           ✓
  out[6] lsnr       f53bb3276ffa9d8d (-2.046)   f53bb3276ffa9d8d (-2.046)  ✓

Frame 1
  out[0] e0 conv    f175696e8b155b51            f175696e8b155b51           ✓
[...]

PR #2295, opened 2026-05-27
onnx: auto-enable external_state to inline single-iteration Scan for streaming RNNs

czoli1976:
## Summary

Companion to #2294, from the same DTLN profiling. tract already has a `declutter_single_loop` pass that inlines a single-iteration `Scan`, but it only fires when `Scan::external_state` is set — which until now was reachable only via the manual `force_scan_external_state` transform. So **streaming RNNs (`seq_len == 1`, one frame/token per call) carried a dead one-iteration Scan on every call**: pure orchestration overhead (state spawn, per-iter setup, output assembly).

This PR has the ONNX recurrent importer set `external_state` automatically when the caller manages state.

## What changed

In `onnx/src/ops/rec/common.rs`, when an LSTM/GRU/RNN exposes its **full** recurrent state both as input and output — `initial_h` + `Y_h` (and `initial_c` + `Y_c` for LSTM) — set `Scan::external_state = true`. That is the caller-managed-state contract (streaming denoisers, autoregressive / RNN-T decoders): the caller supplies state in and reads it out every call. The existing `declutter_single_loop` then inlines the `seq_len == 1` Scan, feeding the body's State input directly from the outer input each call.

13 lines; no new op or pass — it just enables the existing inliner for the case it was designed for.

## Correctness / soundness

`external_state` only affects `declutter_single_loop` (the inline guard); it is inert for multi-iteration Scans, so full-sequence (`seq_len > 1`) runs are unchanged. Inlining a single-iteration Scan with externally-supplied state is sound — the body runs once and state flows through the model I/O (this satisfies the guard from issue #2157).

Measured on DTLN end-to-end: output stays **110.47 dB / Pearson 1.00000** vs the native reference. Existing `tract-onnx` tests pass.

## Perf gain

DTLN, M1 Pro, f32, single-thread, full pipeline, restaurant-noise clip, min of 3 runs; lower = faster:

| config | ms / 1 s audio | ×realtime |
|---|---|---|
| tract `main` | 14.71 | 68× |
| **tract + this PR** | **13.54** | **74×** |
| tract + #2294 + this PR | 12.68 | 79× |
| TFLite — Ruy (TFLite default) | 13.80 | 72× |
| TFLite — XNNPACK delegate | 11.76 | 85× |

- **−8% vs `main`** (14.71 → 13.54), edging past TFLite's default (Ruy) engine.
- **Stacks with #2294** (the `LstmEpilogue` gate-epilogue fusion): together **−14%** (→ 12.68), comfortably past Ruy. The two are orthogonal — this PR removes Scan orchestration, #2294 fuses the gate epilogue.
- TFLite's XNNPACK delegate (11.76) is still ahead; that residual is matmul memory-bandwidth, not Scan/epilogue overhead.

## Who benefits

Every `seq_len == 1` streaming RNN exported with exposed state — real-time speech enhancement (DTLN, NSNet), streaming ASR LSTM/GRU decoders (RNN-T predictors), autoregressive TTS, and similar on-device / real-time workloads.

### DIFF
--- a/core/src/ops/scan/decluttered.rs
+++ b/core/src/ops/scan/decluttered.rs
@@ -110,18 +110,27 @@ impl Scan {
         // on the first call or when reset_every_turn is set; otherwise the
         // body input is fed from the carried hidden_state.
         //
-        // Inlining is therefore safe iff the caller does not rely on
-        // tract's across-call carry, which we encode as one of:
+        // Inlining is safe iff the caller does not rely on tract's across-call
+        // carry. We accept it when:
         //   - reset_every_turn (carry is cleared every turn anyway), or
-        //   - external_state (caller plumbs the State input every call,
-        //     e.g. parakeet decoder).
-        // Otherwise (internal-managed state, e.g. DFN3 GRU under pulse),
-        // bail. See issue #2157.
-        rule_if!(
-            self.reset_every_turn
-                || self.external_state
-                || !self.input_mapping.iter().any(InputMapping::is_state)
-        );
+        //   - external_state (explicitly asserted, e.g. force_scan_external_state), or
+        //   - there is no State at all, or
+        //   - the caller manages the state: every recurrent state has a
+        //     last-value output that reaches a model output, so the caller can
+        //     read the updated state and feed it back across calls (e.g. DTLN /
+        //     parakeet decoders). A pulse model with tract-managed state (e.g.
+        //     DFN3 GRU) does NOT export its state, so it is not inlined.
+        // See issue #2157.
+        let has_state = self.input_mapping.iter().any(InputMapping::is_state);
+        let state_outputs: Vec<_> = self.output_mapping.iter().filter(|m| m.state).collect();
+        let state_exported = has_state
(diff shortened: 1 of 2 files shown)


kali:
So i have tried this before, but ran into a problem. It don't think it is possible to distinguish reliably a stateful model meant to be called with pulses of length 1 from a "stateless" (from tract POV) model with state managed on the outside. The difference is only manifest once the user call it with Runnable::run or ::spawn then State::run, but we need to apply the Transform before that. (Now that I think of it, this is how 0.23 where breaking DFN3, right ?)

WDYT ?

kali:
Ha, your heuristics is, C and H are outputted so the state is managed externally ?

czoli1976:
You are right, it has an hole now, as it is

image

Trying to refine to see if can make it more robust, let's see

kali:
Should we close this one for the time being ?

czoli1976:
I think the second /last commit made it more robust

kali (approved):

kali:
ok, let's try it.

PR #2297, opened 2026-05-27
tflite: load DTLN models — op coverage + REDUCE_MEAN float-mean fix

czoli1976:
## Summary

Enables `tract-tflite` to load and run  **DTLN** real-time speech-denoiser `.tflite` models. Before this, the loader bailed on the very first op; now both DTLN sub-models load, run, and are **bit-exact vs the ONNX path** and match the native reference at **110.47 dB / Pearson 1.00000** end-to-end.

While getting there I found a real correctness bug (below) that affects far more than DTLN.

## The headline fix — REDUCE_MEAN was rounding to an integer

`de_reduce_mean` computed the mean as `sum * recip(count)` using `core::quant::scale()` — which **rounds its result to the nearest integer** (`round_ties_to_even`, for requantization). So any non-integer mean was silently corrupted: e.g. a layer-norm mean of `0.0284` became `0`, a variance of `13.4` became `13.0`. The fix uses a plain float multiply (`core::math::mul`). This affects any tflite model using `MEAN` / global-average-pool / layer-norm, not just DTLN.

## What changed (all in `tflite/`)

- **Op coverage** (`array.rs`, `math.rs`): add `UNPACK`, `PACK`, `SPLIT`, `SQUARED_DIFFERENCE` handlers.
- **`STRIDED_SLICE` `shrink_axis_mask`** (`array.rs`): the guard rejected `shrink_axis_mask != 0`, but tract's `StridedSlice` already honours it (it's passed straight through) — relaxed to reject only the genuinely-unsupported `new_axis_mask`.
- **`REDUCE_MEAN` correctness** (`nn.rs`): float-mean fix (above) + negative-axis normalization in `de_reduce` / `de_reduce_mean` (a `-1` axis was cast to a huge `usize` and indexed out of bounds).
- **`FULLY_CONNECTED` optional bias** (`nn.rs`): `args_3!` required exactly 3 inputs, but the body already handled the no-bias case by input count; accept 2-or-3.
- **Optional input `-1` sentinel** (`model.rs`, `registry.rs`): tflite encodes an absent optional input as `-1`; skip it in fact-building and filter it from an op's input list (previously `tensors().get((-1) as usize)` panicked).
- **Empty quantization guard** (`tensors.rs`): float tensors often carry an empty `quantization` struct; only treat as quantized when `scale`/`zero_point` actually have entries (previously `scale.get(0)` panicked out of bounds).

These are additive (new op handlers) or strictly-corrective (the guards/bugfixes), so currently-loading models are unaffected — the only behavioural change to an existing op is `REDUCE_MEAN` no longer rounding, which can only make results *more* correct.

## Validation

Per-model, identical random input fed to both frontends:
- `model_1.tflite` vs `model_1.onnx`: **max|diff| = 0** on both outputs.
- `model_2.tflite` vs `model_2.onnx`: **max|diff| = 0** on both outputs.

End-to-end (full DTLN pipeline, real audio) via `tract-tflite` vs the native DataDog output: **Pearson 1.00000, diff SNR 110.47 dB, max abs diff 1e-5**.

## Notes

- `tract-tflite` currently has no in-crate test suite, and TFLite's existing coverage is real-model harnesses (`harness/tfl-mobilenet-v2-q` downloads a model). Happy to add a `harness/dtln-tflite` along those lines if you can point me at where the DTLN `.tflite` should be hosted for CI — or to add round-trip unit tests if you'd prefer that style.
- This bundles several independent fixes for one coherent goal; glad to split (e.g. the `REDUCE_MEAN` correctness fix on its own) if you'd rather review them separately.

### DIFF
--- a/tflite/src/model.rs
+++ b/tflite/src/model.rs
@@ -94,6 +94,9 @@ impl Framework<TfliteProtoModel, TypedModel> for Tflite {
         }
         for op in main.operators().context("No operators in Tflite model")? {
             for input in op.inputs().context("No input in Tflite  operator")? {
+                if input == -1 {
+                    continue; // tflite uses -1 as the "no tensor" sentinel for absent optional inputs
+                }
                 if let Entry::Vacant(slot) = mapping.entry(input) {
                     let (fact, name) = flat_tensor_to_tract_fact(&root, main, input)?;
                     let value = fact.konst.with_context(|| format!("Error in TF file for operator {op:?}. No prior computation nor constant for input {input}"))?;
--- a/tflite/src/ops/array.rs
+++ b/tflite/src/ops/array.rs
@@ -26,14 +26,17 @@ pub fn register_all(reg: &mut Registry) {
     reg.reg_to_tract(BuiltinOperator::BROADCAST_TO, de_broadcast_to);
     reg.reg_to_tract(BuiltinOperator::CONCATENATION, de_concat);
     reg.reg_to_tract(BuiltinOperator::EXPAND_DIMS, de_expand_dims);
+    reg.reg_to_tract(BuiltinOperator::PACK, de_pack);
     reg.reg_to_tract(BuiltinOperator::PAD, de_pad);
     reg.reg_to_tract(BuiltinOperator::PADV2, de_padv2);
     reg.reg_to_tract(BuiltinOperator::RESHAPE, de_reshape);
     reg.reg_to_tract(BuiltinOperator::SHAPE, de_shape);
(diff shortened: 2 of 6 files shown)


kali:
Mmmm... you want to go there ? :) Because I feel like tflite is on its way out. I was letting tflite and tf die. You think we should keep tfilet ?

czoli1976:
you have a point .... happy to close or stop here, PLMK

kali:
yeah. energy and bandwidth is better used elsewhere i think.

PR #2311, opened 2026-05-28
linalg/x86_64 + core/nn: fused AVX-512 RmsNorm kernel

czoli1976:
## Summary

Adds a fused AVX-512 RmsNorm kernel + scalar fallback at the linalg level, and routes `core::ops::nn::RmsNorm::eval`'s common case (contiguous trailing axis, F32 or F16) through it instead of the existing 4-call composition.

**Why**: `RmsNorm::eval` currently runs `Reducer::MeanOfSquares` + `Add` + `Rsqrt` + `Mul` as 4 separate ops, each writing/reading the full input through L1/L2. CUDA and Metal already expose a fused `rms_norm` kernel (`cuda/src/kernels/nn/rms_norm.rs`, `metal/src/kernels/nn/rms_norm.rs`); this closes the CPU side of that gap.

**What**:
- New `Ops::rms_norm_f32: Box<dyn Fn(&mut [f32], f32) + Send + Sync>` — a simple closure slot since the op operates on a whole row (doesn't fit the per-element `ElementWise` pattern).
- `linalg/src/generic/rms_norm.rs` — scalar default.
- `linalg/src/x86_64_fma/rms_norm.rs` — AVX-512 kernel. Two passes: (1) sum-of-squares via 4 zmm FMA accumulators + zmm→scalar reduce + rsqrt, (2) multiply-back via 4 zmm broadcast-multiplies. Scalar tail handles `row_len % 64 ≠ 0`. Uses `vmovups` (per-row slices aren't guaranteed 64-byte aligned).
- `plug_avx512f` overrides the closure on AVX-512 hosts.
- `core::ops::nn::RmsNorm::eval` adds a fast path: when `dtype ∈ {F32, F16}` and `self.axis == input.rank() - 1`, iterates over outer dims and dispatches to the linalg primitive. Other axes keep the original composition.

Bench (single-thread, Cascade Lake, throughput Gelem/s):
| row | composed | generic scalar | AVX-512 | AVX-512 vs composed |
|----:|---------:|---------------:|--------:|--------------------:|
| 1024 |   0.77  |     0.75       |  12.4   |   **16.2×**         |
| 2048 |   0.77  |     0.75       |  13.8   |   **17.9×**         |
| 4096 |   0.77  |     0.75       |  13.8   |   **17.9×**         |

The composed and generic-scalar numbers are nearly identical because both are memory-bandwidth bound on a 4-pass / 2-pass workload of the same total work; the AVX-512 win comes from doing both passes in 1/4 the loop iterations (zmm width) and from avoiding the inter-op allocation + dispatch overhead in the composed version.

## Test plan
- [x] `cargo test --release -p tract-linalg --lib rms_norm` — 8 tests (4 generic, 4 AVX-512 against the scalar reference, including a length-with-tail case).
- [x] `cargo test --release -p tract-core --lib rms_norm` — `eval_with_f16_eps_and_f16_input` (now exercises the new fast path: rank-1 F16 input + F16 eps) plus new `eval_with_non_trailing_axis_f32` (rank-2, axis=0) which asserts the slow-path 4-call composition still matches a hand-computed reference within 1e-5.
- [x] `cargo test --release -p tract-linalg` — 2665 passed, 0 failed.
- [x] `cargo test --release -p tract-core --lib` — 246 passed, 0 failed.
- [x] `cargo bench --bench rms_norm` — numbers above.
- [x] Non-AVX512 x86 hosts unchanged (scalar fallback exercised via `plug_avx512f` gating).
- [x] Non-trailing-axis case unchanged (slow path = original 4-call composition; defended by the new test above).
- [x] Cross-arch builds clean (`aarch64-unknown-linux-gnu`, `wasm32-unknown-unknown`): the AVX-512 asm in `linalg/src/x86_64_fma/rms_norm.rs` is walled off by `#[cfg(target_arch = "x86_64")]` on the `x86_64_fma` module in `linalg/src/lib.rs`.

## Validation environment

x86_64 KVM guest, Ubuntu 24.04.4 LTS (kernel 6.18.5), rustc 1.94.1.

- CPU: Intel Xeon @ 2.80 GHz, family 6 / model 85 / stepping 7 (Cascade Lake-SP); 4 vCPU, 1 thread/core, 1 socket.
- AVX-512 features (all confirmed by `is_x86_feature_detected!`): f, vnni, dq, bw, vl, cd.
- Cache: L1d 32 KiB/core, L2 1 MiB/core, L3 33 MiB shared. 15 GiB RAM.
- Bench discipline: `RAYON_NUM_THREADS=1`, `taskset -c 0`, Criterion warm-up 5 s / measure 15 s, sample size 100.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

---
_Generated by [Claude Code](https://claude.ai/code/session_01MtQweWhf8E1W7pEJMF6ywf)_

### DIFF
--- a/core/src/ops/nn/rms_norm.rs
+++ b/core/src/ops/nn/rms_norm.rs
@@ -28,7 +28,34 @@ impl EvalOp for RmsNorm {
 
     fn eval(&self, inputs: TVec<TValue>) -> TractResult<TVec<TValue>> {
         let input = args_1!(inputs);
+        let in_dt = input.datum_type();
 
+        // Fast path: F32 or F16 input where the normalised axis is the last
+        // (contiguous) one. Use the fused tract_linalg::rms_norm_f32 kernel
+        // (AVX-512 when available; scalar fallback otherwise) instead of the
+        // 4-call MeanOfSquares + Add + Rsqrt + Mul composition below. ~16-18x
+        // faster on Cascade Lake AVX-512, ~equivalent on the scalar fallback
+        // since the composition is also memory-bandwidth bound.
+        if matches!(in_dt, DatumType::F32 | DatumType::F16)
+            && input.rank() > 0
+            && self.axis == input.rank() - 1
+        {
+            let eps_f32: f32 = self.eps.cast_to_scalar::<f32>()?;
+            let mut buf = input.cast_to::<f32>()?.into_owned();
+            let row_len = buf.shape()[self.axis];
+            if row_len > 0 {
+                let n_rows: usize = buf.shape().iter().take(self.axis).product();
+                let data = unsafe { buf.as_slice_mut_unchecked::<f32>() };
+                let rms_norm = &tract_linalg::ops().rms_norm_f32;
+                for r in 0..n_rows {
+                    let start = r * row_len;
+                    rms_norm(&mut data[start..start + row_len], eps_f32);
+                }
+            }
+            return Ok(tvec![buf.cast_to_dt(in_dt)?.into_owned().into()]);
+        }
+
+        // Slow path: original 4-call composition (kept for non-contiguous axes).
         let input_f32 = input.cast_to::<f32>()?.into_owned();
         // eps inherits the input dtype from the declutter pattern (F16 when the
         // surrounding LayerNorm chain is F16). The MeanOfSquares + Add + Rsqrt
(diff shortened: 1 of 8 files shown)


kali (dismissed):

kali:
rebased!

czoli1976:
image

kali:
dafuk is this bird ? :) plus, i hope there is never a 0.24, i think it's time to start adulting.

czoli1976:
It's a peacock, beautiful and elegant, admittedly not considered a very smart bird

kali:
I know, I'm a country guy. And yeah, they're pretty dumb.

PR #2319, opened 2026-05-30
transformers: CPU FlashSdpa — contiguous P·V GEMM + head-parallel exec + seq-len lowering heuristic

czoli1976:
## What

Three related changes to CPU scaled-dot-product attention:

1. **Contiguous P·V GEMM.** `FlashSdpaOp` computed the `P·V` step as `head_dim` separate dot products against **strided** columns of V (`s.row(i).dot(&vblock.column(j))`), which is neither a GEMM nor vectorization-friendly. It is now a single contiguous tile GEMM (`s.dot(&vblock)`) plus the online-softmax rescale. **Bit-exact** (`max_abs = 0` vs a naive `softmax(QKᵀ·scale)·V` reference).

2. **Head-parallel execution.** The independent `(batch, q-head)` tasks now run across cores on rayon's global pool — heads share only read-only Q/K/V and write disjoint output slices. Disable with `TRACT_FLASH_SDPA_ST=1`; single-threaded on wasm (no thread spawn).

3. **Sequence-length lowering heuristic.** `Sdpa::codegen` now consults `flash_min_seq_len()`: an f32 SDPA whose K/V length is below `TRACT_FLASH_SDPA_MIN_SEQ_LEN` lowers to the decomposed matmul+softmax path instead of `FlashSdpaOp`. Default `0` keeps flash for every length; the knob is there for hosts where short-sequence decompose wins.

## Measurements (Apple M1 Pro, 6 performance + 2 efficiency cores)

**The op now scales ~5× across the performance cores** (Hq32/Hkv8/S512/D64):

| threads | 1 | 2 | 4 | 6 | 8 |
|---|---|---|---|---|---|
| speedup | 1.0× | 1.8× | 3.5× | 4.7× | 5.0× |

Near-linear to 4 (89% efficiency), plateauing once the efficiency cores come in — i.e. compute-bound, not memory-bound, in that range.

**With head-parallelism, flash beats the decomposed path at every sequence length** (it loses 2.5–3.6× single-threaded, which is what motivated the default-flash heuristic):

| S | flash (head-parallel) | decomposed |
|---|---|---|
| 128 | 1.9 ms | 3.0 ms |
| 512 | 22.9 ms | 32.7 ms |
| 2048 | 340 ms | 489 ms |
| 4096 | 1372 ms | 3416 ms |

## Honest scope (replaces the earlier "~3.1× / PP512" framing)

The original "~3.1×" was an isolated-op microbench, not an end-to-end number. Running the actual `bundle-entrypoint.sh` benches on CPU:

- The **LLM PP512 models** (`Llama-3.2-1B-q40ef32`, `OpenELM-270M-q40ef16`) **decompose** SDPA to `MatMul` + `ScaledMaskedSoftmax` (their acc dtype isn't f32) — the optimized graph has **0 `FlashSDPA` nodes**, so this PR doesn't touch them. PP512 is unchanged (~40 tok/s, no regression).
- The one bench model that uses `FlashSdpaOp` is the **`parakeet-…-f32f32` encoder** (24 nodes). Head-parallelism there is ~0% at the benched `S=100` and ~3% at `S=512`, because attention is only ~7% of the encoder.
- The dominant CPU cost is the matmuls (conv / FFN / projections). Threading **those** (`multithread-mm`, currently off by default) gives **1.64×** on the parakeet encoder vs ~1.05× from threading attention — a bigger, orthogonal lever.

So this is a correctness-neutral op-level improvement that pays off for **attention-dominated / longer-sequence f32** workloads; it is not a PP512 win. Happy to follow up on the seq-len threshold tuning and the broader "thread the matmuls" direction.

## Tests

- `flash_sdpa_pv_matches_naive` — bit-exact vs naive reference (non-square, multi-head; exercises the parallel path).
- `bench_flash_sdpa_pv` (`#[ignore]`) — the strided-vs-GEMM A/B harness.
- Full `tract-transformers` suite green; `fmt` + `clippy` clean; `wasm32-unknown-unknown` check passes (rayon is cfg-gated off there).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5052,6 +5052,7 @@ name = "tract-transformers"
 version = "0.23.0-pre"
 dependencies = [
  "float-ord",
+ "rayon",
  "tract-nnef",
 ]
 
--- a/transformers/Cargo.toml
+++ b/transformers/Cargo.toml
@@ -17,3 +17,9 @@ maintenance = { status = "actively-developed" }
 [dependencies]
 float-ord.workspace = true
 tract-nnef.workspace = true
+
+# Used to spread FlashSDPA's independent attention heads across cores (rayon's
+# global pool). Not available on wasm32-unknown-unknown (no thread spawn), where
+# the op stays single-threaded.
+[target.'cfg(not(target_family = "wasm"))'.dependencies]
+rayon.workspace = true
--- a/transformers/src/ops/flash_sdpa.rs
+++ b/transformers/src/ops/flash_sdpa.rs
@@ -174,108 +174,292 @@ impl FlashSdpaOp {
         v: ArrayView4<f32>,
         mask: Option<ArrayView4<f32>>,
     ) -> Array4<f32> {
-        // Explicit dimensions
+        self.flash_attention_gqa_impl(q, k, v, mask, false)
+    }
+
+    fn flash_attention_gqa_impl(
+        &self,
+        q: ArrayView4<f32>,
+        k: ArrayView4<f32>,
+        v: ArrayView4<f32>,
+        mask: Option<ArrayView4<f32>>,
+        pv_strided: bool,
+    ) -> Array4<f32> {
         let (batch_size, num_q_heads, q_len, head_dim) = q.dim();
-        let (_, num_kv_heads, kv_len, _) = k.dim();
-        let scale = self.scale.unwrap_or((head_dim as f32).recip().sqrt());
+        let (_, num_kv_heads, _, _) = k.dim();
         let group_size = num_q_heads / num_kv_heads;
+
+        let mut out = Array4::<f32>::zeros((batch_size, num_q_heads, q_len, head_dim));
+
+        // One independent task per (batch, q-head). Heads share only read-only Q/K/V
+        // and write disjoint output slices, so this is embarrassingly parallel.
+        let tasks: Vec<(usize, usize)> =
+            (0..batch_size).flat_map(|b| (0..num_q_heads).map(move |qh| (b, qh))).collect();
+        let compute = |&(b, qh): &(usize, usize)| {
+            self.attend_one_head(b, qh, qh / group_size, q, k, v, mask.as_ref(), pv_strided)
+        };
+
+        let results: Vec<Array2<f32>> = if Self::should_thread(tasks.len()) {
+            #[cfg(not(target_family = "wasm"))]
+            {
+                use rayon::prelude::*;
+                tasks.par_iter().map(compute).collect()
+            }
+            #[cfg(target_family = "wasm")]
+            {
+                tasks.iter().map(compute).collect()
+            }
+        } else {
+            tasks.iter().map(compute).collect()
+        };
+
+        for (&(b, qh), head_out) in tasks.iter().zip(results) {
+            out.slice_mut(s!(b, qh, .., ..)).assign(&head_out);
+        }
+        out
+    }
+
+    /// Whether to spread the per-head tasks across cores (rayon's global pool).
+    /// Off for a single task, on wasm (no thread spawn), or when `TRACT_FLASH_SDPA_ST` is set.
+    fn should_thread(num_tasks: usize) -> bool {
+        num_tasks > 1
+            && cfg!(not(target_family = "wasm"))
+            && std::env::var_os("TRACT_FLASH_SDPA_ST").is_none()
+    }
+
+    /// Flash-attention for a single (batch, q-head) pair. Returns O of shape
+    /// `[q_len, head_dim]`. Pure function of the read-only Q/K/V/mask views, so it
+    /// is safe to call concurrently for distinct heads.
+    #[allow(clippy::too_many_arguments)]
+    fn attend_one_head(
+        &self,
+        b: usize,
+        qh: usize,
+        kvh: usize,
+        q: ArrayView4<f32>,
+        k: ArrayView4<f32>,
+        v: ArrayView4<f32>,
+        mask: Option<&ArrayView4<f32>>,
+        pv_strided: bool,
+    ) -> Array2<f32> {
+        let (_, _, q_len, head_dim) = q.dim();
+        let kv_len = k.dim().2;
+        let scale = self.scale.unwrap_or((head_dim as f32).recip().sqrt());
 
         let block_kv_len = 32;
         let block_q_len = 32;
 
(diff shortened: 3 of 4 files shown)


czoli1976:
@kali  this one looks promising

kali (dismissed):

kali:
Sending this one to the bench.

czoli1976:
safe journey

kali:
Bench is underwhelming. We were expecting this to boost PP512 on cpu, right ? did you observed it with the example in the bench ? (as a reminder, the benches are in .travis/bundle-entrypoint.sh, including the llm ones)

czoli1976:
image

kali:
ok, but these numbers are not model e2e right ? just measuring FlashSDPA in isolation ? is it why i don't see anything on the bench ?

czoli1976:
Correct, not true E2E, went by Proxy ....

The code is actually 3x faster, but the CPU Flash Attention is still 3x Slower than the Tract Optimised Decomposed one, which makes sense, but Flash Attention is used to keep memory usage low, no?

So I am confused now, is this CPU Flash Attention Path never actually used ?

Running your Bench now but now really not sure if this Path is used and when ...

czoli1976:
ok, bench run and ... drummrolls ....

You're right it's underwhelming on the bench, and digging in explains why

I built the CLI at this PR's commit and its parent (diff is only flash_sdpa.rs) and ran the bundle-entrypoint.sh invocations on CPU.

The PP512 LLM models don't use FlashSdpaOp at all !
 On Llama-3.2-1B-q40ef32 and OpenELM-270M-q40ef16, the optimized graph (--llm -O dump) has 0 FlashSDPA nodes — attention lowers to MatMul + ScaledMaskedSoftmax via Sdpa::codegen's decomposition (their acc dtype isn't f32). So PP512 is unchanged — ~40.1 vs ~41.0 tok/s PR vs base, i.e. within noise. The P·V change never executes.
 
 So it is not a surprise PP512 did not move.

The one bench model that does use it is the parakeet-...-f32f32 encoder because it actually has  FlashSDPA (24 FlashSDPA nodes; f32 acc → codegen picks the flash op). 

Running the bench is with  S=100: neutral (0.242 vs 0.236 s — attention's too short to matter) but once I run it for longer periods like with  S=512 I actually got a consistent  gain : about  14% encoder speed -> 0.338 vs 0.384 s, every round
So the P·V GEMM is a real E2E win for attention-heavy f32 models at realistic sequence lengths — the current bench points just don't exercise that.

The bigger thing I noticed: on CPU, FlashSdpaOp is ~3× slower than the decomposed patch_sdpa path it's chosen over. Building one Sdpa{f32} and running it both ways in-tract (codegen→FlashSdpaOp vs patch_sdpa→EinSum→linalg):

image


The flash path is single-threaded ndarray on 32×32 tiles (off linalg); the decomposed path rides tract's threaded matmul + softmax. So the f32→FlashSdpaOp routing is effectively backwards perf-wise — the f32 model (parakeet) gets the slow flash path while the f16 LLMs get the fast decomposed one. Flash's real advantage is memory (it never materializes the full S×S scores), which only matters at long context; decomposed wins on speed at least through S=2048.

So, two possible options here seem to be

- Have CPU f32 SDPA decompose like the f16 path (perhaps gating FlashSdpaOp to very long S where the memory saving matters) — that'd be ~3× on parakee vs just  14%.
-  keep FlashSdpaOp on CPU, parallelise it across attention heads  (hence across CPU cores too)— it's single-threaded today; estimate a  ~3.5–5× there.


Am I making any sense?

kali:
You're making sense. tract state of affairs is this area does not :) it boils down to one thing, LLM on cpu was never high priority. The consequence is what you observed: unfinished business, inconsistent lowering. Looking at it from the Parakeet angle actually makes it somewhat more interesting, because it's typically a smaller model than LLM, so running on big CPU makes more sense.

I think all options also make sense. consistent "atomic" lowering to leverage optimized multipliers, Flash for long sequences, multi-threading. My main concern here is again picking the right option at runtime (and we may have the same question with the bf16 kernels). I don't think tract has a clean story for "runtime tweaks". We have the MT setup on one side, we have "fast compact" exp on another, we also have some env vars to change behaviours in im2col, right ? All implemented in very different ways. This topic goes beyond the scope of this PR, but I think we need to think about how to tune lowering (is it enough to do it globally ? do we want locality ? some kind of annotation in NNEF maybe ?)

So my idea, short term for this PR: find heuristic, and use env vars for thresholds. S > some tweabel => go flash. About  multi-head/multi-thread, i'm curious to see if it helps, as i suspect we will be memory bound, but MMU are so complicated to predict... maybe on AMD and Apple it will be good ?

czoli1976:
I am happy to give it a try for the fun of it but is it realistic to run an LLM and/or Parakeet entirely on CPU?

As for the tuning that is quite a topic TBH, I had lots of fun benching different paths when looking at where to dispatch (e.g. SME vs AMX) but clearly that is often machine dependent (as M1 not the same of M4).

I have attempted to do an Auto-Tuner so run into the lack of Knobs but also found that sometimes my Auto-Tuner was dumb as testing needs to be E2E.

Even made a TuningSettings File that gets produced by the Auto-Tuner and can be deployed to the device so TRACT can load that (or use the defaults if that not present).

kali:
I think nemotron could translate live on a laptop without gpu, but it's attention is so broken that it will not use the sdpa. Could parakeet be useful for short content ? I'm not sure...

czoli1976:
I guess we will find out only by testing it

czoli1976:
So multi-head MT is scaling on FlashSdpa works  (so compute-bound).

S=512image

The threadable part (attention) is only ≈ 0.022 s of 0.342 s ≈ 7% 

In other words, the model barely scales due to the small attention share: ~5% total, saturating by 4 threads. This is Amdahl in the raw — the op scales ~5× internally (107→21.6 ms), but the encoder only drops 0.342→0.324 s.
image

It's compute-bound on the P-cores, not memory-bound (clean near-linear region)

czoli1976:
Parakeet E2E with everything threaded (matmuls via --threads N, attention via RAYON_NUM_THREADS=N), N=1→8, min-of-3, at S=512

image

image

czoli1976:
@kali PR squash updated, PR description as well, all-in including a K/V length threshold, unsurprisingly named TRACT_FLASH_SDPA_MIN_SEQ_LEN

kali (approved):

kali:
I want to land this one quickly too, running a bench just to make sure.

czoli1976:
image

Issue #2323, opened 2026-06-01
Sliding-Window Attention


czoli1976:
Hi @kali , I have noticed that there is now no support for it.

WDYT? Would it be a nice addition ?

kali:
Mmm...  I jsut did some form of it for nemotron, we could do more, which variant do you have in mind ?

czoli1976:
LLM decode case — causal sliding-window like Mistral

kali:
ok, how do the model encode this in the wild ? onnx with an explicit mask ? It will be interesting to see if the recent work i did for nemotron generalizes at all :)

czoli1976:
Dug into it — neither common path uses an explicit mask, which surprised me:

ORT-GenAI / Olive / AMD on-device exports: the window is a local_window_size=4096 attribute on the fused GroupQueryAttention op (checked a real Mistral-7B-v0.1 export — all 32 layers, no Trilu/banded mask, just a 2D padding mask).
optimum / transformers.js decomposed exports: the window is dropped — I exported a tiny Mistral with sliding_window=64 and it comes out as plain full-causal attention; the window value isn't in the graph at all.
So there's no banded mask to reuse the nemotron machinery on — in the fused path it's an op attribute (which tract's GroupQueryAttention currently rejects), and in the decomposed path it's gone. 

so ...: step 1 is accepting local_window_size on GQA, step 2 is the bounded ring-buffer cache (the real win, encoding-independent). 

Happy to take the decode-cache piece — it builds on the in-place cache in #2321.

czoli1976:
@kali see https://github.com/sonos/tract/pull/2327 about this

PR #2327, opened 2026-06-01
onnx,transformers: sliding-window attention — GQA window + bounded ring-buffer decode (#2323)

czoli1976:
Follow-up to #2323 (@kali) — sliding-window attention for the **LLM decode** case (Mistral pure-SWA, Gemma-style interleaved local/global). As discussed there, this is a different flavour from the nemotron streaming-encoder one: the win is on the **KV cache**, not the mask.

## How SWA actually reaches tract (verified on real exports)
I checked rather than assumed:
- **ORT-GenAI / Olive / on-device exports** carry the window as a **`local_window_size` attribute on the fused `GroupQueryAttention` op** — inspected a real Mistral-7B-v0.1 export: all 32 layers `local_window_size=4096`, no Trilu/banded mask, only a 2D padding mask.
- **optimum / transformers.js decomposed exports** *drop* the window entirely (exported tiny Mistral with `sliding_window=64` → plain full-causal, the `64` isn't in the graph).
- An **explicit banded mask** shows up in **neither** common path.

So the window lives, in practice, as an **op attribute** — which tract's `GroupQueryAttention` was rejecting outright (so those models failed to import).

## What's in the PR

**Prefill (correctness) —** `GroupQueryAttention` now **accepts `local_window_size`** and applies it as a banded causal mask in the existing concrete-shape mask path: query `i` attends to key `j` iff `j <= i` *and* `i - j < window`. `window=0` stays plain causal; symbolic seq lengths bail (a static band can't be built). Models that used to error on import now load.

**Decode (efficiency) —** a bounded sliding-window KV cache + a fused op:
- **`WindowKvCache`** — a fixed-capacity ring buffer (overwrite oldest). Decode runs at **constant memory and per-step cost** regardless of context length, **losslessly** (the model is trained to attend only within the window). It's cheap because **decode attention is order-invariant over keys** (`O = Σ softmax_j·V_j` is unchanged under a (K,V) permutation), so the ring buffer never needs un-rotating.
- **`WindowKvSdpa`** — a stateful fused op that owns the K/V ring buffers and attends `Q` over the (≤window) cache. The bounded cache *is* the window, so attending over it equals windowed attention.
- **`WindowKvSdpaTransform { window }`** — auto-wires it: strips the GQA broadcast chain (reusing `fuse_kv_cache_broadcast_rule`) then fuses `{kv_cache(K), kv_cache(V), Sdpa}` → `WindowKvSdpa{window}`, the window threaded via the `Rewriter` context.
- **NNEF ser/de** for `WindowKvSdpa`.

## Design note (where the window comes from)
The decode graph itself doesn't carry the window (GQA is prefill-only; decode goes through the `DynKeyValueCache` fold), so the transform takes the `window` as a parameter — supplied from the model config / the GQA `local_window_size` above. Fully-automatic detection (threading the prefill attribute through to the decode cache) is the natural follow-up; this PR makes the wiring available and correct.

## Validation
- `WindowKvCache`: holds the last-W as a set (incl. prefill chunk > window), windowed attention == ordered-last-W (close, float summation order), memory bounded at W.
- `WindowKvSdpa`: builds a real `TypedModel`, runs it through the engine (`into_runnable`/`spawn`/`run`) over 15 decode steps with window=5, matches full attention over the last-W each step, cache bounded to W throughout.
- `WindowKvSdpaTransform`: fuses the decode subgraph (caches + Sdpa removed) and the rewritten model does correct windowed decode.
- GQA: banded-mask unit test; NNEF round-trip.
- Full workspace build + blast-radius/proptest suite green; fmt + clippy clean.

## Apple research & prior art
- Motivation: Apple Foundation Models 2025 use **interleaved local/global (sliding-window) attention** to bound the KV cache with no quality loss; same pattern in Mistral / Gemma / Phi-3.
- Bounded recent-window cache: StreamingLLM (`attention_sinks`), Mistral's rolling-buffer cache.

### DIFF
--- a/onnx/src/ops/nn/group_query_attention.rs
+++ b/onnx/src/ops/nn/group_query_attention.rs
@@ -23,11 +23,8 @@ pub fn group_query_attention(
         "GroupQueryAttention: internal rotary (do_rotary) is unsupported; apply RotaryEmbedding separately"
     );
     // Sliding-window attention: a query attends only to the last `window` keys (in
-    // addition to causal). <0 means no window (full causal). Applied as a banded mask.
-    let window = match node.get_attr_opt::<i64>("local_window_size")?.unwrap_or(-1) {
-        w if w < 0 => 0usize,
-        w => w as usize,
-    };
+    // addition to causal). <0 / absent means no window (full causal). Applied as a band.
+    let window = node.get_attr_opt::<i64>("local_window_size")?.unwrap_or(0).max(0) as usize;
     ensure!(
         node.get_attr_opt::<f32>("softcap")?.unwrap_or(0.0) == 0.0,
         "GroupQueryAttention: softcap is unsupported"
--- a/onnx/src/ops/nn/group_query_attention.rs
+++ b/onnx/src/ops/nn/group_query_attention.rs
@@ -9,7 +9,8 @@ use tract_transformers::ops::sdpa::Sdpa;
 //   outputs: output(0), present_key(1), present_value(2)
 // Scoped to the prefill case (no past KV cache): query/key/value are [B, S, heads*head_size],
 // attention is causal with q_seq == kv_seq, and present_key/value are the reshaped K/V.
-// Decode-step KV cache, internal rotary (do_rotary) and local-window attention are rejected.
+// Sliding-window attention (local_window_size) is applied as a banded causal mask.
+// Decode-step KV cache and internal rotary (do_rotary) are still rejected.
 pub fn group_query_attention(
     _ctx: &ParsingContext,
     node: &NodeProto,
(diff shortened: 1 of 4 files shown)


onnx/src/ops/nn/group_query_attention.rs
```diff
@@ -21,10 +22,12 @@ pub fn group_query_attention(
...
-    ensure!(
-        node.get_attr_opt::<i64>("local_window_size")?.unwrap_or(-1) < 0,
-        "GroupQueryAttention: local_window_size is unsupported"
-    );
+    // Sliding-window attention: a query attends only to the last `window` keys (in
+    // addition to causal). <0 means no window (full causal). Applied as a banded mask.
+    let window = match node.get_attr_opt::<i64>("local_window_size")?.unwrap_or(-1) {
+        w if w < 0 => 0usize,
+        w => w as usize,
+    };
```
kali: convoluted way to write .max(0) ?

kali (dismissed):

onnx/src/ops/nn/group_query_attention.rs
czoli1976: better ?

let window = node.get_attr_opt::<i64>("local_window_size")?.unwrap_or(0).max(0) as usize;

kali: I'd like it better, yes :)

czoli1976:
Thanks @kali — collapsed it to `.unwrap_or(0).max(0)` as suggested. Heads up: the push auto-dismissed your approval (dismiss-stale-on-push), so it's back in your court — good to merge whenever you have a sec.

kali (approved):

czoli1976:
So WDYT about this, does it make sense for TRACT ?

PR #2338, opened 2026-06-03
docs: add CLAUDE.md with contributor rules

kali:
Curated behavioral subset of AGENTS.md (fmt, commit/comment style, model-edit tooling, public API, test placement, PR etiquette) inlined so it lands in the auto-loaded context; AGENTS.md stays the full reference.


czoli1976:
great!

kali:
this is broken on so many levels. claude code only read deterministically claude.md. and ignores AGENTS.md more or less completely. and the runner will only present it to the LLM when it is at the top of the ide/pwd in terminal...

czoli1976:
it's a whole new world with new problems

Issue #2341, opened 2026-06-04
MatMulNBits / Q4_0: int4-compute CPU kernel — W4A8 (int8-dot) or W4A16 (f16-unpack)?


czoli1976:
Follow-up to #2340 (int4 `MatMulNBits` → fused `Q4_0` block-quant matmul) — a direction
question before I build the next piece. @kali

## Where #2340 leaves things
#2340 keeps the weight as a `Q4_0` constant instead of expanding it to f32 at load. What each
backend then does differs:
- **Metal**: the `Q4_0` weight goes straight to the existing ggml `mul_mv_q4_0` kernel — true
  int4 compute (locally ~2.3× vs f32 at K=N=4096, M=1).
- **CPU**: the block-quant packer dequantizes `Q4_0` → f32 and runs the f32 microkernel. So CPU
  gets the memory win (~7× smaller weights) but only a bandwidth-grade speedup (~1.1–1.33×),
  not int4 compute.

So there's clear CPU headroom. Since int4 isn't a compute type, a "true int4 CPU kernel" means
unpacking to int8 **or** f16 and riding an existing GEMM — and tract already has both. Two paths,
with a numerical-contract tradeoff:

## Option A — int8-dot (W4A8)
Unpack int4 → int8, quantize activations → int8, run the existing int8 GEMM kernels
(`sdot` #2278 / SME `smopa` #2279 / Intel AMX+VNNI #2339 / apple_amx).
- **Fastest** — rides all the int8 paths already in tree.
- **But**: activations become int8 too → effectively **W4A8**, a different contract than #2340
  (this is the llama.cpp `Q4_0 × Q8_0` approach; near-lossless in practice, but it *is* a change).

## Option B — f16-unpack (W4A16)
Unpack int4 → f16, use the existing f16 FMA kernels (`fmla` on NEON-fp16 / SVE).
- **Keeps #2340's exact contract** (f16/f32 activations).
- Faster than the current f32 dequant, slower than int8-dot. **No new ISA needed.**

## The shared part
Either way, the only new code is ISA-agnostic: a `Q4_0` → {int8 | f16} panel extractor (a variant
of `extract_panel`), plus — for A — on-the-fly activation quant + dual-scale combine. It reuses the
per-arch kernels already in tree; the one missing ISA piece is NEON `smmla` (I8MM), which would be a
bonus, not a prerequisite.

## Question
For the CPU int4 path, which numerical contract do you want — **A (W4A8, max speed)** or
[...]

PR #2348, opened 2026-06-07
linalg/block_quant: W4A8 int8-dot decode GEMV for Q4_0

czoli1976:
Adds a **W4A8** (4-bit weight, int8 activation) decode GEMV on the `Q4_0` block-quant format —
the CPU primitive for an int4-*compute* matmul path, as an alternative to today's
dequantize-to-f32-then-f32-kernel route. `Q4_0::w4a8_gemv` quantizes the activation row to int8
per block, dots it with the unpacked 4-bit weights, and scales each block by the weight and
activation scales.

### Why
The GPU backends already do exactly this — CUDA routes `Q4_0` to a `Q4_0 × Q8_1` kernel
(`cuda/src/kernels/cu/mm_mv_q.cu`), Metal to `mul_mv_q4_0`. On CPU the block-quant path only
dequantizes to f32, so the int4 win is just reduced weight bandwidth, not int8 compute. This is
the missing CPU piece.

### Notes / numbers
- **Near-lossless**: validated against the f32-dequant GEMV over the same `Q4_0` weight — the int8
  activation adds ≈0.3% (W4A8 ≈ W4A16); test included.
- **No intrinsics, no `unsafe`**: the contiguous `i8→i32` inner loop autovectorizes to a NEON
  `sdot` on its own (integer reductions are associative; strict-IEEE f32 ones aren't).
- Decode-shape (4096², M=1) it measured ~1.8× over `mmv_f32` in a local prototype — that includes
  op-wrapper overhead, so a kernel-selection integration would be ≥ that.

### Scope / integration
This PR is the **kernel primitive only** — it is intentionally *not* yet wired into kernel
selection. The integration question (the W4A8-vs-W4A16 numerical contract, and whether it lands as
a dedicated op like the Metal/CUDA `Q4_0×Q8` ops or via `strategize` in `einsum_matmul`) is the
open discussion in #2341. Filing the validated primitive + numbers to ground that — happy to wire
it whichever way you prefer.

### DIFF
--- a/linalg/src/frame/block_quant/q4_0.rs
+++ b/linalg/src/frame/block_quant/q4_0.rs
@@ -157,6 +157,62 @@ impl<const QK: usize> BaseQ4_0<QK> {
         }
         Ok(())
     }
+
+    /// Quantize one activation row to symmetric int8 with a per-block f32 scale (block length
+    /// `QK`): `scale_b = max(|a| in block) / 127`, `q[i] = round(a[i] / scale_b)`. Returns
+    /// `(q, scales)` with `q.len() == a.len()` and `scales.len() == a.len() / QK`.
+    fn quantize_row_q8(&self, a: &[f32]) -> (Vec<i8>, Vec<f32>) {
+        let nb = a.len() / QK;
+        let mut q = vec![0i8; a.len()];
+        let mut scales = vec![0f32; nb];
+        for b in 0..nb {
+            let blk = &a[b * QK..][..QK];
+            let amax = blk.iter().fold(0f32, |m, &v| m.max(v.abs()));
+            let scale = if amax > 0.0 { amax / 127.0 } else { 1.0 };
+            scales[b] = scale;
+            let r = scale.recip();
+            for i in 0..QK {
+                q[b * QK + i] = (blk[i] * r).round().clamp(-127.0, 127.0) as i8;
+            }
+        }
+        (q, scales)
+    }
+
+    /// W4A8 GEMV (`M == 1`): `y[n] = Σ_k dequant(W[n, k]) · a[k]`, computed as per-block int8 dot
+    /// products of the unpacked 4-bit weights with an int8-quantized activation, each scaled by
+    /// `w_block_scale · a_block_scale`. `weight` is this format's storage for a `[n, k]` matrix
+    /// (row-major blocks, `k % QK == 0`); `a` is the length-`k` activation, quantized to int8 on
+    /// the fly, so the result is near- but not bit-exact against the f32-dequant path.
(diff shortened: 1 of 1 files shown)


czoli1976:
@kali would benefit from the kernel getting into main, I am building on top of it for an upcoming PR

linalg/src/frame/block_quant/q4_0.rs
```diff
@@ -157,6 +157,62 @@ impl<const QK: usize> BaseQ4_0<QK> {
...
+        }
+        (q, scales)
+    }
+
+    /// W4A8 GEMV (`M == 1`): `y[n] = Σ_k dequant(W[n, k]) · a[k]`, computed as per-block int8 dot
+    /// products of the unpacked 4-bit weights with an int8-quantized activation, each scaled by
+    /// `w_block_scale · a_block_scale`. `weight` is this format's storage for a `[n, k]` matrix
+    /// (row-major blocks, `k % QK == 0`); `a` is the length-`k` activation, quantized to int8 on
+    /// the fly, so the result is near- but not bit-exact against the f32-dequant path.
+    pub fn w4a8_gemv(&self, weight: &[u8], n: usize, k: usize, a: &[f32]) -> TractResult<Vec<f32>> {
```
kali, line 186: Just a comment, I will approve as is, since this is a stepping stone.

I think we probably want the output to be &mut [f32], this could become a hotpath, and avoiding allocation may help later on.

kali (approved):

czoli1976:
@kali this one's now on my critical path, so I'd like to get it into main.

The consumer it was waiting for now exists: I'm making `block_quant` actually usable on imported transformers. #2428 fixes the orientation bug that made it crash on standard ONNX `activation @ weight` matmuls (BERT/MiniLM etc. fail on the first attention matmul today). With that fix block_quant runs correctly — but on CPU it's ~5.7× *slower* than f32 (MiniLM, seq 64: 8.7ms → 49.8ms), because the Q4_0 OptMatMul path dequantises the weights to f32 and then does an f32 GEMM. Memory saved, no compute saved.

The fix is precisely this kernel's approach: keep weights int8, quantise activations to int8, accumulate in int32. So my next step is to generalise this W4A8 GEMV to GEMM and route the block_quant codegen through it instead of the dequant-to-f32 path. Having it in main gives me a stable base for the GEMM + kernel-selection wiring rather than stacking on an unmerged branch.

It's approved and mergeable; the only red check is `cargo-deny`, which looks like an unrelated advisory rather than anything in this 82-line kernel — is that blocking on your side, or fine to merge through? And noted on `&mut [f32]` for the output — I'll fold the allocation-free buffer into the GEMM follow-up, where it's actually the hotpath.

kali:
i fixed the deny on main.

czoli1976:
mercie!

PR #2349, opened 2026-06-07
linalg: reusable CPU cache-size detection module

czoli1976:
Factors the L2 cache-size probe that drives the single-thread MMM block budget out of `frame/mmm/mod.rs` into a small, reusable `cache` module, so other heuristics (im2col lowering thresholds, kernel dispatch, future K-blocking) can read the same memoised numbers instead of re-implementing a platform probe.

`cache::cache_info()` returns L1d / L2 / L3 (bytes; `0` = undetected) from one lazily-detected, process-cached probe: macOS/iOS via `sysctlbyname` (FFI, replacing the previous `sysctl` subprocess; `hw.perflevel0.*` preferred on hybrid Apple Silicon), Linux **and** Android via the `/sys` cache walk keyed off each entry's reported level+type, and Windows via `wmic`. `block_budget_bytes` now reads `cache_info().l2`; the existing L2 budget formula and the 256 KiB unknown-cache fallback are unchanged, so single-thread blocking behaviour is preserved. Honest `0`-on-unknown semantics with `*_or_default` helpers for callers that want an arch-based guess.

Verified on Apple M-series (L1d 128 KiB, L2 12 MiB, L3 not exposed → reported 0): full `tract-linalg` suite green, `cargo fmt --all` clean, no new clippy findings.

Citations for the runtime cache-size detection driving blocking:

- Eigen queryCacheSizes() ([Eigen/src/Core/util/Memory.h](https://github.com/czoli1976/tract/blob/claude/loom-tract-optimization-xn9Rk/Eigen/src/Core/util/Memory.h)) — runtime L1/L2/L3 via CPUID (x86) / sysctl-sysconf, ~0.5 ms once per process. Direct analog of this module. ([ref](https://eigen.tuxfamily.narkive.com/mPHp5H3Y/automatic-cache-and-block-size-determination))

-  glibc sysconf(_SC_LEVELx_CACHE_SIZE); x86 CPUID leaf 0x04/0x18; Linux sysfs cacheinfo ([drivers/base/cacheinfo.c](https://github.com/czoli1976/tract/blob/claude/loom-tract-optimization-xn9Rk/drivers/base/cacheinfo.c)); ACPI PPTT (ARM cache topology).

- hwloc (Portable Hardware Locality) — cross-platform topology incl. shared/last-level caches. 

Note this PR does not make hwloc  a dependency, but borrows the idea from it

### DIFF
--- a/linalg/src/cache.rs
+++ b/linalg/src/cache.rs
@@ -1,0 +1,225 @@
+//! Best-effort runtime CPU data-cache geometry detection.
+//!
+//! Cache blocking (panel-block sizing in `mmm`, im2col lowering thresholds, …)
+//! is only correct when the block budget is derived from the *actual* cache the
+//! code runs on, not a hard-coded constant. This module centralises that
+//! detection so every heuristic reads the same memoised numbers instead of each
+//! re-implementing a platform probe.
+//!
+//! All sizes are **bytes**, with `0` meaning "could not detect on this platform"
+//! — callers must treat `0` as unknown and fall back conservatively (never
+//! over-block a cache you cannot see). The raw fields stay honest; the
+//! `*_or_default` helpers apply an architecture-based guess for callers that
+//! prefer a number to a zero.
+//!
+//! Detection is done once, lazily, and cached for the process lifetime.
+
+use std::sync::OnceLock;
+
+/// Detected data-cache sizes in bytes. `0` == unknown on this platform.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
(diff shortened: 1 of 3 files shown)


kali (approved):

czoli1976:
doing some cleanup: @kali : keep or close?

kali:
I'm interested. It's an enabler for matmul sizing optims. So I say keep.

Don't read too much into my limited pace at looking and integrating. It's not lack of interest, it's limited bandwidth.

PR #2350, opened 2026-06-07
linalg/mmm: L3-resident outer tier for the single-thread block walk

czoli1976:
**Stacked on #2349** (the reusable `cache` module). GitHub can't target a fork branch as the base, so this PR is opened against `main` and currently shows **both** commits — the first is #2349's cache module, the second (`linalg/mmm: add an L3-resident outer tier…`) is the actual content here. Once #2349 merges I'll rebase and the diff will drop to the single L3 commit. **Please review only the top commit; merge #2349 first.**

---

Adds a second cache-blocking tier to the single-thread MMM tile walk: the existing L2-resident inner block is wrapped in an outer super-block sized to **L3**, so a group of inner blocks stays L3-resident as the walk sweeps the grid (instead of re-fetching shared A/B panels from DRAM at large k).

The outer tier engages only when `cache::cache_info()` reports an L3 larger than L2; otherwise the outer edge spans the whole grid and the walk is byte-for-byte the current single-tier loop. Either way it's pure tile reordering — each tile still computes its own full-k reduction into a disjoint output region — so it stays **bit-exact** with the naive loop. The nesting is extracted into `for_each_blocked_tile` and unit-tested (visits every tile exactly once across single-tier / two-tier / degenerate edges; the no-L3 path matches the original order exactly).

**Not yet benchmarked** — hence draft. Apple Silicon doesn't expose `hw.l3cachesize`, so the outer tier is a guaranteed no-op on the hardware I can test; it only engages on x86/Linux boxes that report L3 via `/sys`. It's regression-safe by construction (no-op without L3, bit-exact with it), but the speedup claim needs an A/B on an L3-reporting machine at a large-k shape (grid > L2, super-block ⊂ L3).

Verified on Apple M-series: full `tract-linalg` suite green (incl. the real blocked-path bit-exactness tests + the new nesting tests), `cargo fmt --all` clean, no new clippy findings.

Citation for standard multi-level cache-blocked GEMM:

- Goto & van de Geijn, Anatomy of High-Performance Matrix Multiplication, ACM TOMS 2008 — the MC/KC/NC blocking; this outer tier is the NC (L3) block.
- BLIS (Van Zee & van de Geijn, ACM TOMS 2015); GotoBLAS/OpenBLAS; ATLAS (Whaley & Dongarra) — empirical block-size auto-tuning.
- Eigen manage_caching_sizes — derives GEMM block sizes from detected L1/L2/L3 (closest precedent for cache-derived blocking).
>> Contrast: Frigo et al., Cache-Oblivious Algorithms, FOCS 1999 — the no-cache-sizes alternative this PR deliberately doesn't take.

### DIFF
--- a/linalg/src/cache.rs
+++ b/linalg/src/cache.rs
@@ -1,0 +1,225 @@
+//! Best-effort runtime CPU data-cache geometry detection.
+//!
+//! Cache blocking (panel-block sizing in `mmm`, im2col lowering thresholds, …)
+//! is only correct when the block budget is derived from the *actual* cache the
+//! code runs on, not a hard-coded constant. This module centralises that
+//! detection so every heuristic reads the same memoised numbers instead of each
+//! re-implementing a platform probe.
+//!
+//! All sizes are **bytes**, with `0` meaning "could not detect on this platform"
+//! — callers must treat `0` as unknown and fall back conservatively (never
+//! over-block a cache you cannot see). The raw fields stay honest; the
+//! `*_or_default` helpers apply an architecture-based guess for callers that
+//! prefer a number to a zero.
+//!
+//! Detection is done once, lazily, and cached for the process lifetime.
+
+use std::sync::OnceLock;
+
+/// Detected data-cache sizes in bytes. `0` == unknown on this platform.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct CacheInfo {
+    /// L1 data cache (per core), bytes. `0` if unknown.
+    pub l1_data: usize,
+    /// L2 cache (per perf-core / cluster), bytes. `0` if unknown.
+    pub l2: usize,
+    /// L3 / last-level cache, bytes. `0` if unknown.
+    pub l3: usize,
+}
+
+impl CacheInfo {
+    /// L1 data cache, or an architecture-based guess when undetected
+    /// (64 KiB on arm64, 32 KiB elsewhere — matches common silicon).
+    pub fn l1_data_or_default(&self) -> usize {
+        if self.l1_data > 0 {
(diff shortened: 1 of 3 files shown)


czoli1976:
When [#2349](https://github.com/sonos/tract/pull/2349) merges, ping me and I'll rebase [#2350](https://github.com/sonos/tract/pull/2350).

czoli1976:
tested on a Xeon container(*)

image

(*) = Intel Xeon @ 2.80 GHz, 4 cores, AVX2 + AVX-512 (F/DQ/BW/VL/CD) + AVX-512-VNNI, F16C/FMA. Caches: L1d 32 KB/core, L2 1 MB/core, L3 33 MB shared.

czoli1976:
To set expectations:

Expect gain on Cortex-A7X (A72…A78 / A710 / A715 — big cores) and any other Big Boy CPU (not on A5x as no L3 Cache over there)

kali:
Rebased!

czoli1976:
Closing as subsumed: this branch was the base of the stack under #2352, and #2352 was squash-merged into `main` with base `main` (commit a443189290b3), so the squash carried this PR's L3-resident outer tier into `main` along with the LLC/SLC budget. The outer-tier code (`for_each_blocked_tile`/`blk_outer`/`st_outer_block_edge`) is already in `main`; this branch has no unique content left (its tip predates the #2352 refinements), so there's nothing further to merge here.

Follow-up coming separately: a working-set gate so the outer tier only engages when A+B actually spill the LLC — addresses the `voicecom_float` +15.6% on jetson-orin-nx that was the lone regression after #2352.

kali:
Ha, apologies for the confusion.

PR #2351, opened 2026-06-07
linalg: add Q2_0_T ternary (BitNet b1.58) block-quant format

czoli1976:
2-bit packed ternary weights {-1,0,+1} + one f16 abs-mean scale per block (2.5 bits/weight at QK=32). Registered as generic-kernel packings with an inline dequant handler, plus an AVX2 panel extractor (packed_32_q1_58_to_f32, ~8.8x faster than scalar). Validated against the f32 reference via the auto-generated packed-packed and panel-extractor suites.

Citations: BitNet b1.58 (arXiv:2402.17764), llama.cpp TQ1_0/TQ2_0, Microsoft bitnet.cpp (I2_S/TL1/TL2) + T-MAC.

### DIFF
--- a/linalg/src/frame/block_quant/helpers.rs
+++ b/linalg/src/frame/block_quant/helpers.rs
@@ -71,3 +71,75 @@ impl<W: Write> NibbleWriter<W> {
         self.writer.write_i8(q).unwrap()
     }
 }
+
+/// Reads/writes 2-bit values ("crumbs"), four per byte, least-significant pair first.
+/// Used by the ternary (Q1_58 / BitNet b1.58) block-quant format. f16 reads/writes are
+/// only legal on a crumb (byte) boundary, which the ternary layouts always respect.
+pub struct CrumbReader<R> {
+    acc: u8,
+    remaining: u8,
+    reader: R,
+}
+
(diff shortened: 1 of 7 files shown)


kali (commented):
Just a terminology / naming doubt. is Q1_58 the right name ? should it be qualified it with the TQ2_0 (if i'm not mistaken) ? Or do we assume the other encoding are irrelevant ?

czoli1976:
what abt Q2_0_T ?

czoli1976:
Q2_0_T / TQ2  that says "2-bit ternary" and matches the Q4_0 naming style sounds more appropriate

TQ2_0 may not be good as It is actually TQ2_0-style, not byte-identical,In fact llama.cpp's TQ2_0 uses 256-element superblocks with a specific interleave; yours is a simpler QK=32 layout. TQ2_0 would imply a wire-compatibility and not 1:1 equivalent.

czoli1976:
@kali what's your pick ?

kali:
Q2_0_T feels right.

czoli1976:
image

kali (approved):

PR #2363, opened 2026-06-12
move clean Resize subset and GridSample into tract-core

JulienBalianSonos:
Moves the common Resize path and GridSample out of `onnx-opl` into `tract-core`, leaving the ONNX edge cases behind a declutter, and exposes them as first-class NNEF ops.

- `core::ops::nn::{Resize, GridSample}` — clean Resize subset (4 coord transformers × {linear, nearest, cubic} × {Floor, RoundPreferCeil}, fixed `cubic_coeff_a=-0.75`, no ROI / `exclude_outside`) plus the shared cubic kernel and the nearest-integer-upsample lowering; GridSample moved verbatim.
- `onnx-opl::Resize` reduced to the edge-case op (ROI, `exclude_outside`, non-default `cubic_coeff_a`, the extra nearest modes, the `axes` attribute); decltters into `core::Resize` when it fits the subset, else falls back to the lowering.
- NNEF: `tract_core_resize` (optional `scales` **or** `sizes` tensor input) and `tract_core_grid_sample`. The stdlib `nearest_upsample` / `multilinear_upsample` build `core::Resize` instead of descending to `debox` (active under `extension tract_core`). A load-only `tract_onnx_grid_sample` alias is kept for back-compat.
- Fixes a pre-existing border bug: the linear/nearest 2-tap eval cast the source coordinate with `as usize`, extrapolating past the top/left edge for negative half-pixel coordinates; it now floors and clamps, matching ONNX and PyTorch.
- f16/non-f32 inputs are cast to f32 for compute and back to the input type (the eval previously returned f32 regardless, mismatching the output fact).

Consumed by **sonos/torch-to-nnef#205**, which exports the full PyTorch resize/grid_sample family (`upsample_*`, `_upsample_nearest_exact*`, `grid_sampler*`) to these ops.

Test plan:
- `cargo test -p tract-core` resize unit tests: 3 ✓
- `cargo test -p test-onnx-core` resize 48 / gridsample 52 / upsample 24 — 0 failures (covers the optimized declutter path)
- `cargo test -p test-nnef-cycle` resize 24 / gridsample 16 / upsample 12 — 0 failures (dump→reload round-trip)
- `harness/nnef-test-cases`: `resize/{nearest_upsample,core_resize_nearest,core_resize_sizes,core_resize_linear,core_resize_f16}`, `grid_sample/grid_sample_identity` — all pass; corrupting a golden fails as expected
- `cargo fmt --all --check` clean; no clippy errors / new warnings


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali (approved):

kali:
Rebased for benches, even if it's not super relevant, I'm still testing the workflow :)

kali:
Interesting, I'm pretty sure you're not responsible for these improvements (I guess we're comparing to the wrong, or "a" wrong reference point).

PR #2366, opened 2026-06-13
metal: f16 activations/output for the GGML matmul (drop f32 round-trips)

czoli1976:
## What

The GGML matmul kernels hardcoded **f32 output**, and the q4_0 / f16-weight GEMV+GEMM paths required **f32 activations**. So a `q40ef16` model (Q4_0 weights + f16 activations — the common on-device LLM layout, e.g. the Qwen builds in `examples/causal_llm`) bounced through f32 at *every* matmul: the metal transform inserted a `f16→f32` cast on the activation and a `f32→f16` cast on the output.

This makes the GGML matmul keep f16 end-to-end:

- **`ggml_mm_mv.metal`**
  - `kernel_mul_mv` output pointer is now the activation type `T1` (f16 activations → f16 output; `*_f32` paths unchanged).
  - The q4_0 GEMV (`mul_vec_q_n_f32_impl`) is templated on the activation/output type — new `kernel_mul_mv_q4_0_f16`, still accumulating in f32.
  - The GEMM (`kernel_mul_mm`) is templated on the activation/output type: f16 activations are converted to f32 in threadgroup memory on load, and f16 output is written through the f32 simdgroup scratch (`simdgroup_store` can only target a float buffer). New `kernel_mul_mm_f16_f16` / `kernel_mul_mm_q4_0_f16`.
- **`ggml_gemm/mod.rs`**: `output_dt` returns the activation dtype; the GEMV/GEMM dispatch and dtype guards accept f16 activations and select the f16 kernels.
- **`transform.rs`**: drop the forced `f16→f32` activation upcast; with `output_dt` following the activation, the post-matmul `f32→f16` cast also becomes a no-op and is no longer inserted.

## Correctness

- All **53 `tract-metal` GPU tests pass**, including a **new `mmm_ggml_prop_q4_f16`** prop test (q4_0 weights × f16 activations vs an f32 CPU reference).
- End-to-end on Qwen3-1.7B q40ef16 (Metal), greedy decode output is **identical** before/after.

## Benchmark

Qwen3-1.7B q40ef16, Metal decode, via the added `examples/causal_llm` `complete_bench` bin (mean of 3 × 96 tokens):

| | tok/s | ms/token |
|---|---:|---:|
| baseline (f32 round-trip) | ~41.6 | 24.0 |
| **this change (f16 direct)** | **~45.6** | **21.9** |

**≈10% faster decode**, identical output. The win comes from dropping ~2 cast dispatches per matmul (~200 matmuls/token) plus reading f16 activations / writing f16 output instead of f32.

## No clash with open PRs

The only open-PR edit to the matmul area is #2320 flipping `mod mfa` → `pub mod mfa` on line 1 of `matmul/mod.rs`. This PR touches the `ggml_gemm` kernels, `output_dt`, the matmul lowering, and adds a test at the end of `matmul/mod.rs` — no overlap. Confirmed mergeable.

## Files
- `metal/src/kernels/matmul/ggml_gemm/ggml_mm_mv.metal` — templated activation/output dtype
- `metal/src/kernels/matmul/ggml_gemm/mod.rs` — `output_dt` + dispatch
- `metal/src/transform.rs` — drop activation upcast
- `metal/src/kernels/matmul/mod.rs` — `mmm_ggml_prop_q4_f16` test
- `examples/causal_llm/src/bin/complete_bench.rs` — decode benchmark

### DIFF
--- a/metal/src/transform.rs
+++ b/metal/src/transform.rs
@@ -366,11 +366,9 @@ fn convert_matmul_to_metal(
                 inputs[a_pos] = target.wire_node(perm_a_name, perm_a_op, &[inputs[a_pos]])?[0];
             }
 
-            if input_facts[0].datum_type == DatumType::F16 {
-                let in_cast_op = metal_cast_new(DatumType::F32).unwrap();
-                inputs[0] =
-                    target.wire_node(node.name.clone() + ".in_cast", in_cast_op, &[inputs[0]])?[0];
-            }
+            // The GGML kernels now consume f16 activations directly (and emit
+            // f16 output via output_dt), so no f16->f32 activation upcast is
+            // inserted here anymore.
 
             if !op.transpose_b {
                 ensure!(
--- a/examples/causal_llm/src/bin/complete_bench.rs
+++ b/examples/causal_llm/src/bin/complete_bench.rs
@@ -1,0 +1,47 @@
+//! Decode-throughput benchmark for the Metal GGML f32-round-trip change.
+//! Loads a causal LLM (default = Metal runtime), prefills a prompt, then times
+//! N decode steps and prints tokens/sec + the generated text (for a correctness
+//! eyeball). Usage: complete_bench <tokenizer.json> <model.nnef.tgz> [n] [--cpu]
+use anyhow::Result;
+use causal_llm::{CausalLlmModel, CausalLlmModelConfig};
+use std::time::Instant;
+
+fn main() -> Result<()> {
+    env_logger::init();
+    let args: Vec<String> = std::env::args().collect();
+    let tokenizer = &args[1];
+    let model = &args[2];
+    let n: usize = args.get(3).and_then(|s| s.parse().ok()).unwrap_or(64);
+    let force_cpu = args.iter().any(|a| a == "--cpu");
+
+    let conf = CausalLlmModelConfig { force_cpu };
+    let t0 = Instant::now();
+    let llm = CausalLlmModel::from_paths_and_conf(tokenizer, model, conf)?;
+    eprintln!("model load: {:?} (force_cpu={force_cpu})", t0.elapsed());
+
+    let prompt = "The capital of France is";
+    let mut state = llm.spawn()?;
+    state.append_text(prompt)?;
+    let prompt_len = state.encode(prompt, true)?.len();
+
+    // Prefill + warm up.
+    for _ in 0..3 {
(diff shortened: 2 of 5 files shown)


metal/src/transform.rs
```diff
@@ -393,11 +366,9 @@ fn convert_matmul_to_metal(
...
             }
 
-            if input_facts[0].datum_type == DatumType::F16 {
-                let in_cast_op = metal_cast_new(DatumType::F32).unwrap();
-                inputs[0] =
-                    target.wire_node(node.name.clone() + ".in_cast", in_cast_op, &[inputs[0]])?[0];
-            }
+            // The GGML kernels now consume f16 activations directly (and emit
+            // f16 output via output_dt), so no f16->f32 activation upcast is
+            // inserted here anymore.
```
kali, line 371: I've put instruction in CLAUDE.md to avoid this kind of silly comments. But of course, claude will only read it when it feel like it.

I did instruct "mine", in "memory", to read claude.md before any work on tract. I feel like it helped a bit.  But i could not find a way to force yours or anybody else's to do the same, so I suggest you give yours a similar permanent rule. If you don't mind.

(Something like: "In memory: always read  CLAUDE.md before doing work on tract.").

kali (commented):
I wonder about the templating approach. I think a regular parameter should do the trick here, avoiding duplicating the entire code generation. It create a branching in the kernel, but all thread will take the same one so it should not hurt. WDYT ?

metal/src/transform.rs
czoli1976: You are right, kinda keeps it doing

czoli1976:
> I wonder about the templating approach. I think a regular parameter should do the trick here, avoiding duplicating the entire code generation. It create a branching in the kernel, but all thread will take the same one so it should not hurt. WDYT ?

OK, let's try like that, good idea

czoli1976:
@kali done

kali:
@czoli1976  Mmmm... pretty strong signal here.

czoli1976:
@kali M1-Max bench again, please

kali (approved):

kali:
Sorry for the rotten idea.

czoli1976:
> Sorry for the rotten idea.

It was worth a try now got an hydrid between the two, the bigger kernel became too big and that seemed to be the problem, let's see the bench

kali:
ha !

PR #2370, opened 2026-06-14
examples/causal_llm: speculative decoding (n-gram + draft-model)

czoli1976:
Adds speculative decoding to the `causal_llm` example: a cheap drafter proposes the next few tokens, the target model verifies them in a single forward pass and accepts the longest greedy-matching prefix plus its own next token. Output is identical to plain greedy decoding.

## What's here

- **`expose_all_logits` transform** (`transformers`) + **`Nnef::load_without_decluttering`** (`api/rs`): exports emit last-token-only logits, but verifying K drafts needs a next-token distribution at every draft position. The transform recomputes the final projection over the full hidden states (the last-position values are unchanged, so it stays a drop-in at decode time). It must run before declutter rearranges the last-token slice, hence the raw-load hook.
- **`generate_speculative`** with KV-cache rollback: one forward over `[tail · drafts]`, greedy-verify, accept the matching prefix + correction, truncate the cache to drop rejected drafts. A distribution-preserving rejection sampler (Leviathan/Chen) is included for `temperature > 0`.
- **Two drafters**: `NgramDrafter` (prompt-lookup, no second model) and `ModelDrafter` (a smaller causal LM).
- **Benchmarks**: `bench_spec` (E2E throughput + acceptance + lossless check), `bench_micro` (forward latency vs tokens-per-pass), and a criterion host-side bench.

## Correctness

Greedy speculative output is **bit-identical** to `generate_next_token`, verified on real models (Qwen3-1.7B and Llama-3.2-1B). The real-model tests are gated on a local `.cached` model path and skip when absent, matching the existing `truncate_prefix_cache_hit` test; the verification/drafter logic also has model-free unit tests. Lossless in exact arithmetic; rare divergences on high-entropy text come from floating-point differences between batched verification and single-token decode (standard for speculative decoding).

## Performance

n-gram speculation on Qwen3-1.7B (Metal): ~**1.2–1.4×** decode speedup at k=2 on repetitive / code-like text; neutral-to-slower on prose (low acceptance) and at larger k. The useful draft length is currently capped by the q4 GEMV→GEMM kernel selection; #2369 widens that and lifts k=4 from a slowdown to ~1.19×.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -532,6 +532,7 @@ dependencies = [
  "axum",
  "axum-macros",
  "clap",
+ "criterion",
  "env_logger",
  "float-ord",
  "log",
--- a/examples/causal_llm/Cargo.toml
+++ b/examples/causal_llm/Cargo.toml
@@ -22,6 +22,13 @@ tokio-scoped = "0.2.0"
 rand.workspace = true
 
 
+[dev-dependencies]
+criterion.workspace = true
+
+[[bench]]
+name = "spec_host"
+harness = false
+
 # tokenizer special handle for wasm
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 tokenizers.workspace = true
--- a/examples/causal_llm/README.md
+++ b/examples/causal_llm/README.md
@@ -62,3 +62,41 @@ for _ in 0..20 {
     state.generate_next_token()?;
 }
 ```
+
+## Speculative decoding
+
+A cheap *drafter* proposes several tokens; the target verifies them in one
+forward pass and accepts the longest greedy-matching prefix plus its own next
+token. Output is identical to plain greedy decoding (`generate_next_token`).
+
+The target must expose logits for every input position. Export it with
+all-position logits — torch2nnef `--num-logits-to-keep 0` (matching
+transformers' `logits_to_keep == 0`) — so there is no last-token slice in the
+graph, then load it with `from_paths_speculative`. Two drafters are provided:
+`NgramDrafter` (prompt-lookup, no second model) and `ModelDrafter` (a smaller
+causal LM).
+
+```rust
+let llm = CausalLlmModel::from_paths_speculative("tokenizer.json", "model.nnef.tgz", Default::default())?;
+let mut state = llm.spawn()?;
+state.append_text("The capital of France is")?;
+let mut drafter = NgramDrafter::default();
+while state.seq.len() < target_len {
+    let stats = state.generate_speculative(&mut drafter, 4)?; // k = 4
+}
+```
+
+Benchmarks:
+
+- `cargo run --release --bin bench_spec -- -t tok.json -m model.nnef.tgz` —
+  decode throughput, acceptance, and speedup of greedy vs n-gram speculation
+  across prompt types and draft lengths (checks losslessness each run).
+- `cargo run --release --bin bench_micro -- -t tok.json -m model.nnef.tgz` —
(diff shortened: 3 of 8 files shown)


czoli1976:
@kali look at me, I am pretty

czoli1976:
clean up: @kali keep or close?

kali:
This one is interesting, but needs a bit of work.

* I don't want to change the API scope. I ditch the "declutter" bit from the API because it was confusing to users, misunderstood and misused, I'd like to push back before bringing it back. So we need a model without the "last token only" slicing. I think we have options that suppose fixing the model before hard. It may be a flag in torch2nnef (or we may ask @JulienBalianSonos to make it a flag). We could untar/patch/retar the nnef. We could have an adhoc tract-based executable that link into the private crates to "fix" the model as a one-shot.
* I'd like to keep the causal llm example relatively simple and easy to understand (you may argue it's already a miss) so I'm concerned about adding too much layers on top. Maybe the right option is to have a very stripped down example, single-file-ish, absolute minimal demonstrator (no temperature, no http, no tweaks), while opening a more ambitious chatbot runner on the side.

JulienBalianSonos:
Hi Mathieu & @czoli1976,

I agree with the vision: causal_llm should stay a minimal core example that other packages build on top of or copy. So the goal is to keep its API and abstractions flexible enough for that, not to grow features inside it.

On "a model without the last-token-only slicing": t2n already does this. No untar/patch/retar, no private-crate fixer.

There is a num_logits_to_keep flag (--num-logits-to-keep, default 1), and 0 means "keep all positions". Exporting with --num-logits-to-keep 0 gives a logits output of shape [B, S, vocab] with no last-token narrow, on every architecture path. It matches transformers' own convention (logits_to_keep=0 == keep all). So the "make it a flag in torch2nnef" ask is already done.

To get the best of both worlds, the durable optimization is a dynamic logits_to_keep in t2n (opened as [sonos/torch-to-nnef#216](https://github.com/sonos/torch-to-nnef/issues/216)). It serves prompt and speculative from one artifact, and it makes both the tract transform and the load_without_decluttering API change unnecessary.

What this means for the PR: exporting all-position from t2n (or ONNX) lets us drop the `Nnef::load_without_decluttering` addition entirely and keep causal_llm's contract unchanged. If the model is exported properly there is no slice in the graph to strip. So no expose_all_logits transform, no raw load, and the model goes through the normal fully decluttered from_paths path. That removes exactly the "declutter bit" you did not want back in the API.

@czoli1976 thanks for the demonstration of this common LLM optimization, and for surfacing what tract and causal_llm would need to support it. Worth keeping the reference.

czoli1976:
ok, shall I close it then ?

kali:
I think we drop the "model patching" bit by using t2n export correctly, and we keep the speculative decoding runtime part.

And in another PR, we'll strip down causal_llm to derive a new simple_causal_llm.

czoli1976:
Done — dropped the model-patching: removed `expose_all_logits` and the `load_without_decluttering` API hook, so the API scope is unchanged. The target now just needs to be exported with all-position logits (`--num-logits-to-keep 0`, thanks @JulienBalianSonos) and loads through the normal decluttered path; `generate_speculative` checks the logit row count at runtime. Rebased onto main, runtime + benches kept. The model-dependent tests skip unless the export has all-position logits. Ready for another look.

kali (dismissed):

czoli1976:
merge ?

kali (approved):

PR #2381, opened 2026-06-16
feat(causal_llm): add LogitsProcessor hook for guidance / constrained decoding

JulienBalianSonos:
## What

Adds a small, generic hook to `CausalLlmState` so callers can adjust the
next-token logits before selection — the building block for constrained
decoding / guidance (grammar or JSON-Schema masking, logit biasing, custom
sampling), without baking any such logic into `causal_llm`.

```rust
/// Adjust next-token logits before selection (constrained decoding / guidance).
pub trait LogitsProcessor {
    fn process(&mut self, logits: &mut [f32], tokens: &[u32]);
}

impl CausalLlmState {
    pub fn generate_next_token(&mut self) -> Result<()>;                  // unchanged
    pub fn generate_next_token_with(&mut self, p: &mut dyn LogitsProcessor) -> Result<()>; // new
}
```

- `LogitsProcessor::process` is called once per token, **after** the repeat
  penalty and **just before** argmax, with the full token sequence so far.
- A blanket impl makes any `FnMut(&mut [f32], &[u32])` a `LogitsProcessor`, and
  `NoLogitsProcessor` is a no-op default.
- `generate_next_token()` now simply delegates to `generate_next_token_with(&mut NoLogitsProcessor)`, so **existing behavior and the public API are unchanged**.

## Why

The sampler does argmax internally with no way to influence it, so downstream
projects can't do grammar/JSON-constrained decoding. This is the minimal seam
to enable it.

We've used it downstream to drive [llguidance](https://github.com/guidance-ai/llguidance)
for guaranteed JSON / JSON-Schema / Lark-grammar / regex output. **None of that
lands here** — `causal_llm` gains no dependency; the hook is engine-agnostic.

## Design notes

- Minimal & non-invasive: no struct fields added, so `Debug`/`freeze`/`Send` of
  `CausalLlmState` are untouched. The processor is passed per call.
- Masking pattern: set disallowed tokens' logits to `f32::NEG_INFINITY`, then
  the existing argmax picks the best allowed token.

## Tests

Added unit tests (no model needed): a closure masks a token; `NoLogitsProcessor`
leaves logits unchanged. `cargo check -p causal_llm` passes.

### DIFF
--- a/examples/causal_llm/src/lib.rs
+++ b/examples/causal_llm/src/lib.rs
@@ -225,6 +225,35 @@ impl CausalLlmStateConfig {
     }
 }
 
+/// Hook to adjust the next-token logits before selection — the generic
+/// mechanism for constrained decoding / guidance (masking tokens that would
+/// break a grammar or JSON schema, biasing, etc.).
+///
+/// Called once per generated token by
+/// [`CausalLlmState::generate_next_token_with`], after the repeat penalty and
+/// just before the token is picked.
+pub trait LogitsProcessor {
+    /// `logits` holds the score for every vocabulary token of the next
+    /// position; `tokens` is the full sequence generated so far (prompt +
+    /// output). Mutate `logits` in place — e.g. set disallowed entries to
+    /// `f32::NEG_INFINITY` to forbid them.
+    fn process(&mut self, logits: &mut [f32], tokens: &[u32]);
+}
+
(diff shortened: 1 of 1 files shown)


JulienBalianSonos:
Thinking a bit more on that PR, + comments in #2370 by Kali (and recent changes making 'causal_llm' an example). It feels inappropriate and should be part of another crate. Discarded to avoid expanding the 'example' and reducing its readability.

PR #2394, opened 2026-06-20
Revert "linalg/arm64simd: keep proven kernel speedups" (#2393)

kali:
The arm64 kernel micro-opts shifted f16 logits enough to regress the LLM RBO nightly: meta-llama--Llama-3.2-3B-Instruct-q40ef16.p0s100.arm64.cpu dropped 0.94 -> 0.87, with smaller arm64 drops on other models. The change is arm64-only; x86_64 cpu and the gpu paths of the same models are unaffected. #2393 was validated for speed only (kernel-microbench ns), and the arm64 softmax/activation kernel tests run at Approximation::SuperApproximate, too loose to catch a logit ranking shift. The end-to-end speed it bought is negligible against matmuls, so restore the proven-correct composed silu/gelu kernels and the pre-fcvtnu softmax clamp. Re-land individual variants gated by an RBO run if the speed is wanted.

This reverts commit fbbacc468f8aafe88aa60a4b57754d82375f8a02.

### DIFF
--- a/linalg/src/arm64/arm64simd/gelu.rs
+++ b/linalg/src/arm64/arm64simd/gelu.rs
@@ -13,9 +13,28 @@ ew_impl_wrap!(
     (),
     #[inline(never)]
     fn run(buf: &mut [f32], _: ()) {
-        // Keep the composed symbol but route to the single-pass fused kernel:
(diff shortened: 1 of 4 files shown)


czoli1976:
try PR #2395
image

PR #2352, opened 2026-06-07
linalg/cache sorcery: LLC/SLC-aware budget for the L3 outer blocking tier

czoli1976:
Stacked on #2349 (cache module) + #2350 (L3 outer tier). **Review only the top commit** — GitHub shows the whole stack until the lower PRs merge (a fork branch can't be a PR base in upstream).

## Problem

#2350's outer tier fires only when the per-CPU cache topology exposes an architectural L3 (`cpu/cache` `level 3`). Many Arm SoCs (and Apple) have **no cluster L3** but a **System-Level Cache (SLC)** shared with the GPU/NPU/display (Qualcomm LLCC, Apple SLC), which is never listed under `cpu/cache/index*` — so those parts see `l3 == 0` and miss a real last-level cache the tier could use.

"LLC" = role (last cache before DRAM). "SLC" = a specific shared interconnect cache that is *one kind* of LLC. This generalises "detect L3" → "establish the LLC budget," and treats a contended SLC more conservatively than a dedicated L3.

## Change

New `cache::last_level_cache() -> Option<(usize, LlcKind)>` resolving, first hit wins:

1. `TRACT_LLC_BYTES` env override (`"8M"`, `"33554432"`); kind = `SystemLevel` iff `TRACT_LLC_CONTENDED` is set, else `Dedicated`.
2. architectural **L3** (`cache_info().l3 > l2`) → `Dedicated`.
3. best-effort **Linux devicetree** SLC probe (`cache-level == 3` + `cache-size`, outside `/cpus`) → `SystemLevel`.

`l3_block_budget_bytes()` now budgets a `Dedicated` L3 at ~½ and a contended `SystemLevel` cache at ~¼ (can't assume residency of lines the GPU/NPU keep evicting). Unknown ⇒ `None` ⇒ no outer tier (unchanged, regression-safe).

## Notes

- Purely additive; default behavior on parts with a normal L3 is identical (L3 path, ½ budget).
- SLCs whose size is fixed in the controller (e.g. Qualcomm LLCC carries no `cache-size` in DT) fall to the `TRACT_LLC_BYTES` override — documented.
- Pure resolver `resolve_llc()` is unit-tested for priority/regression-safety; the DT probe is panic-tested.

## Prior art

- **Runtime cache sizing:** Eigen `queryCacheSizes` (CPUID/sysctl) + `manage_caching_sizes`; glibc `sysconf(_SC_LEVELx_CACHE_SIZE)`; ACPI PPTT; hwloc.
[...]

PR #2410, opened 2026-06-25
linalg/mmm: gate the L3 outer tier on the working set spilling the LLC

czoli1976:
Follow-up to #2352 (squash-merged as `a443189290b3`), which was the lone post-merge bench regression:

> 🔴 **+15.6%** `voicecom_float` · `jetson-orin-nx` · 13.4 ms → 15.5 ms

### Root cause

`st_outer_block_edge` engaged the L3 outer super-block tier whenever an L3 > L2 was detected — it never checked whether the matmul's operands actually exceed the LLC. `voicecom_float` is a small model whose packed `A + B` already sit inside the Orin's 2 MB cluster L3 (confirmed on the bench box: `index3` = `level 3`, `2048K`, `shared_cpu_list 0-3`). So the tier fired anyway, reordering tiles to chase reuse that was already resident — buying nothing and only disturbing the A78AE hardware prefetchers.

This is a missing precondition, not a wrong cache value: the tier was introduced precisely for the case where *"a grid that exceeds L2 still re-fetches A/B from DRAM"*, but that guard was never actually applied.

### Fix

Engage the outer tier only when the packed working set spills the LLC:

```
(m·mr + n·nr)·k·elem  >  llc_bytes
```

Otherwise stay single-level — the existing `usize::MAX` no-op path, bit-exact. `l3_block_budget_bytes` now returns the raw LLC size alongside the budget, so the fit check and the super-block sizing read one source of truth.

Conservative by construction: it only *skips* when the operands provably fit, so it cannot lose the tier's wins (those have `A + B ≫ L3`). Pure tile reordering either way — results are unchanged.

### Tests

- New `outer_tier_gated_on_working_set_spilling_llc` covers fit/spill/boundary/`k=0`/unknown-LLC.
- Existing `covers_every_tile_once` and `outer_max_matches_single_level` (permutation + single-level equivalence) still pass.
- `cache::tests` and the `single_thread_blocking` end-to-end blocked-matmul tests pass.

The bench bot will confirm `voicecom_float` on `jetson-orin-nx` returns to neutral.

### DIFF
--- a/linalg/src/frame/mmm/mod.rs
+++ b/linalg/src/frame/mmm/mod.rs
@@ -305,10 +305,13 @@ fn l2_block_budget_bytes() -> usize {
     tier_budget_bytes(crate::cache::cache_info().l2, 1, 3, 256 * 1024)
 }
 
-/// Outer tier: ~half of L3, but only when an L3 is detected and is meaningfully
-/// larger than L2 (otherwise an outer tier just duplicates the inner one).
-/// `None` ⇒ no outer tier; the walk stays single-level (identical to before).
-fn l3_block_budget_bytes() -> Option<usize> {
+/// Outer tier: `(llc_bytes, budget_bytes)` — the raw last-level-cache size and the
+/// fraction of it a single thread may budget for the outer super-block — but only
+/// when an L3/LLC larger than L2 is detected (otherwise an outer tier just
+/// duplicates the inner one). `None` ⇒ no outer tier; the walk stays single-level
+/// (identical to before). The raw size is returned alongside the budget so the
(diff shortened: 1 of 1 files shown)


czoli1976:
@kali let's measure again voicecom_float on jetson-orin-nx vs main, want to see if this proves the theory of prefetched being disturbed

kali:
I don't even want to guess what the temperature is in the lab...

czoli1976:
> I don't even want to guess what the temperature is in the lab...

It must be closer to an AirFryer than a Lab now :P

czoli1976:
merge me ?

kali (approved):

PR #2413, opened 2026-06-25
fix(data): reject raw tensor whose data length != declared shape (#2390)

### LINKED ISSUE #2390: Panic (copy_from_slice length mismatch) loading an ONNX initializer whose raw_data size != declared shape
## Summary

`tract-onnx` builds a tensor for an ONNX initializer by allocating `∏(dims) × sizeof(dtype)` bytes and then `copy_from_slice`-ing the protobuf `raw_data` into it, with **no check that the two lengths match**. A crafted (or merely malformed) `.onnx` whose `raw_data` byte count does not match the declared shape causes a `copy_from_slice` panic, aborting the process during model load.

This was reported privately by email on 2026-05-28; the disclosure window has elapsed, so filing publicly. Re-verified **live on `main` @ `e7f4449`** (2026-06-18).

## Affected version

`main` @ `e7f4449` (current at time of filing). Fires in release builds — `copy_from_slice` validates lengths regardless of optimization level.

## Root cause

The raw-tensor load path performs no length reconciliation:

- `onnx/src/tensor.rs` — `if t.raw_data.len() > 0 { create_tensor(shape, dt, &t.raw_data) }`
- `onnx/src/tensor.rs` — `create_tensor` dispatches by dtype to `Tensor::from_raw::<T>(&shape, data)`
- `data/src/tensor.rs:502` — `from_raw_dt_align`:

```rust
let mut tensor = unsafe { Tensor::uninitialized_aligned_dt(dt, shape, align) }?;
tensor.as_bytes_mut().copy_from_slice(content);   // data/src/tensor.rs:502 — panics if lengths differ
```

`tensor.as_bytes_mut()` has length `∏(shape) × sizeof(dtype)`, computed from the attacker-controlled `dims`. `content` is the attacker-controlled `raw_data`. Nothing ensures they are equal before the copy.

## Reproduction

The panic occurs during model construction, reached by any standard load:

```rust
let _ = tract_onnx::onnx().model_for_path("poc.onnx"); // panics
```

Observed on `e7f4449`:

```
thread '...' panicked at data/src/tensor.rs:502:31:
copy_from_slice: source slice length (12) does not match destination slice length (440)
```

PoC `.onnx` (base64; an initializer declaring a shape that needs 440 bytes while supplying 12 bytes of `raw_data`):

```
CA06TAovEgFZIghDb25zdGFudCogCgV2YWx1ZSoUCBQICxAFSgwAzACACAAAuL+/vQCgAQQSCGNvbnN0YW50Qg8KCAAAAAAAAAABWRIKCggIQRIEGgIJA0IECgAQEQ==
```

Decode with: `base64 -d > poc.onnx` (paste the block above).

## Impact

Denial of service: a small `.onnx` aborts any process that loads it. No memory-safety consequence — `copy_from_slice` panics rather than reading out of bounds — but the panic is unrecoverable for the loading thread/process. tract loads externally-supplied ONNX (model exchange, pipeline artifacts), so this is reachable from untrusted or partially-trusted files.

## Suggested fix

Validate the length before the copy, e.g. in `from_raw_dt_align` (or in `create_tensor` before dispatch):

```rust
let expected = tensor.as_bytes().len();
ensure!(
    content.len() == expected,
    "raw tensor data length {} does not match shape {:?} × {} = {} bytes",
    content.len(), shape, dt.size_of(), expected
);
tensor.as_bytes_mut().copy_from_slice(content);
```

## Note

On current HEAD the file originally attached to the private report no longer reaches line 502 (incidentally, due to unrelated parsing changes) — but there is no length check anywhere in the path, and the PoC above (found by a fresh fuzzing pass against HEAD) confirms the panic is fully live. This is a "looks fixed but isn't" case, hence the explicit re-verification.

*Found with Crucible, a structure-aware fuzzer for ML model parsers (Halo Forge Labs).*

czoli1976:
## Summary

Fixes #2390. `Tensor::from_raw_dt_align` allocated `∏(shape) × sizeof(dt)` bytes and then `copy_from_slice`-d the caller-supplied content into it **with no length check**. When the two lengths differ — e.g. a crafted or malformed `.onnx` whose initializer `raw_data` byte count does not match its declared `dims` — `copy_from_slice` panics, aborting model load in **both debug and release** builds. Since tract loads externally-supplied ONNX, this is a reachable DoS.

This validates the content length against the freshly-allocated tensor and returns a `TractResult` error instead of panicking. The check sits at `from_raw_dt_align`, the single choke point for all `from_raw` callers (ONNX, NNEF, …), so every raw-load path is covered.

## Reproduction (before this PR)

The 94-byte PoC from #2390 (`raw_data` of 12 bytes against a shape needing 440):

```
thread '...' panicked at data/src/tensor.rs:502:31:
copy_from_slice: source slice length (12) does not match destination slice length (440)
```

After this PR the same file loads to a graceful `Err` (no panic), verified in release.

## Change

```rust
let mut tensor = unsafe { Tensor::uninitialized_aligned_dt(dt, shape, align) }?;
let expected = tensor.as_bytes().len();
ensure!(
    content.len() == expected,
    "Raw tensor data length ({}) does not match shape {:?} of {:?} ({} bytes)",
    content.len(), shape, dt, expected
);
tensor.as_bytes_mut().copy_from_slice(content);
```

## Tests

- New regression test `from_raw_rejects_length_mismatch` in `data` (short, long, and exact-match cases).
- `cargo test -p tract-data --lib` (140) and `cargo test -p tract-onnx --lib` (10) pass; the end-to-end PoC load now returns `Err` instead of panicking.

### DIFF
--- a/data/src/tensor.rs
+++ b/data/src/tensor.rs
@@ -499,6 +499,15 @@ impl Tensor {
         align: usize,
     ) -> TractResult<Tensor> {
         let mut tensor = unsafe { Tensor::uninitialized_aligned_dt(dt, shape, align) }?;
+        let expected = tensor.as_bytes().len();
+        ensure!(
+            content.len() == expected,
+            "Raw tensor data length ({}) does not match shape {:?} of {:?} ({} bytes)",
+            content.len(),
+            shape,
+            dt,
+            expected
+        );
         tensor.as_bytes_mut().copy_from_slice(content);
         Ok(tensor)
     }
@@ -1848,6 +1857,20 @@ mod tests {
     use litteral::tensor0;
     use proptest::collection::vec;
     use proptest::prelude::*;
+
+    // Regression for sonos/tract#2390: from_raw must reject a content length that
+    // does not match the declared shape rather than panicking in copy_from_slice.
+    #[test]
+    fn from_raw_rejects_length_mismatch() {
+        // shape [2, 3] of f32 needs 24 bytes; supply 12.
+        let err = unsafe { Tensor::from_raw_dt(f32::datum_type(), &[2, 3], &[0u8; 12]) }
+            .expect_err("from_raw must reject a short content buffer, not panic");
+        assert!(err.to_string().contains("does not match shape"), "unexpected error: {err}");
+        // Too-long content is rejected as well.
+        assert!(unsafe { Tensor::from_raw_dt(f32::datum_type(), &[2, 3], &[0u8; 32]) }.is_err());
+        // Exact match still succeeds.
+        assert!(unsafe { Tensor::from_raw_dt(f32::datum_type(), &[2, 3], &[0u8; 24]) }.is_ok());
+    }
 
     #[derive(Debug)]
     struct PermuteAxisProblem {


czoli1976:
@kali

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2418, opened 2026-06-28
doc: add a profile-guided optimization build recipe

czoli1976:
Adds `doc/pgo.md`: a short recipe for building `tract-cli` with PGO on top of the existing LTO release profile, plus guidance on where it does and does not help.

The win concentrates in tract's engine glue (plan executor, op dispatch, tensor bookkeeping) — the hand-written linalg asm kernels are not instrumented by PGO, so kernel-bound graphs barely move. Measured on arm64 (M-series), release `lto=true, opt-level=2` baseline, basket-profiled build, median of paired interleaved runs:

| model | speedup |
|---|---|
| mobilenetv2 | +5.5–7.0% |
| matmul 16×576×1024 (skinny M=16) | +6.0% |
| DFN3 enc | +1.5–2.6% |
| inception_v3 | +1.3–2.1% |
| yolov8n | +0.5% |
| matmul 256×256 (square, control) | 0.0% |

The square-matmul control at 0.0% confirms the deltas are real PGO layout effects, not measurement noise. A profile from a single model (mobilenet-only) beat the diverse basket by only ~0.4pp on that model, so the doc recommends profiling a representative basket and shipping one binary — the shared glue dominates, per-model specialization buys almost nothing.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -11,6 +11,7 @@ that is harder to derive from reading the source.
 * [`graph.md`](graph.md) — Graph, Node, Outlet, Fact, model pipeline
 * [`op.md`](op.md) — anatomy of an Op (`Op` / `EvalOp` / `TypedOp` / `InferenceOp`)
 * [`cli-recipe.md`](cli-recipe.md) — `tract` command-line cookbook
+* [`pgo.md`](pgo.md) — profile-guided optimization build recipe
 * [`kernel-notes.md`](kernel-notes.md) — tract-linalg kernels and debugging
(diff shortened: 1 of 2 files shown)


kali (approved):

czoli1976:
@kali would be interesting to stack BOLT on top of PGO but do not have Linux here. Do you use PGO+BOLT already for Sonos Target?

kali:
We've been sticking to regular LTO. I think we tried PGO a while ago but decided it was not worth the trouble on our use-case. And it the first time i hear about BOLT.

PR #2420, opened 2026-06-29
Faster pooling: concrete geometry at codegen + opt-in separable average pool

czoli1976:
Two pooling changes for fixed-shape models: a `codegen` lowering MaxPool/SumPool to OptMaxPool/OptSumPool with the Patch geometry pre-resolved to `Concrete` (byte-identical; symbolic shapes untouched), and an opt-in separable running-sum average-pool kernel behind `TRACT_AVGPOOL_SEPARABLE` that trades bit-exactness — it reassociates the sum, permitted by `SumPool`'s `Validation::Rounding` — for a ~2× speedup on the avg-pool nodes. Happy to split the two if you'd rather take the byte-identical geometry change on its own.

🍍

### DIFF
--- a/core/src/ops/cnn/maxpool.rs
+++ b/core/src/ops/cnn/maxpool.rs
@@ -70,6 +70,23 @@ impl TypedOp for MaxPool {
         Ok(None)
     }
 
+    /// Lower to `OptMaxPool` with the geometry pre-resolved to `Concrete` when the
+    /// input shape is fixed, so the `Patch` is built once here rather than per eval.
+    /// Symbolic shapes are left as `MaxPool`.
+    fn codegen(
+        &self,
+        model: &TypedModel,
+        node: &TypedNode,
+    ) -> TractResult<Option<TypedModelPatch>> {
+        let fact = model.outlet_fact(node.inputs[0])?;
+        if fact.shape.as_concrete().is_none() {
+            return Ok(None);
+        }
+        let mut op = self.to_optimized(&fact.shape.to_tvec())?;
+        op.geometry = op.geometry.optimize_if(fact.shape.as_concrete())?;
+        Ok(Some(TypedModelPatch::replace_single_op(model, node, &node.inputs, op)?))
+    }
+
     as_op!();
 }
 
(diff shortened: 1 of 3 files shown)


czoli1976:
@kali

kali (approved):

czoli1976:
@kali heads-up from local profiling on M-series (using this run's CLI artifact):

The separable avg-pool path is **NCHW-only** (the `w_stride == 1` guard), so the "(Inception)" mention in the doc table is worth making precise:

| Inception variant | avg-pool nodes off→on |
|---|---|
| NNEF (NCHW) | 10.0 → 5.6 ms/i (**1.78×**); MaxPool control flat 3.7 → 3.7; E2E −3.7% |
| TF `..._frozen.pb` (NHWC) | 11.4 → 11.9 ms/i (**no-op** — guard rejects, falls back) |

The kernel does exactly what it claims when eligible, and the fallback is correct — the only nit is that the doc line reads as if the TF Inception benefits, when it doesn't. Suggest wording it as "stride-1 **NCHW** pools; NCHW-imported pool-heavy nets (the NNEF Inception) see it E2E." Byte-identical codegen change looks clean.

🍍

czoli1976:
should do NHWC too ? wdyt?

kali:
if it's feasible, why not... I'm not yet decided about which NCHW or NHWC is intrinsicly superior, tract does not make a strong choice there, tries to tolerate both.

czoli1976:
ok, you have been granted another wish, will try! but will need in the bench/tests  both types I guess

czoli1976:
Followed up on the NCHW-only point: I prototyped the NHWC path and it's up as #2440 (draft, stacked on this — will rebase once this merges).

Turns out NHWC benefits *more* than NCHW, since channels are innermost so the running sums vectorize over C. On the NHWC TF Inception (M-series), avg-pool nodes go ~10.9 → ~1.84 ms/i (~5.9×, vs 1.78× for NCHW), MaxPool flat as a control, ~7.4% E2E. Matches the generic kernel within the same rounding tolerance.

So the knob needn't stay NCHW-only — happy to fold #2440 in here or keep it separate, whichever you prefer for review.

🍍

PR #2421, opened 2026-06-29
autobatch: add a cuda-batched runtime that micro-batches concurrent run() …

kali:
…calls

Independent caller threads each fire tiny per-token predictor/joiner calls; with many concurrent streams that is one serialized CPU<->GPU round-trip per token per stream. A per-model background worker drains the in-flight queue, concatenates inputs along the symbolic batch axis, runs the inner CUDA runnable once, and scatters the row-slices back to each caller. WIP parked for later.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4784,6 +4784,7 @@ dependencies = [
  "rustfft",
  "tract-core",
  "tract-gpu",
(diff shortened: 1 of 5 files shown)


kali:
superseded spike.

PR #2428, opened 2026-06-30
block_quant: quantize einsum weights at either input slot

czoli1976:
The `block_quant` transform assumed the constant weight was always EinSum input 0 (tract's canonical `gmk,nk` form), so on imported ONNX `activation @ weight` matmuls — where the weight is input 1 — it wired the block-quant tensor and the injected group axis to the wrong operands and tripped the EinSum input-rank check. This made `block_quant` unusable on essentially any imported transformer (a BERT/MiniLM encoder fails on the first attention matmul). The slot-0 assumption went unnoticed because the only coverage, the `matmul_q40` suite, constructs its weights at slot 0.

Fix: derive the weight and activation slots from the matched input and wire them accordingly. Adds three regression tests covering both ONNX orientations (`mk,kn`, batched `bmk,kn`) and the canonical `mk,nk`, each asserting numerical equivalence to `X @ Q4_0(W)`.

🍍

### DIFF
--- a/core/src/ops/matmul/de_block_quant.rs
+++ b/core/src/ops/matmul/de_block_quant.rs
@@ -62,7 +62,8 @@ fn block_quant_einsum_weights(
         } else {
             format.quant_f32(a.cast_to::<f32>()?.try_as_plain()?.as_slice::<f32>()?)?
         };
-        let name = &model.node(node.inputs[0].node).name;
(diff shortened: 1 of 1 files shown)


kali (approved):

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2431, opened 2026-06-30
linalg/block_quant: W4A8 GEMM for Q4_0 (int8-dot decode path)

czoli1976:
Follow-up to #2348: generalises the `M == 1` W4A8 GEMV to a GEMM that decodes each Q4_0 weight block once and reuses it across the `M` activation rows, writing into a caller-provided `[m, n]` buffer — the allocation-free `&mut [f32]` output you flagged on #2348. Correctness is covered by asserting the GEMM is bit-identical to the per-row GEMV, plus an approximate check against the f32-dequant matmul.

🍍


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2440, opened 2026-07-02
Separable average pool: NHWC kernel (stacked on #2420)

czoli1976:
Extends the opt-in `TRACT_AVGPOOL_SEPARABLE` separable average-pool kernel to the NHWC layout, which the NCHW-only fast path previously skipped — so TF/TFLite pool-heavy models now benefit too. Channels are innermost in NHWC, so both running-sum passes accumulate contiguous C-wide vectors.

Stacked on #2420 (reuses its knob, dispatch, and `fast_2d_separable`); the diff here shows both commits until #2420 merges, after which I'll rebase. Draft for that reason.

Correctness: `separable_nhwc_matches_generic_kernel` checks the kernel against the generic zoned path (max abs diff < 1e-4), same contract as the existing NCHW test. Not bit-identical by design — reassociates the sum, allowed by `SumPool`'s `Validation::Rounding`.

Local NHWC Inception v3 (M-series), avg-pool nodes vs the generic kernel with MaxPool as an untouched control; end-to-end is min-of-6 interleaved:

| | generic | separable NHWC |
|---|---|---|
| SumPool nodes | ~10.9 ms/i | ~1.84 ms/i |
| MaxPool (control) | ~4.1 ms/i | ~4.2 ms/i |
| E2E | ~69.9 ms/i | ~64.7 ms/i |

🍍

### DIFF
--- a/core/src/ops/cnn/sumpool.rs
+++ b/core/src/ops/cnn/sumpool.rs
@@ -8,7 +8,7 @@ crate::declare_knob!(
     TRACT_AVGPOOL_SEPARABLE,
     bool,
     false,
-    "Use the separable average-pool kernel for stride-1 NCHW pools. Not bit-identical: \
+    "Use the separable average-pool kernel for stride-1 NCHW/NHWC pools. Not bit-identical: \
      it reassociates the sum, permitted by SumPool's Validation::Rounding contract."
 );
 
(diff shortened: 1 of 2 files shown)


czoli1976:
rebased, ready

kali (approved):

PR #2445, opened 2026-07-06
MatMulNBits: report not stateless to skip eager const-eval

Reza2kn:
HIR inference eager-const-evaluated any stateless op whose inferred inputs were concrete, which made large constant MatMulNBits nodes expensive to load. Limit that eager path to small plain tensors, matching the typed model guard, so larger folds are left to the normal optimization pipeline.

🍍

### DIFF
--- a/hir/src/infer/ops.rs
+++ b/hir/src/infer/ops.rs
@@ -28,11 +28,18 @@ pub trait InferenceOp: Op {
         let (infered_inputs, infered_outputs, observed) =
             self.infer_facts(inputs, outputs, observed).context("Infering facts")?;
 
-        if self.is_stateless() && infered_inputs.iter().all(|i| i.value.is_concrete()) {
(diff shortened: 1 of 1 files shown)


kali:
Hey ! thanks for your interest in tract !

Good eye, right diagnostic, but I think we have a better fix. I plead guilty on having abused the stateless flag myself, so it's very understandable.

I think the right fix is in hir/src/infer/ops.rs, around the line 30, mimicking core/src/model/typed.rs:89, and if it works, it should fix in one place this class of bug. Care to give it a shot ?

Reza2kn:
Took another pass and moved the fix to HIR as you suggested 😁 
Now eager const-eval only runs for tiny plain tensors, matching the typed model path, instead of special-casing MatMulNBits.

kali (approved):

kali:
Thank you !

PR #2446, opened 2026-07-06
MultiBroadcastTo: don't swap with an AxisOp successor across a rank change

czoli1976:
The swap-with-AxisOp declutter guards with an izip over the input and output shapes, which silently truncates when the broadcast adds leading axes; the successor AxisOp's indices (valid on the output) are then out of range for the lower-rank input and `AxisOp::change_shape` panics inside `ShapeFact::insert_axis` while the patch is being wired. This requires equal ranks for the swap, makes the `Add` arm of `change_shape` bail on out-of-range indices like the `Rm` arm already does, and adds a regression test that panicked before the fix.

Hit on a valid torch-2.11 ONNX export (Kyutai Pocket TTS) that runs in onnxruntime.

🍍

### DIFF
--- a/core/src/ops/array/broadcast.rs
+++ b/core/src/ops/array/broadcast.rs
@@ -156,6 +156,11 @@ impl TypedOp for MultiBroadcastTo {
         for succ in &*node.outputs[0].successors {
             let succ = model.node(succ.node);
             let Some(op) = succ.op_as::<AxisOp>() else { continue };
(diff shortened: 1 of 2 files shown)


czoli1976:
@kali this helps to unblock Pocket TTS which is a good candidate for the Zoo as well

kali (approved):

PR #2447, opened 2026-07-06
Iff: broadcast the selected branch when folding a constant condition

czoli1976:
Iff's constant-condition fold shunted the selected branch as-is, but the op's output shape is the broadcast of the condition and both branches; a branch narrower than the condition (e.g. `select(cond[1,512,16], t, f[1,512,1])`) made the patch validator reject the substitution ("Trying to substitute a 1,512,16,F32 by 1,512,1,F32") and decluttering failed. Wire a MultiBroadcastTo up to the output shape when the shapes differ; regression test included (fails before the fix).

Hit on Kyutai's Pocket TTS `mimi_encoder.onnx` (valid graph, runs in onnxruntime).

🍍


czoli1976:
@kali this also helps to unblock Pocket TTS, which is a good candidate for the Zoo as well

kali (approved):

PR #2448, opened 2026-07-06
ScatterNd: mutate the destination in place instead of round-tripping through ndarray

czoli1976:
`ScatterNd::eval` converted the data tensor to an owned ndarray and back (`into_array_unchecked` copies, and the return conversion copies again), so every eval paid two full copies of the destination even when it exclusively owned the input. Mutating through a mutable array view instead makes the common KV-cache-write pattern (ScatterND into a [2,1,1000,16,64] f32 cache, one row updated per step) drop from 8.0 ms to 1.6 ms per decode step on an Apple M-series core — a 25% end-to-end improvement on a 6-layer streaming transformer decode step (15.6 → 11.6 ms), reaching parity with onnxruntime single-thread on the same graph.

Found while running Kyutai's Pocket TTS ONNX export on tract (one of three independent fixes from that investigation, together with #2446 and #2447). onnx-core suite: 4428 pass; linalg proptests, nnef-cycle, f16 and unit suites all green.

🍍

### DIFF
--- a/core/src/ops/array/scatter_nd.rs
+++ b/core/src/ops/array/scatter_nd.rs
@@ -49,11 +49,11 @@ impl Op for ScatterNd {
 
 impl ScatterNd {
     unsafe fn eval_t<T: Datum>(
-        data: TValue,
+        data: &mut Tensor,
         indices: &ArrayViewD<i64>,
-        updates: TValue,
-    ) -> TractResult<TValue> {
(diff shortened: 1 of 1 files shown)


czoli1976:
@kali this a possible optimisation found while unblocking  Pocket TTS

kali (approved):

PR #2453, opened 2026-07-06
cpu decode: route f16 SDPA through FlashSdpa, fuse GQA broadcast through cache-read cast/rope

czoli1976:
Routes f16-accumulator SDPA through `FlashSdpaOp` on CPU and lets the GQA broadcast fusion look through cache-read `Cast`/`ApplyRope` ops, so torch-style f16 GQA exports stop paying a materialized head broadcast plus full-KV-cache repacking on every decode step.

Profiling an OpenELM-270M q40ef16 decode step at P=1000 showed the real GEMV work (~14 ms) buried under per-token recompute over unchanged data: 13.1 ms `OptMatMulPack` re-packing the KV cache, 8.2 ms `MultiBroadcastTo` materializing the repeat-interleave, 4.1 ms `ApplyRope` re-roping the whole window. Two routing gaps kept the existing flash path out of reach: the codegen gate only admitted f32 accumulators (the kernel itself takes any float and accumulates in f32), and the fusion matcher required the `DynKeyValueCache` directly under the repeat chain, which torch exports interpose with cache-read casts and rope.

Measured on Apple M-series, 1 thread, interleaved A/B against main via `llm_decode_bench` (medians):

| context | main ms/tok | this PR | speedup |
|---|---|---|---|
| 100 | 15.1 | 13.9 | 1.09× |
| 1000 | 43.0 | 30.5 | 1.41× |
| 2000 | 75.3 | 48.1 | 1.57× |

`llm-bench`: PP512 262 → 338 tok/s (+29%). One trade-off to be aware of: TG128 is −3% (75.1 → 72.7 tok/s) — at very shallow context the decomposed path still competes, and `TRACT_FLASH_SDPA_MIN_SEQ_LEN` can't gate it because kv_len is symbolic at codegen time. A runtime-side dispatch could recover it if you think it's worth the complexity.

Tests: f16-eval-vs-f32 parity on the op, and a model-level test replicating the torch GQA idiom (cache → cast → rope → repeat-interleave → SDPA) asserting the broadcasts fuse away and the fused model matches the unfused reference through real cache state. transformers/onnx-core/core/unit/linalg suites green.

Follow-ups I'd like to explore next, in order: folding the f16→f32 cast per-block into the flash streaming loop, folding rope-at-read into the kernel, and letting `InPlaceKvSdpa` match through the same cast/rope chain (which also removes the concat copy).

🍍

### DIFF
--- a/transformers/src/ops/flash_sdpa.rs
+++ b/transformers/src/ops/flash_sdpa.rs
@@ -404,6 +404,36 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn flash_sdpa_eval_f16_matches_f32() -> TractResult<()> {
+        let (b, hq, hkv, sq, sk, d) = (1usize, 4, 2, 1, 33, 8);
+        let qv = rng(b * hq * sq * d, 4);
+        let kv = rng(b * hkv * sk * d, 5);
+        let vv = rng(b * hkv * sk * d, 6);
+        let q = Tensor::from_shape(&[b, hq, sq, d], &qv)?;
+        let k = Tensor::from_shape(&[b, hkv, sk, d], &kv)?;
+        let v = Tensor::from_shape(&[b, hkv, sk, d], &vv)?;
+        let op = FlashSdpaOp { causal: true, scale: None };
+        let f32_out = op
+            .eval(tvec!(q.clone().into_tvalue(), k.clone().into_tvalue(), v.clone().into_tvalue()))?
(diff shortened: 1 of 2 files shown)


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2458, opened 2026-07-08
transformers: quantized KV-cache storage op (int8/int4)

czoli1976:
Adds `QuantizedDynKeyValueCache`, a drop-in for `DynKeyValueCache` that stores the resident KV cache in int8/int4 and dequantizes on read — cutting decode KV memory (~2× int8, ~4× int4 vs an f16 cache) while leaving RoPE/mask/attention untouched. It's opt-in via the `quantize_kv_storage` / `quantize_kv_storage_int4` transforms, which walk each `Sdpa`'s K/V inputs back through the cache-read plumbing (broadcast/reshape/cast/on-read RoPE) and swap those caches; because it stays a cache-shaped op, the past-length symbol still resolves.

KIVI layout: Keys per channel (quantized in fixed-size blocks with frozen scales for streaming consistency), Values per token. The first few tokens are kept in f32 (attention sinks carry outsized weight), a constant tiny cost that matters most for int4. Distinct from the `block_quant` weight-quantization path — this quantizes the activation cache, not the weights.

Validated end-to-end on OpenELM-270M (f16): the transform fires on all layers and runs, and the dequant-on-read adds only ~1–4% to a decode step (the step is FFN-dominated). Perplexity and next-token agreement over an 80-token greedy continuation:

|      | perplexity | top-1 agreement vs f16 |
|------|-----------|------------------------|
| f16  | 1.595     | —                      |
| int8 | 1.597     | 100%                   |
| int4 | 1.592     | 99%                    |

Both are perplexity-lossless; int4's single top-1 disagreement is a statistical tie (the reference distribution puts the two candidates at p=0.019 vs 0.018). Unit tests cover accumulation vs a plain concat, block-wise per-channel consistency over a growing range, and NNEF round-trip.

🍍

### DIFF
--- a/transformers/src/lib.rs
+++ b/transformers/src/lib.rs
@@ -19,6 +19,14 @@ register_simple_model_transform!(
     ops::inplace_kv_cache::InPlaceKvSdpaTransform
 );
 register_simple_model_transform!("transformers_detect_all", TransformersTransform);
+register_simple_model_transform!(
+    "quantize_kv_storage",
+    ops::quant_dyn_kv_cache::QuantizeKvStorageTransform { bits: 8 }
+);
+register_simple_model_transform!(
+    "quantize_kv_storage_int4",
+    ops::quant_dyn_kv_cache::QuantizeKvStorageTransform { bits: 4 }
+);
 
 pub fn register(registry: &mut Registry) {
     ops::apply_rope::register(registry);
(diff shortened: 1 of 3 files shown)


czoli1976:
!?!

kali:
we just need a rebase :)

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2462, opened 2026-07-10
bench: let append-dinghy run per-leg so one stuck board doesn't drop …

kali:
…the fleet's data

A single dinghy leg timing out failed the bench-dinghy matrix, which skipped append-dinghy entirely and discarded the results of the boards that did finish. Gate the job on always() and each leg on its own result artifact.


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2469, opened 2026-07-13
linalg/mmm: naive tile walk when L2 is undetectable

kali:
The single-thread block budget fell back to a 256 KiB guess when the cache probe found nothing, which over-blocks the very cores that fallback is for: small embedded arm parts whose cache the OS does not expose and whose tiny L1 an L2-sized block blows, making the blocked walk slower than the naive stream the hardware prefetcher handled well. The budget helpers now return None on an unknown cache and the walk takes the naive loop (edge 1) there, so only cores with a detected L2 get the blocking they were tuned for.

### DIFF
--- a/linalg/src/frame/mmm/mod.rs
+++ b/linalg/src/frame/mmm/mod.rs
@@ -328,20 +328,19 @@ const ST_BLK_L3_MAX: usize = 64;
 
 /// Panel-block working-set budget (bytes) from a detected cache size: a fraction
 /// `num/den` of the cache (leaving room for the C accumulator tile + packing
-/// metadata), clamped to a sane range. `0` (cache unknown) ⇒ `fallback`, which
-/// is kept small so the block ≈ the naive loop and can never over-block a cache
-/// it can't see. Sizes come from the shared [`crate::cache`] probe.
-fn tier_budget_bytes(cache_bytes: usize, num: usize, den: usize, fallback: usize) -> usize {
-    if cache_bytes == 0 {
-        fallback
-    } else {
-        (cache_bytes * num / den).clamp(64 * 1024, 64 * 1024 * 1024)
-    }
+/// metadata), clamped to a sane range. `None` when the cache size is unknown:
+/// there is no size that is safe to assume, since the cores whose size the OS
+/// hides are the small in-order embedded parts a guessed budget would over-block
+/// (an L2-sized block blows their L1). Callers fall back to the naive loop.
+/// Sizes come from the shared [`crate::cache`] probe.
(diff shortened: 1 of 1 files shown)


kali:
ping @czoli1976

kali:
Mmmm tricky one. I was chasing the inception regression, and I broke everything else.

czoli1976:
At which point did you detect the regression? which PR? is the eval time the concern or wha?

czoli1976:
The tricky part is that "undetectable L2" isn't one workload — the little cores split on what they want:

- **mobilenet_v1/v2, en_tdnn_\*** regress because they *were* getting real reuse from the 256 KiB guess — which roughly matches the actual (just-unreported) L2 on a53/a55/a7.
- **inception / kws / hey_snips** improve because 256 KiB over-blocks their tiny L1.

So `256 KiB` and `edge 1` are both extremes, and the sweet spot is in between. Your own diagnosis points right at it: the failure mode is "an L2-sized block blows L1", so budget from **L1** rather than dropping the tile entirely. `CacheInfo::l1_data_or_default()` already hands back 64 KiB on arm64 when L1 is hidden:

```rust
fn l2_block_budget_bytes() -> Option<usize> {
    let ci = crate::cache::cache_info();
    // detected L2 → tuned budget; hidden → block to L1, not an L2-sized guess
    // that blows L1, and not naive which strips the reuse the GEMM-bound models need.
    Some(tier_budget_bytes(ci.l2, 1, 3).unwrap_or_else(|| ci.l1_data_or_default()))
}
```

That keeps a real but small tile — shouldn't blow L1, should hand mobilenet/tdnn back their reuse without re-inflating inception. The L1 fraction is a knob to tune. Pure hypothesis on my end though: I've got no a7/a53/a55 to bench on, so it'd need a run through the bot to confirm it lands in the middle instead of just tilting the see-saw the other way.

czoli1976:
My imaginary friend seems confident here

kali:
I'm trying to recover performance that slipped on long term time series that i was not monitoring very closely (e.g. inceptionv3 on intel). I think these regressions started before you started working on tract, and I'm trying to address a wider bag of performance waste with #2478 . It looks like a simple linear model can equal the choices of the DNN on the cortex devices, so I'm looking into generalizing the approach. I'll come back to the cache once I'm convinced we're not picking stupid kernels to start with.

czoli1976:
?

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali:
superseded by #2501

PR #2477, opened 2026-07-15
core: don't fuse AddUnicast into a native block-quant matmul

czoli1976:
`OptMatMul::fuse` folds a following same-shape `Add` into the matmul as a fused `AddUnicast`. When the matmul consumes a block-quant weight natively (no panel extractor), that chain — `AddMatMul → AddUnicast → Store` in one kernel call — miscomputes and saturates the output to `inf`, even with finite, contiguous inputs. This gates the AddUnicast fast-path off for native block-quant matmuls, leaving the residual add as its own (correct) node.

Found on `Qwen2.5-7B-Instruct-q40ef16`: `PushSliceUp` produces a sliced `n=1` GEMV over the last token at the tail, the residual add fuses into it (`arm64simd_mmm_f32_32x1_gen`, q40 packing), and the last block's logits go NaN. With the fusion disabled the whole model decodes correctly (token-identical to Metal); llama3.2-1b / Qwen3-1.7b / openelm are unaffected.

The kernels are individually covered — `AddMatMul` alone (`packed_packed`) and `AddUnicast` after `Clear` (`add_unicast_non_contiguous`) — but not the two chained, which is the real gap. The proper follow-up is to add the `AddMatMul(native block-quant) → AddUnicast` chain to the linalg fuse proptests and fix the kernel; this change is the safe correctness guard in the meantime.

🍍

### DIFF
--- a/core/src/ops/matmul/optimized.rs
+++ b/core/src/ops/matmul/optimized.rs
@@ -568,6 +568,13 @@ impl TypedOp for OptMatMul {
                 let other_fact = patch.outlet_fact(other_input)?;
 
                 if other_fact.shape == self.c_fact.shape {
+                    // A native block-quant AddMatMul chained with a fused AddUnicast in the
+                    // same kernel call miscomputes on some kernels (that pairing has no direct
+                    // test coverage). Leave the add as its own node; it is always correct
+                    // unfused, and this pairing only arises for a sliced quant GEMV.
+                    if self.consumes_native_block_quant() {
+                        return Ok(None);
+                    }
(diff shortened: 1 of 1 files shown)


kali:
Same question as the saturation discussion, really. Are we observing a saturation because the model can not tolerate f16 at this specific place ? If that the case, the model should tell tract so, and tract should honor it.

I hear that the combination AddMM/AddU is not unit-test covered, but I feel this is not strong enough an argument to say it's bogus. The MM kernels are interpreters, we assume that operations can combine. Or show me the bug.

kali:
Thanks for going deeper :)

czoli1976:
image

PR #2487, opened 2026-07-17
perf(linalg): drop per-iteration input copies in arm64 fused silu

cverrier:
The fused SiLU kernel kept the unclamped input for its final multiply by copying each freshly-loaded vector aside before clamping. Load the input straight into the register bank it lives in and clamp into the working bank instead, so the input stays live with no copy. This removes four movs per 16-lane iteration and one in the scalar tail, with no change to results or register usage.

Also clamp the original the original input on the negative tail so the kernel is total.

### DIFF
--- a/linalg/src/arm64/arm64simd/silu_fused.rs
+++ b/linalg/src/arm64/arm64simd/silu_fused.rs
@@ -1,21 +1,20 @@
 // Fused SiLU kernel
 //
 // Exact formula computed per element (z = clamp(x, -18.6, 18.6), w = z²):
-//   silu(x) = x * (0.5 + z * P(w) / Q(w))
-// where P is the degree-6 Horner polynomial over coeffs COEFFS[2..=8] and Q the
-// degree-3 Horner polynomial over coeffs COEFFS[9..=12].
+//   SiLU(x) = f * (0.5 + z * P(w) / Q(w))
+// where P is the degree-6 Horner polynomial over coeffs COEFFS[2..=8], Q the
+// degree-3 Horner polynomial over coeffs COEFFS[9..=12], and f = max(x, -18.6)
+// the factor the sigmoid multiplies.
 //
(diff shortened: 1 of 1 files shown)


linalg/src/arm64/arm64simd/silu_fused.rs
```diff
@@ -184,12 +191,11 @@ ew_impl_wrap!(
                 cbz {len}, 3f
 
                 2:
-                    ld1 {{ v16.4s }}, [{ptr}]
-                    mov v8.16b, v16.16b
+                    ld1 {{ v8.4s }}, [{ptr}]
```
cverrier, line 193: Given that in the unrolled loop, the raw input already lives in the callee-saved registers `v8`-`v11`, I decided to also reuse `v8` here for consistency.

linalg/src/arm64/arm64simd/silu_fused.rs
```diff
@@ -44,35 +56,30 @@ ew_impl_wrap!(
...
+                dup v5.4s, v0.s[0]             // v5 <- low, broadcasted
+                dup v6.4s, v0.s[1]             // v6 <- high, broadcasted
+                dup v7.4s, v3.s[1]             // v7 <- 0.5, broadcasted
 
                 cmp {len}, #16
                 blt 9f
 
                 1:
-                    ld1 {{ v16.4s, v17.4s, v18.4s, v19.4s }}, [{ptr}]
+                    ld1 {{ v8.4s, v9.4s, v10.4s, v11.4s }}, [{ptr}]
```
cverrier, line 66: I chose to load the raw input in `v8`-`v11` registers so it minimizes the diff, and the resulting kernel is still mirroring the sigmoid one defined in `linalg/src/arm64/arm64simd/sigmoid_f32_4n.S.j2`.

kali:
FYI, for the first bench run, we are observng the improvement that landed in yesterday. Second bench is the honest one. No impact, which is what we were expecting.

PR #2492, opened 2026-07-18
perf(linalg): add NEON sigmoid_f16 fallback on non-fp16 arm64

cverrier:
On arm64 cores without FEAT_FP16, f16 sigmoid had no SIMD kernel registered and dropped to tract's generic scalar path. It now runs on the existing NEON f32 sigmoid, giving these cores NEON-f32 throughput on f16 models. The native arm64fp16 kernel still takes precedence wherever FEAT_FP16 is present.

> [!NOTE]
> I based this work on what is done in [`linalg/src/x86_64_fma/act_f16.rs`](https://github.com/cverrier/tract/blob/a3997659bf8bcb7e6315e67cc1fc476d8a7b7e09/linalg/src/x86_64_fma/act_f16.rs), so I inlined the Assembly code in a `.rs` file rather than creating a brand new `.S.j2` file for the kernel: consequently, the f16↔f32 conversion uses an intermediate scratch buffer that should always live in L1 cache. The reason I chose this approach is that it avoids code duplication from [`linalg/arm64/arm64simd/arm64simd_sigmoid_f32_4n.S.j2`](https://github.com/cverrier/tract/blob/a3997659bf8bcb7e6315e67cc1fc476d8a7b7e09/linalg/arm64/arm64simd/arm64simd_sigmoid_f32_4n.S.j2). The price to pay is that we need an intermediate buffer in memory (L1, so high-bandwidth) to handle the conversion, rather than handling it inside registers. I thought this would be a good tradeoff, though I might be wrong. @kali if you think the conversion might be exclusively handled inside registers, I can adapt.

Add a criterion bench comparing the scalar path, the f32-roundtrip NEON fallback, and the native fp16 kernel for f16 sigmoid, across L1, L2, and DRAM-sized buffers (this is hardware-dependent, but I tried to pick numbers that would make sense on a small chip). The kernels are invoked by name so the fallback can be measured on fp16-capable hardware, where dispatch would otherwise always pick the native kernel.

Regarding the ASM unrolling in the f16↔f32 conversion loops I introduced in commit eed405aa3318acb8cdc32ccd323aa391a0c550e7 (this hash will change because I'll likely rebase onto the `main` branch, so I'm talking about the commit whose title is "_perf(linalg): unroll arm64 f16 sigmoid conversions to 32 lanes_"): This `sigmoid_f16` path is the fallback we dispatch to on ARMv8.2 cores without FEAT_FP16: we widen f16→f32, put the result in a scratch register, run the existing f32 sigmoid kernel, and narrow back. The two conversion loops are just chains of FCVTL/FCVTN, which are latency-bound (each iteration's few instructions don't fill the pipeline on their own). My hypothesis was that unrolling from 8 to 32 lanes would keep several independent convert chains in flight, hiding that latency and amortizing the loop overhead. The expected beneficiary is an in-order core (e.g., a Cortex-A55 chip), where the CPU can't reorder around the convert latency itself.

I ran the following Criterion benchmarks on a Cortex-A55 chip (frequency-locked with `performance` governor so we have the CPU running at maximum frequency during the whole experiment) using `dinghy`, to see whether this unrolling is truly beneficial (again, the commit hash I use below is likely to change, see my explanation before to find the right commit):

```sh
# Keep the deploy small
printf '/*\n' > .dinghyignore

# 1. BEFORE — parent commit, save a named baseline
git checkout eed405aa3~1
cargo dinghy -d <CORTEX_A55_DEVICE> bench -p tract-linalg --bench sigmoid_f16_arm64 \
    -- --save-baseline before

# 2. AFTER — the commit itself, compare against that baseline
git checkout eed405aa3318acb8cdc32ccd323aa391a0c550e7
cargo dinghy -d <CORTEX_A55_DEVICE> bench -p tract-linalg --bench sigmoid_f16_arm64 \
    -- --baseline before

# Cleanup
rm -f .dinghyignore
git checkout perf/add-neon-sigmoid-f16-fallback-on-non-fp16-arm64
```

Results after the comparison for the new kernel (the generic code and the native f16 kernel didn't significantly change, which is expected):

```
sigmoid_f16/neon-f32-roundtrip/1024
                        time:   [7.3282 µs 7.3302 µs 7.3321 µs]
                        thrpt:  [139.66 Melem/s 139.70 Melem/s 139.73 Melem/s]
                 change:
                        time:   [−4.7514% −4.7130% −4.6740%] (p = 0.00 < 0.05)
                        thrpt:  [+4.9031% +4.9461% +4.9884%]
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
  
sigmoid_f16/neon-f32-roundtrip/32768
                        time:   [234.32 µs 234.36 µs 234.42 µs]
                        thrpt:  [139.79 Melem/s 139.82 Melem/s 139.85 Melem/s]
                 change:
                        time:   [−4.4795% −4.4324% −4.3948%] (p = 0.00 < 0.05)
                        thrpt:  [+4.5968% +4.6380% +4.6896%]
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
  1 (1.00%) high severe

sigmoid_f16/neon-f32-roundtrip/1048576
                        time:   [7.5719 ms 7.5744 ms 7.5768 ms]
                        thrpt:  [138.39 Melem/s 138.44 Melem/s 138.48 Melem/s]
                 change:
                        time:   [−5.1631% −5.1122% −5.0641%] (p = 0.00 < 0.05)
                        thrpt:  [+5.3343% +5.3876% +5.4442%]
                        Performance has improved.
```

The medians from each `time:`/`change:` triple are reported below for clarity:

| Benchmark (`neon-f32-roundtrip`) | Time (median) | Throughput | Δ Time | Δ Throughput | Verdict |
|---|---|---|---|---|---|
| `sigmoid_f16/.../1024` | 7.3302 µs | 139.70 Melem/s | −4.71% | +4.95% | improved (p<0.05) |
| `sigmoid_f16/.../32768` | 234.36 µs | 139.82 Melem/s | −4.43% | +4.64% | improved (p<0.05) |
| `sigmoid_f16/.../1048576` | 7.5744 ms | 138.44 Melem/s | −5.11% | +5.39% | improved (p<0.05) |

This shows a significant improvement, about 5%.

I also ran that same commands (without `dinghy`) on a laptop equipped with an Apple M3 Pro chip, but I couldn't observe a significant improvement: I guess this might be due to general noise, and also due to the fact that it's an out-of-order chip, so the loop unrolling for the conversions isn't really necessary.

### DIFF
--- a/linalg/Cargo.toml
+++ b/linalg/Cargo.toml
@@ -104,6 +104,10 @@ name = "sigmoid"
 harness = false
 
 [[bench]]
+name = "sigmoid_f16_arm64"
+harness = false
+
+[[bench]]
 name = "softmax"
 harness = false
 
--- a/linalg/benches/sigmoid_f16_arm64.rs
+++ b/linalg/benches/sigmoid_f16_arm64.rs
@@ -35,7 +35,7 @@ fn sigmoid_f16(c: &mut Criterion) {
         let mut tf = aligned_input(n);
         let sf = unsafe { tf.as_slice_mut_unchecked::<f16>() };
         group.bench_with_input(BenchmarkId::new("neon-f32-roundtrip", n), &(), |b, _| {
-            b.iter(|| tract_linalg::arm64::arm64simd_sigmoid_f16_32n::run(sf, ()))
+            b.iter(|| tract_linalg::arm64::arm64simd_sigmoid_f16_4n::run(sf, ()))
         });
 
         if tract_linalg::arm64::has_fp16() {
--- a/linalg/benches/sigmoid_f16_arm64.rs
+++ b/linalg/benches/sigmoid_f16_arm64.rs
@@ -1,0 +1,53 @@
+// Microbenchmark: arm64 f16 sigmoid across the cache hierarchy.
+// The three sizes span working sets from cache-resident (compute-bound) to
+// larger-than-cache (bandwidth-bound), so the generic-vs-roundtrip comparison
+// holds in both regimes. Exact cache tiers are hardware-specific — these are
+// representative, not calibrated to a particular core.
+// Kernels are called by name to bypass dispatch, which always selects the
+// native fp16 kernel on an fp16-capable core. The f32-roundtrip fallback only
+// uses baseline FCVTL/FCVTN, so it is safe to call on any arm64.
+#![cfg(target_arch = "aarch64")]
+
+use criterion::*;
+use tract_data::prelude::*;
+use tract_linalg::element_wise::ElementWiseKer;
+
+fn aligned_input(n: usize) -> Tensor {
+    let mut t = unsafe { Tensor::uninitialized_aligned::<f16>(&[n], 16).unwrap() };
+    let s = unsafe { t.as_slice_mut_unchecked::<f16>() };
+    for (i, x) in s.iter_mut().enumerate() {
+        *x = f16::from_f32((i as f32 / 10.0).sin() * 5.0);
+    }
+    t
+}
+
+fn sigmoid_f16(c: &mut Criterion) {
+    for n in [1024usize, 32_768, 1_048_576] {
+        let mut group = c.benchmark_group("sigmoid_f16");
+        group.throughput(Throughput::Elements(n as u64));
+
+        let mut tg = aligned_input(n);
+        let sg = unsafe { tg.as_slice_mut_unchecked::<f16>() };
+        group.bench_with_input(BenchmarkId::new("generic", n), &(), |b, _| {
+            b.iter(|| tract_linalg::generic::sigmoid::HSigmoid8::run(sg, ()))
+        });
+
+        let mut tf = aligned_input(n);
+        let sf = unsafe { tf.as_slice_mut_unchecked::<f16>() };
+        group.bench_with_input(BenchmarkId::new("neon-f32-roundtrip", n), &(), |b, _| {
+            b.iter(|| tract_linalg::arm64::arm64simd_sigmoid_f16_8n::run(sf, ()))
+        });
+
+        if tract_linalg::arm64::has_fp16() {
+            let mut tn = aligned_input(n);
+            let sn = unsafe { tn.as_slice_mut_unchecked::<f16>() };
+            group.bench_with_input(BenchmarkId::new("native-fp16", n), &(), |b, _| {
+                b.iter(|| tract_linalg::arm64::arm64fp16_sigmoid_f16_8n::run(sn, ()))
+            });
+        }
+        group.finish();
+    }
+}
+
+criterion_group!(g, sigmoid_f16);
+criterion_main!(g);
--- a/linalg/benches/sigmoid_f16_arm64.rs
+++ b/linalg/benches/sigmoid_f16_arm64.rs
@@ -35,7 +35,7 @@ fn sigmoid_f16(c: &mut Criterion) {
         let mut tf = aligned_input(n);
         let sf = unsafe { tf.as_slice_mut_unchecked::<f16>() };
         group.bench_with_input(BenchmarkId::new("neon-f32-roundtrip", n), &(), |b, _| {
-            b.iter(|| tract_linalg::arm64::arm64simd_sigmoid_f16_8n::run(sf, ()))
+            b.iter(|| tract_linalg::arm64::arm64simd_sigmoid_f16_32n::run(sf, ()))
         });
 
         if tract_linalg::arm64::has_fp16() {
--- a/linalg/src/arm64.rs
+++ b/linalg/src/arm64.rs
@@ -509,7 +509,7 @@ pub fn plug(ops: &mut Ops) {
         ops.mul_by_scalar_f16 = Box::new(|| arm64fp16_mul_by_scalar_f16_32n::ew());
     } else {
         log::info!("No native fp16 support; f32-roundtrip NEON sigmoid_f16 activated");
-        ops.sigmoid_f16 = Box::new(|| arm64simd_sigmoid_f16_32n::ew());
+        ops.sigmoid_f16 = Box::new(|| arm64simd_sigmoid_f16_4n::ew());
     }
     #[cfg(any(target_os = "macos", all(target_os = "ios", feature = "apple-amx-ios")))]
     {
--- a/linalg/src/arm64.rs
+++ b/linalg/src/arm64.rs
@@ -509,7 +509,7 @@ pub fn plug(ops: &mut Ops) {
         ops.mul_by_scalar_f16 = Box::new(|| arm64fp16_mul_by_scalar_f16_32n::ew());
     } else {
         log::info!("No native fp16 support; f32-roundtrip NEON sigmoid_f16 activated");
-        ops.sigmoid_f16 = Box::new(|| arm64simd_sigmoid_f16_8n::ew());
+        ops.sigmoid_f16 = Box::new(|| arm64simd_sigmoid_f16_32n::ew());
     }
     #[cfg(any(target_os = "macos", all(target_os = "ios", feature = "apple-amx-ios")))]
     {
--- a/linalg/src/arm64.rs
+++ b/linalg/src/arm64.rs
@@ -508,7 +508,8 @@ pub fn plug(ops: &mut Ops) {
         ops.sum_f16 = Box::new(|| arm64fp16_sum_f16_32n::red());
         ops.mul_by_scalar_f16 = Box::new(|| arm64fp16_mul_by_scalar_f16_32n::ew());
     } else {
-        log::info!("No native fp16 support");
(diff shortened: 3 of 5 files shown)


kali:
It's an intresting compromise, I would be curious to see how it compares with:
* the presumably "best-in-class" option: in-kernel conversions
* tract-core sigmoid implementaiton (implement the sigmoid closure by cast_to(f32), sigmoid, cast_to(f16))
* tract-core multi op decluttering (replace during the "codegen" phase a f16 sigmoid by three op cast, sigmoid(f32), cast back)

Another interesting option would be to see if it would make sense to chunk the buffer in llinalg, loop overs chunks doing cvt->sigmoid->cvt, to avoid cache spills.

cverrier:
> It's an intresting compromise, I would be curious to see how it compares with:
> 
> * the presumably "best-in-class" option: in-kernel conversions
> * tract-core sigmoid implementaiton (implement the sigmoid closure by cast_to(f32), sigmoid, cast_to(f16))
> * tract-core multi op decluttering (replace during the "codegen" phase a f16 sigmoid by three op cast, sigmoid(f32), cast back)

On it, I'll run these experiments in the experimental branch: [`cverrier/tract/tree/exp/sigmoid-f16-approaches-pr-2492`](https://github.com/cverrier/tract/tree/exp/sigmoid-f16-approaches-pr-2492).

> Another interesting option would be to see if it would make sense to chunk the buffer in llinalg, loop overs chunks doing cvt->sigmoid->cvt, to avoid cache spills.

I think this is exactly the approach I've already implemented in this PR, right? See

https://github.com/cverrier/tract/blob/eed405aa3318acb8cdc32ccd323aa391a0c550e7/linalg/src/arm64/arm64simd/act_f16.rs#L123-L132

Or do you mean something else?

kali:
> 
> On it, I'll run these experiments in the experimental branch: `cverrier/tract/tree/exp/sigmoid-f16-approaches-pr-2492`.
> 
> > Another interesting option would be to see if it would make sense to chunk the buffer in llinalg, loop overs chunks doing cvt->sigmoid->cvt, to avoid cache spills.
> 
> I think this is exactly the approach I've already implemented in this PR, right? See
> 
> https://github.com/cverrier/tract/blob/eed405aa3318acb8cdc32ccd323aa391a0c550e7/linalg/src/arm64/arm64simd/act_f16.rs#L123-L132
> 
> Or do you mean something else?

Right, I had missed it. 256 may be on the small size... it means a 1kB buffer, I think we can safely count on 16kB on all configurations...

cverrier:
> 256 may be on the small size... it means a 1kB buffer, I think we can safely count on 16kB on all configurations...

Regarding this: I ran some experiments that vary the chunk size (up to 4096, to approximately reach your ~16 kB suggestion). And increasing the chunk size seems to introduce regressions. I initially thought that filling the L1 cache level as much as we could was the right thing to do, but it's not, because this also increases the reuse distance between writes and reads:
1. `cvt_f16_to_f32` writes the scratch buffer (pass 1),
2. the f32 sigmoid kernel reads it and writes it (pass 2),
3. `cvt_f32_to_f16` reads it again (pass 3).

Since the L1 cache will also contain the input buffer's cache lines, plus whatever is needed, (at least) the first values in the scratch buffer written at the end of pass 1 might have been evicted from L1 when pass 2 comes back to read the scratch buffer, and so now we have increased the latency to retrieve that data.

So as far as I understand, the scratch buffer size `CHUNK` only sets how far apart the scratch's writes and re-reads are, and staying close enough to keep those re-reads in the hottest cache level (not merely "fitting in L1") is the entire win.

Apart from that, I've also:
- Lowered the roundtrip kernel's `nr` from 32 to 4 so f16 sigmoid on arm64 without native fp16 accepts any length that is a multiple of 4, leaving far less of the tensor on the slow scalar path.
- Changed the `cvt_` functions that handle the f16<->f32 conversions so they handle arbitrary lengths entirely in NEON asm.

I re-ran benchmarks to compare this new implementation (commit bfdb58ab7cf28121d074a0000f815fa5d8c1bd66, titled "perf(linalg): tune arm64 non-fp16 f16 sigmoid fallback" in case we rebase) to the previous one (commit eed405aa3318acb8cdc32ccd323aa391a0c550e7, titled "perf(linalg): unroll arm64 f16 sigmoid conversions to 32 lanes") on a Cortex-A55 CPU: no regression.

Regarding the questions you had about comparing this new implementation to others: I'm done, though I first have to re-run the comparisons with the new implementation that landed on this PR. I'll keep you posted soon.

cverrier:
@kali I got the results of [the experiments you asked](https://github.com/sonos/tract/pull/2492#issuecomment-5019791508) for, see [the experimental branch](https://github.com/cverrier/tract/tree/exp/sigmoid-f16-approaches-pr-2492) for the full code.

### Setting

| ID | Name | Altitude | Measured via |
|----|------|----------|--------------|
| **A** | `neon-f32-roundtrip` | linalg kernel | already exists (`arm64simd_sigmoid_f16_4n`) |
| **B** | `neon-f32-fused` | linalg kernel | new fused kernel (`arm64simd_sigmoid_f16_4n_fused`): convert→sigmoid→convert in registers, one pass, no scratch |
| **C** | `core-cast-roundtrip` | tract-core closure | proxy: `cast_to::<f32>` → f32 kernel → `cast_to::<f16>` |
| **D** | `codegen-3op` | tract-core graph | proxy: `source(f16)→Cast(f32)→sigmoid→Cast(f16)`, optimized+run |

### Results

Measured on a Cortex-A55 (in-order) CPU, locked 1908 MHz (`performance` governor), via dinghy. Throughput in Melem/s, higher is better.

> [!NOTE]
> This Cortex-A55 has FEAT_FP16, so model-level `one-op` dispatches to the native fp16 kernel, *i.e.* it is not representative. The fallback kernels A/B/C are measured directly by name (bypassing dispatch), so they faithfully reflect the non-fp16 fallback cost.

Kernel-level (`sigmoid_f16_arm64`):

| Input buffer size | Generic | A | B | C | Native-fp16 |
|------|---------|-------------|---------|-------------|-------------|
| 1024 (L1)      | 15.09 | 140.89 | **159.89** | 54.17 | 664.54 |
| 32768 (L2)     | 15.08 | 140.30 | **159.31** | 57.63 | 656.09 |
| 1048576 (DRAM) | 15.03 | 137.83 | **157.86** | 57.53 | 655.77 |

Model-level (`sigmoid_f16_model`):

| Input buffer size | one-op (native on this chip) | D |
|------|-----------------------------|---------------|
| 1024 (L1)      | 201.61 | 43.95 |
| 32768 (L2)     | 493.57 | 56.87 |
| 1048576 (DRAM) | 459.41 | 57.27 |

Reproduce (A55 frequency-locked, `performance` governor):

```sh
git checkout exp/sigmoid-f16-approaches-pr-2492

printf '/*\n' > .dinghyignore   # keep the dinghy deploy small

# 1) kernel-level: A vs B vs C vs native vs generic (one group, read the medians)
cargo dinghy -d <CORTEX_A55_DEVICE> bench -p tract-linalg --bench sigmoid_f16_arm64

# 2) model-level: D-proxy (3-op) vs one-op
cargo dinghy -d <CORTEX_A55_DEVICE> bench -p tract-core --bench sigmoid_f16_model

rm -f .dinghyignore
```

So in conclusion, B (the in-kernel conversions) is the fastest, as expected, but A (the "neon-f32-roundtrip" that is implemented in this PR) is much faster than C and D.

kali:
OK, so we're wasting 15% or so by going through conversion to cache, but it's a pretty good compromise. We can always go after the 15% remaioning at a later stage if we need them.

I think we want to push this prototype the extra mile: 
* define converters for arm7 and avx2
* make the deriving of a f16 ew from a f32 ew a one line macro (or function, or closure)

cverrier:
> I think we want to push this prototype the extra mile:
> 
> * define converters for arm7 and avx2
> * make the deriving of a f16 ew from a f32 ew a one line macro (or function, or closure)

I can do both. Should we address these next steps in future separate PRs so we keep this one tight and focused on the sigmoid f32 fallback?

kali:
your choice !

cverrier:
> your choice !

Then if you are satisfied with this work, let's merge it and I will open separate PRs soon for the rest :)

kali (approved):

cverrier:
@kali I just saw that all the commits that belong to this branch ended into `main`. I think we want to have these commits squashed into a single one in `main`. I usually don't squash commits on an open pull request that has been reviewed at least once to avoid confusing the reviewers (and also to preserve the commit history for future reference), but I think it's possible to do a "Squash and merge" when merging into `main`. Or do you want to have these intermediate commits visible in `main` for any specific reason?

kali:
My policy is to merge by rebase by default, no squashing. There is value in the intermediate steps, and the project is not so big it is a problem.

PR #2493, opened 2026-07-18
linalg: probe the SVE f16 -march spelling instead of hardcoding +fp16

### LINKED ISSUE #2483: linalg: aarch64 SVE f16 kernels fail with zig cc / cargo-zigbuild — unknown CPU feature 'fp16' (spelling: fp16 vs fullfp16)
## Summary

Cross-compiling a project that depends on tract to `aarch64-unknown-linux-musl` with [cargo-zigbuild](https://github.com/rust-cross/cargo-zigbuild) (a common setup for producing static aarch64 Rust binaries from an x86 CI runner) fails building `tract-linalg`:

```
error: unknown CPU feature: 'fp16'

command did not execute successfully (status code exit status: 1):
".../zigcc-aarch64-unknown-linux-musl-....sh" ... "-march=armv8.2-a+sve+fp16"
  -o ".../out/...-sve_mmv_f16_64x1.o" -c "arm64/sve/sve_mmv_f16_64x1.c"
```

Seen with tract-linalg 0.23.4; the flag is still present on `main` (`linalg/build.rs`, the SVE f16 kernel block):

```rust
.flag("-march=armv8.2-a+sve+fp16")
```

## Root cause

`+fp16` is the GCC/clang arch-extension spelling; the LLVM target-feature name is `fullfp16`. GCC accepts only `fp16`, clang accepts both (its AArch64 parser maps `fp16` → `fullfp16`), but `zig cc` — which cargo-zigbuild uses as the cross C compiler — accepts only the LLVM name. The exact same source compiles cleanly through zig with `-march=armv8.2-a+sve+fullfp16`, so it's purely a spelling issue:

```console
$ cargo-zigbuild zig cc -- -target aarch64-linux-musl -march=armv8.2-a+sve+fp16 -c sve_f16_probe.c
error: unknown CPU feature: 'fp16'
$ cargo-zigbuild zig cc -- -target aarch64-linux-musl -march=armv8.2-a+sve+fullfp16 -c sve_f16_probe.c
(succeeds)
```

I've filed the complementary issue on cargo-zigbuild to alias the spelling on their side too: rust-cross/cargo-zigbuild#456. Fixing either side unblocks the ecosystem; fixing both makes each robust.

## Suggested fix

`build.rs` already probes the toolchain for SVE support before enabling those kernels (`compiler_supports_sve()`). The f16 kernel block could pick the `-march` spelling the same way: try `-march=armv8.2-a+sve+fp16`, fall back to `-march=armv8.2-a+sve+fullfp16`, and use whichever compiles. That keeps GCC (which only knows `fp16`) working and fixes zig.

(If neither compiles, note that skipping only the f16 kernels isn't currently safe — `src/arm64/sve.rs` declares `sve_mmm_f16_kernel`/`sve_mmv_f16_64x1_kernel` under the same `tract_sve` cfg as the f32 kernels, so it would need a separate cfg or falling back to no SVE at all to avoid link errors.)

## Environment

- tract-linalg 0.23.4 (crates.io), reproduced against `main`
- cargo-zigbuild 0.23.0, ziglang 0.16.0 (pip), rustc 1.95, host x86_64-linux
- target: aarch64-unknown-linux-musl

Happy to send a PR for the spelling probe if that approach sounds right.

thevoiceguy:
Fixes #2483.

## Problem

`linalg/build.rs` compiles the SVE f16 kernels with a hardcoded `-march=armv8.2-a+sve+fp16`. GCC's aarch64 backend spells native FP16 `+fp16`, but LLVM/clang — and therefore `zig cc`, which `cargo-zigbuild` uses to cross-compile the `*-musl` targets — only accepts `+fullfp16`. So `cargo zigbuild --target aarch64-unknown-linux-musl -p tract-linalg` fails with:

```
error: unknown CPU feature: 'fp16'
```

even though the base `+sve` f32/i32 kernels compile fine (the failure is specifically the f16 kernel `cc` invocation).

## Fix

Per your suggestion on #2483:

1. **Probe the spelling** — a new `compiler_sve_fp16_flag()` trial-compiles a tiny `arm_sve.h` TU against `-march=armv8.2-a+sve+fp16` then `-march=armv8.2-a+sve+fullfp16`, returning the first the compiler accepts (mirroring the existing `compiler_supports_sve` / assembler probes). The `-march` string is rejected during the compiler's target parse, before codegen, so a base-SVE probe source is enough to discriminate an accepted spelling.
2. **Split the cfg** — `tract_sve` still gates the f32/i32 SVE kernels; a new `tract_sve_fp16` gates the f16 kernels and is set only when an accepted spelling is found. The f16 externs, `MMMRustKernel!` defs, `SVE2_FP16`, the `f16` imports, and the `plug()` dispatch move to `tract_sve_fp16`. If no spelling works, the f16 kernels are skipped and the f32/i32 SVE kernels still build instead of the whole crate failing.

Native aarch64 (GCC) is unaffected: `+fp16` is tried first and accepted, so codegen and emitted cfgs are identical to before.

## Testing

On an x86_64 host with `zig` + `cargo-zigbuild`, target `aarch64-unknown-linux-musl`:

- **Before:** reproduced `error: unknown CPU feature: 'fp16'`.
- **After:** builds clean; build script emits both `cargo:rustc-cfg=tract_sve` and `cargo:rustc-cfg=tract_sve_fp16`, and both `libtract_sve_kernels.a` and `libtract_sve_f16_kernels.a` are compiled and linked (the fallback selects `+fullfp16`).
- **Degradation path:** with the fp16 probe forced to return `None`, the crate still builds — `tract_sve` set, `tract_sve_fp16` unset, f16 archive absent, no unresolved symbols.
- **Native/host:** `cargo build -p tract-linalg` (x86_64) unaffected; `cargo fmt --check` clean.

I don't have SVE hardware to run the runtime kernel tests — this is validated at the build/link level across the toolchains and cfg states.

### DIFF
--- a/linalg/build.rs
+++ b/linalg/build.rs
@@ -141,6 +141,37 @@ fn compiler_supports_sve() -> bool {
         .is_ok()
 }
 
+// Probe which `-march` spelling this C compiler accepts for the SVE f16 kernels.
+// Native FP16 arithmetic is spelled `+fp16` by GCC's aarch64 backend but
+// `+fullfp16` by LLVM/clang — and therefore by `zig cc`, which cargo-zigbuild
+// drives to cross-compile the *-musl targets. Hardcoding `+fp16` makes the whole
+// crate fail to build under zig cc with "unknown CPU feature: 'fp16'". Trying the
+// spellings in turn keeps the f16 kernels building under either toolchain with
+// identical codegen; if neither is accepted we return None and the caller skips
+// the f16 kernels (and the `tract_sve_fp16` cfg), leaving the +sve-only f32/i32
+// kernels intact. The `-march` string is rejected during the compiler's target
+// parse, before codegen, so the base-SVE probe source is enough to discriminate
+// an accepted spelling from a rejected one.
+fn compiler_sve_fp16_flag() -> Option<String> {
+    let out_dir = path::PathBuf::from(var("OUT_DIR"));
+    let probe = out_dir.join("sve_fp16_probe.c");
+    fs::write(&probe, "#include <arm_sve.h>\nint p(void){ return (int)svcnth(); }\n").unwrap();
+    ["-march=armv8.2-a+sve+fp16", "-march=armv8.2-a+sve+fullfp16"]
+        .into_iter()
+        .enumerate()
+        .find(|(i, march)| {
+            cc::Build::new()
+                .file(&probe)
+                .flag(march)
+                .cargo_metadata(false)
+                .cargo_warnings(false)
+                .warnings(false)
+                .try_compile(&format!("tract_sve_fp16_probe_{i}"))
+                .is_ok()
+        })
+        .map(|(_, march)| march.to_string())
+}
+
 fn jump_table() -> Vec<String> {
     println!("cargo:rerun-if-changed=src/frame/mmm/fuse.rs");
     std::fs::read_to_string("src/frame/mmm/fuse.rs")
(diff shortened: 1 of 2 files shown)


kali (approved):

kali:
Thank you!

PR #2494, opened 2026-07-20
refactor(silu): delegate f16 silu to the linalg kernel

cverrier:
The f16 silu op computed x * sigmoid(x) with its own scalar loop, duplicating the math already implemented and tested in linalg's HSiLU8 kernel. Route the f16 path through tract_linalg::ops().silu_f16 so it matches the f32 path, drops the duplication, and picks up the vectorized f16 kernel (e.g. AVX-512) where available.

### DIFF
--- a/core/src/ops/nn/silu.rs
+++ b/core/src/ops/nn/silu.rs
@@ -6,13 +6,7 @@ use crate::ops::nn::Sigmoid;
 use tract_data::half::f16;
 
 element_wise!(silu, Silu,
-    [f16] => |_, xs| {
-        xs.iter_mut().for_each(|x| {
-            let xf = x.to_f32();
-            *x = f16::from_f32(xf / (1.0 + (-xf).exp()));
-        });
-        Ok(())
-    },
+    [f16] => |_, xs| { (tract_linalg::ops().silu_f16)().run(xs) },
     [f32] => |_, xs| { (tract_linalg::ops().silu_f32)().run(xs) };
     cost: |dt| {tvec!((Cost::FMA(dt), 12), (Cost::Div(dt), 1))};
     declutter: detect_silu


kali (approved):

kali:
Did we expect this regression ?

cverrier:
> Did we expect this regression ?

No, and I don't think it's related to the changes introduced in this PR: the preprocessor is a f32 mel-spectrogram featurizer, with no Sigmoid/SiLU nodes at all (I also check Sigmoid because we might have some "unfused SiLU" cases) ; I double-checked this by looking at the final op graph as JSON:

```sh
target/release/tract \
  ~/.cache/tract-ci-minion-models/parakeet-tdt-0.6b-v3-f32f32.608.preprocessor.nnef.tgz \
  --nnef-tract-transformers -t transformers_detect_all --set B=1 --set A=16000 \
  dump --audit-json \
  > tmp.json
```

I also ran a `bench-compare` on a AVX-512 machine to see if I could reproduce the regression:

```sh
# a3997659b == the PR's parent commit; HEAD == the PR tip
.travis/bench-compare.sh a3997659b HEAD --filter parakeet-tdt-600m-v3-f32f32-preprocessor
```

Per the results, it seems that there is no regression:

```
# evaltime  A -> B   (+ = B slower)
      +0.2%  net.parakeet_tdt_600m_v3_f32f32_preprocessor_1s.evaltime.cpu      0.0045 -> 0.0045
      +0.0%  net.parakeet_tdt_600m_v3_f32f32_preprocessor_1s.evaltime.cuda      0.0003 -> 0.0003

0 regression(s) >2%, 0 improvement(s) >2%
```

So the reported 2.12 ms → 4.59 ms looks like runner noise on a microbench rather than anything from this change. A re-run of the bench job should clear it. By the way, are the noise floor for microbenches in [`./travis/bench-thresholds.toml`](https://github.com/sonos/tract/blob/a3997659bf8bcb7e6315e67cc1fc476d8a7b7e09/.travis/bench-thresholds.toml) too restrictive?

I also noticed that there is [a failing check](https://github.com/sonos/tract/actions/runs/29728486479/job/88317541756?pr=2494):

```
failures:

---- kernels::flash_attn::tests::test_nernst_fattn stdout ----
Error: Mismatch. First outlier: Approximate for F16) at [1, 17, 0, 8] 0.8417969 != 0.8642578. Outliers: 3968 / 524288 = 0.00757 > 0.00000.


failures:
    kernels::flash_attn::tests::test_nernst_fattn
```

It's also surprising to me to see this failing test pop after the changes introduced by this PR; could the tolerance for this test be too restrictive?

kali:
Yeah, i think they are both false positive...

cverrier:
> 🔴 **Bench vs main — 1 speed regression(s)** · ⚠️ 8 secondary

🫠

cverrier:
@kali I think [the bench](https://github.com/sonos/tract/actions/runs/29808540510) failed to run

kali:
p***in

kali:
i'm having a look. good puzzle.

kali:
Root cause: rustc 1.97.1 miscompiles the preprocessor's STFT op (~2×). Not a tract regression, not the PR code.

  What's happening
  - The parakeet…preprocessor_1s evaltime·cpu "+117%" (2.1→4.6ms) on PR bench comments is entirely in the STFT
  op (0.8→2.6ms via --profile); every other op is unchanged. It's model-specific because only the preproc has
  an STFT, and linalg's mmm kernels are runtime-dispatched assembly (compiler-immune), so convnets/LLMs are
  unaffected.
  - Cause is rustc 1.97.1 (released 2026-07-14; CI's rust-toolchain.toml tracks floating stable). 1.96.1 →
  2.3ms, 1.97.1 → 4.1ms for the same source.
  - It's source-fragile: 1.97.1 emits slow STFT code for some source variants and fast for others. main-head
  compiles fast → the nightly reference stays fast (2.0ms); specific PRs (2494 silu, 2500 einsum — both
  unrelated to STFT) compile slow → flagged as regressions. That's why it's per-PR, not fixed by re-running,
  and started ~07-16 as runners picked up 1.97.1.

  Evidence (on an exact clone of the CI box)
  - Downloaded CI artifacts run slow idle here (2494 = 4.1ms, 2500 = 4.0ms); the nightly artifacts run fast
  (a3997659b = 1.99ms, main = 2.04ms). Same runner/rustc — the only difference for 2494 is one unrelated silu
  commit on top of the fast nightly base.
  - Reproduced from source: 2494 source + rustc 1.97.1 = 4.0ms (byte-identical size to the CI artifact); +
  rustc 1.96.1 = 2.3ms.

kali:
So I'm merging this PR, going to look at the root cause more precisely, and hopefully recover this parakeet preproc perf.

Issue #2475, opened 2026-07-14
Qwen2.5-7B-Instruct-q40ef16: NaN logits (always on CPU, input-dependent on Metal); Qwen3-1.7B-q40ef16 is fine


czoli1976:
`Qwen/Qwen2.5-7B-Instruct` in the CI-published **q40ef16** variant produces **NaN logits** in tract: *always* on the default CPU runtime, and *input-dependently* on Metal. The same code path with the smaller **Qwen3-1.7B-q40ef16** is perfect, so this is not a general q40ef16 problem.

Reproduces with the stock `causal_llm` example only — no custom code.

## Repro

```sh
wget https://s3.amazonaws.com/tract-ci-builds/tests/llm/current/Qwen--Qwen2.5-7B-Instruct-q40ef16/Qwen--Qwen2.5-7B-Instruct-q40ef16.nnef.tgz
wget https://huggingface.co/Qwen/Qwen2.5-7B-Instruct/resolve/main/tokenizer.json

cargo run --bin complete --release -- --force-cpu -t tokenizer.json \
    -m Qwen--Qwen2.5-7B-Instruct-q40ef16.nnef.tgz -n 8 "The capital of France is"
```

| model / runtime | result |
|---|---|
| Qwen2.5-7B-Instruct-q40ef16, **CPU** | empty / garbage — logits are all NaN on every step |
| Qwen2.5-7B-Instruct-q40ef16, **Metal** | *input-dependent*: `"The capital of France is"` → correct (`C. Paris`), but a ChatML-wrapped prompt → NaN on many steps |
| **Qwen3-1.7B-q40ef16**, CPU | ✅ correct: `"The capital of France is Paris. The capital of the United States is Washington"` |

On Metal the NaN is intermittent, which is what makes it look like garbled text rather than an outright failure — real tokens interleaved with the argmax-of-NaN token:

```
Ciao! Come posso aiutarti oggi?   ->   "! Ciao!! Come! pos! posso!!!! ai!ut!i!o!t"
```

(every `!` is token 0, i.e. `argmax` over an all-NaN logit row)

## Where it comes from

I bisected it by splitting the model into one sub-model per transformer block (28 of them) and running them in sequence, dumping `max|residual|` after each block. Prefill of a 21-token prompt, Metal:

```
block  0: max|res|      63.2  non-finite 0
block  1: max|res|      30.4  non-finite 0
...
block 25: max|res|      11.2  non-finite 0
[...]

PR #2495, opened 2026-07-20
arm64: honor f16 unicast operand in mmm_f32 32x1/32x3 add_unicast

czoli1976:
The arm64simd mmm_f32_32x1 / mmm_f32_32x3 and x86_64 fma_mmm_f32_32x1 / fma_mmm_f32_32x3 kernels declare store(f16) but their .add_unicast path never handled an f16 operand: it ignored item_size and always loaded the unicast operand as f32, so a fused f16 AddUnicast (e.g. a residual Add folded into a native q40 GEMV) was read at the wrong width and could saturate to inf.

Each kernel gains the missing item_size == 2 branch that converts the operand before the add — fcvtl on arm64, vcvtph2ps on x86_64 — mirroring the existing store_f16 path (both the contiguous and strided variants).

Tests:

add_unicast_dt (store-dtype-gated): operand dtype = the kernel's declared store dtype, covering the f16-into-f32 case for every f16-storing kernel. Fails on all four kernels without the fix, passes with it.
add_unicast_pattern (new): reads the operand across the col-major, row-major and arbitrary StoreLayouts, so both the contiguous and strided add_unicast load paths are checked for every store dtype.
Full tract-linalg suite green. Qwen2.5-7B-Instruct-q40ef16 decodes correctly on arm64 CPU (previously NaN); on x86_64 CPU the same kernels carried the latent bug (the add_unicast_dt test fails there without the fix) and the model continues to decode correctly.

Root-caused from #2475; complements #2477 (which gates off the fusion) by making the kernels themselves safe.

🍍

### DIFF
--- a/linalg/arm64/arm64simd/arm64simd_mmm_f32_32x1_core.S.j2
+++ b/linalg/arm64/arm64simd/arm64simd_mmm_f32_32x1_core.S.j2
@@ -298,6 +298,10 @@
 
 .add_unicast:
     ldp         x5, x6, [x0, #8]           // c base ptr, rsc
+    ldr         x8, [x0, #32]              // item_size
+    cmp         x8, #2
+    beq         .add_unicast_f16
+
     cmp         x6, #4
     beq         .do_per_row_add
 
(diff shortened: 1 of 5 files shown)


czoli1976:
@JulienBalianSonos

kali (dismissed):

kali:
ok ! bug it is.

kali:
These are the only two kernels impacted ?

kali:
> These are the only two kernels impacted ?

nope.

kali:
#2477 is superseded, right ?

JulienBalianSonos:
@czoli1976 Nice finding, congrats! I will let the review go to @kali, as I am not in my comfort zone with assembly, but it 'sounds' like a legit fix.

czoli1976:
It appears so

kali:
@czoli1976 are you in a position to address the fma ones, or should I do it ?

czoli1976:
> @czoli1976 are you in a position to address the fma ones, or should I do it ?

let me look

kali (dismissed):

kali (approved):

PR #2500, opened 2026-07-21
core/einsum: sink unmerge reshape below matmul epilogue to restore fu…

kali:
…sion

The merge-same-role-axes rule folds spatial axes into the matmul M axis and splits them back with an unmerge reshape placed on the matmul output. That reshape sat between the matmul and its epilogue (concat-split partial-product sums, per-channel BatchNorm affine, Relu), blocking OptMatMul fusion and leaving those as standalone elementwise passes. A companion rule now sinks the reshape below such binaries — collapsing a per-channel constant through the split, or hoisting an identical sibling reshape out of a sum — so the epilogue fuses into the matmul again.


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2501, opened 2026-07-21
linalg/mmm: gate inner cache-block on the streamed operand spilling L2

kali:
The single-thread tile walk blocked the inner tier whenever an L2 size was detected, even for shapes whose re-streamed packed operand already fits L2 — there the naive stream re-reads it from cache and the blocked traversal only disturbs the prefetchers. An inner_tier_pays check, mirroring the existing outer tier, keeps the inner block only when the streamed operand (A for a column-outer walk, B for a row-outer one) spills L2; otherwise the walk stays a single naive block.

### DIFF
--- a/linalg/src/frame/mmm/mod.rs
+++ b/linalg/src/frame/mmm/mod.rs
@@ -384,10 +384,36 @@ fn block_edge_for(
     (budget / per_blk).clamp(1, cap)
 }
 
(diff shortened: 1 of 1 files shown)


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2502, opened 2026-07-21
STFT: gather each frame from a contiguous signal buffer

kali:
The per-frame window was rebuilt by re-walking the strided input view with iter().tuples().skip(stride*f).take(frame), an O(frames^2) gather whose hot ndarray iterator was also left to the optimizer's inlining whims. Collect the signal into a contiguous Vec<Complex> once per outer coordinate and copy each frame out of it with extend_from_slice.

### DIFF
--- a/core/src/ops/fft.rs
+++ b/core/src/ops/fft.rs
@@ -181,16 +181,11 @@ impl Stft {
                     (c..=c).into()
                 }
             });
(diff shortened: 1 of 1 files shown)


kali:
ping @cverrier . This is the actual fix for the regression. #2494 was just unlucky to trigger the mis-compilation.

PR #2503, opened 2026-07-22
cuda: add recurrent Gated DeltaNet decode kernels

Reza2kn:
Adds CUDA single-token decode operators for recurrent Gated DeltaNet state updates and four-tap causal-convolution cache updates.
**Highlights**
- FP16 q/k/v/output with FP32 recurrent state
- device-resident typed operators
- coalesced, race-free recurrent state writes
- CUDA-vs-CPU tests validated on SM120
- Qwen3.5 iterative decoder example with EOS/repetition handling
The model-specific BF16 parity experiments are intentionally excluded from this PR; this is the reusable kernel/runtime portion only.


kali:
Interesting work ! Couple of high-level remarks about architecture / strategy so we can find a path to merging. This is not a line-per-line review, just an overview.

* I try to maintain coverage parity between cuda and metal. Code bots have made this easier. If you're not in a position to test on a mac, get your assistant to do their best, commit it, we'll find somebody to help if the CI is unhappy. This may probably lead you to move the op-level code in tract-gpu, and keep a simple shared interface for the kernel function. You'll find examples all over the place. 
* It is preferred for examples to be "in workspace" so they are covered by the ci cargo check. Give it a "ci.sh" that can do everything (download or export the model, run a test). This will run on demand or once a week so we know the examples are maintained.
* Can you write the example against the public API ? The one from the "tract" crate, from api/rs ? Or did you hit snags trying to do so ? I am reluctant to include examples that use the "old" everything-is-public approach.

Reza2kn:
Thanks @kali! Addressed all three points:

- Shared op logic moved to `tract-gpu`, with both Metal and CUDA kernels.
- Example is now inside the workspace with a self-contained `ci.sh`.
- Example now uses only the public `tract` API.

we now have a genuinely representative benchmark using the full 1.1 GB Qwen3.5 decoder, the same handwriting fixture, and identical 128-token decoding. First run was discarded as warm-up; results are medians of three runs.

| Backend | Hardware | Avg decode | Steady-state | Throughput |
|---|---|---:|---:|---:|
| Metal | Apple M2, 24 GB | 95.73 ms/token | 82.58 ms/token | 10.45 tok/s |
| CUDA | RTX 5080 Laptop, 16 GB | 19.56 ms/token | 18.99 ms/token | 51.13 tok/s |

CUDA was approximately **4.9× faster end-to-end**. Both backends produced identical 128-token output.

The compatibility fix required by the real exported graph was also pushed to the PR as commit `ced931872`.

kali:
Thank you for going the extra mile and doing the metal port. I realised a bit late that our "example" is more a harness test than an actual "example" meant to be demo how to integrate an actual model using tract. We may just want to move it to harness. Should we extract the nnef files from the ci.sh herfiles and make them "models" in their own right ? Is part of what the main.rs achieve possible through the command line : a relatively canonical form for unit testing operators in tract is a small nnef model (model_folder/graph.nnef with inlined tensors) + one or more io.npz files that we can run through the command line. This limit the number of artefacts to build as we can re-user the tract cli for multiple tests.

Reza2kn:
## Refreshed full-decoder benchmark

Re-ran the parent build and `2e83b3898` on hardware we have direct access to, using the same full Qwen3.5 recurrent decoder model and identical NPZ state fixture for both builds.

Protocol: one warm-up discarded, median of three measured process runs, greedy decode to EOS (14 generated tokens). The complete token sequence was identical between parent/new and Metal/CUDA in every run.

| Hardware / backend | Revision | Avg decode (ms/token) | Steady-state median (ms) | Throughput (tok/s) | Delta vs parent |
|---|---:|---:|---:|---:|---:|
| Apple M2 24 GB / Metal | parent (`ced931872`) | 48.964 | 46.755 | 20.42 | baseline |
| Apple M2 24 GB / Metal | `2e83b3898` | 50.139 | 48.092 | 19.94 | -2.35% throughput |
| RTX 5080 Laptop 16 GB / CUDA | parent (`ced931872`) | 16.273 | 12.393 | 61.45 | baseline |
| RTX 5080 Laptop 16 GB / CUDA | `2e83b3898` | 16.411 | 12.390 | 60.94 | -0.83% throughput |

Detailed deltas:

| Backend | Avg decode | Steady-state | Throughput |
|---|---:|---:|---:|
| Metal / M2 | +2.40% | +2.86% | -2.35% |
| CUDA / RTX 5080 Laptop | +0.85% | -0.02% | -0.83% |

For Stallion, the CUDA 13 build used its existing Conda CUDA 13/cuDNN/CCCL packages via a temporary user-space `CUDA_HOME`; stderr confirmed `loaded with cuda` for all four processes. No system packages or symlinks were changed.

These paired deltas are much smaller than the earlier unrelated-model regressions and are consistent with run-to-run/system noise. The CUDA result also confirms that moving GDN out of the eager library list did not regress steady-state full-decoder performance.

kali:
By the way, don't pay attention to the automatic bench messages. It's still pretty fragile.

kali (dismissed):

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

Rebased to hopefully de-noise benches.

kali:
@Reza2kn any other business before we move to final review / merge ?

kali:
Hey @Reza2kn, this PR is still in draft. Should we consider it ready ?

Reza2kn:
Hi @kali sorry for the late reply! and yes! this should be ready now!

kali:
Thank you !

PR #2506, opened 2026-07-22
fix: bound the x86 SiLU negative tail

cverrier:
The AVX-512 SiLU kernels multiplied the sigmoid by the unclamped input, so for large negative x — where the sigmoid kernel floors its output at the constant sigmoid(-18.0) — the result x * sigmoid(-18.0) diverged toward -inf instead of decaying to 0 as true SiLU does. This is the same thing that we did in commit 2aca983fed6ce5b5f6ab948fb340a6763a37704d for the ARMv8.0 SiLU kernel.

I also made the SiLU testing suite more exhaustive: the proptest and trivial cases stay within [-10, 10], so the tail behavior of the SiLU kernels was untested: the arm64 and x86 kernels clamp the sigmoid argument and each saturates the tails differently, while the generic kernels compute exact sigmoid. Add a deterministic case spanning [-100, 100] to pin all of them at and beyond the clamp boundary.

### DIFF
--- a/linalg/src/frame/silu.rs
+++ b/linalg/src/frame/silu.rs
@@ -38,6 +38,15 @@ pub mod test {
                         .unwrap();
                 }
             }
+            #[test]
+            fn tails() {
+                if $cond {
+                    $crate::frame::silu::test::test_silu::<$ker, $t>(&[
+                        -100.0, -30.0, -18.6, -15.0, 15.0, 18.6, 30.0, 100.0,
+                    ])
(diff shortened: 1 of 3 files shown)


kali (approved):

kali:
Your PR is riding my efficiency fixes. don't worry about the jetson regression, they're fake.

kali:
Ready to merge ?

cverrier:
> Ready to merge ?

Ready!

PR #2513, opened 2026-07-23
refactor(linalg): wrap x86 f16 silu/gelu over the fused f32 kernels

cverrier:
The AVX-512 f16 SiLU and GELU kernels re-implemented, on their f32 scratch, the exact composition already performed by the fused f32 SiLU and GELU kernels. Collapse them into thin f16<->f32 conversion wrappers around those kernels, matching how the hardswish/leaky_relu f16 kernels already delegate. The math is unchanged; a single scratch buffer now suffices since the f32 kernels keep their own internal save buffer.

Also, the x86 f16 activation kernels zero-initialised their f32 round-trip scratch even though every element is written by cvt_f16_to_f32 before the f32 kernel or cvt_f32_to_f16 reads it. Use MaybeUninit instead, as the arm64 f16 activations already do, avoiding the pointless clear.

### DIFF
--- a/linalg/src/x86_64_fma/act_f16.rs
+++ b/linalg/src/x86_64_fma/act_f16.rs
@@ -214,8 +214,7 @@ pub mod test_x86_64_avx512_tanh_f16_16n {
     tanh_frame_tests!(is_x86_feature_detected!("avx512f"), f16, x86_64_avx512_tanh_f16_16n);
 }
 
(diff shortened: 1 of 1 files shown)


kali (approved):

cverrier:
@kali Once this one is merged, I'll mark #2512 (implements the f16<->f32 macro) as ready for review: these x86 kernels will also benefit from that macro.

PR #2522, opened 2026-07-24
fix(api): keep runnable state live across run() calls

cverrier:
The facade State froze and unfroze the whole execution state on every run(), reallocating the matmul scratch space and deep-copying the pulsed Delay buffer each frame. This slowed down streaming models invoked once per frame. Keep the core state live between runs and freeze it only on clone so the persistent state is no longer torn down and rebuilt every call.

### DIFF
--- a/api/rs/src/lib.rs
+++ b/api/rs/src/lib.rs
@@ -352,8 +352,7 @@ impl RunnableInterface for Runnable {
     }
 
     fn spawn_state(&self) -> Result<State> {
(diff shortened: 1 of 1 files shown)


kali:
I think we have a soundness problem. State is not Send (it has Rc<...>), so we need to go through freezing when a State needs to be move from one thread to another. I worry about the implications wrt the C API.

cverrier:
Closing this as it's being addressed in #2664 .

PR #2523, opened 2026-07-26
Parallelize the serially-executed operators via a row-parallel linalg primitive

simon0191:
## Parallelize the serially-executed operators

tract multi-threads the matmul microkernel but nothing else, so on a transformer encoder the serial remainder (Softmax, RmsNorm, the element_wise / GELU tail, last-axis reductions, residual Add / Mul) caps multi-core speedup. This PR adds one row-parallel primitive in tract-linalg and routes the safe consumers through it.

### What changed
- **New primitive** `tract_linalg::multithread::par_chunks_mut(out, row_len, total_elems, f)`. It splits a tensor's outer/row axis into contiguous whole-row chunks and dispatches them across the executor already installed during eval (the same one matmul consults), or runs inline when the executor is single-threaded (including a one-thread pool), there are fewer than two rows, or the work is below a new element-count threshold.
- **Threshold + CLI knob**: `THREADING_ELEMENT_THRESHOLD` (default 32768 elements) with `current/set_threading_element_threshold`, plus a `--threading-threshold` CLI flag for tuning and benchmarking (0 = always thread).
- **Consumers routed through it**: the `element_wise!` ops (GELU, SiLU, Sigmoid, Tanh, and the ~40 others), the RmsNorm fast path, last-axis Softmax / LogSoftmax, last-axis Reduce (Sum / MeanOfSquares), and same-shape generic binary.

### Bit-exactness (the one correctness invariant)
Only the outer/row axis is ever split. Each per-row reduction runs entirely inside a single worker, so accumulation order is unchanged and the output is bit-identical to the single-threaded path at any thread count. Elementwise and binary ops have no reduction, so any chunk boundary is safe. The Softmax and Reduce fast paths are additionally guarded to trigger only when the reduced region is a contiguous trailing block of a naturally-strided tensor; anything else falls to the existing serial path.

Verified with an exact (atol = rtol = 0) serial-vs-parallel A/B on a synthetic model exercising the element_wise, softmax, reduce, and binary paths (new `harness/nnef-test-cases/parallel-ops`, whose `runme.sh` regenerates a same-platform serial reference and asserts the forced-parallel run matches it exactly). Also green: `test-unit-core` (783) and `test-nnef-cycle` (2392). Core adds zero net-new `unsafe`.

### Performance
All numbers below were measured on an Apple M4 Pro (14 cores: 10 performance + 4 efficiency), macOS, arm64. Each model is a single op over 16.7M f32 values, comparing `--threads 1` (real serial) against `--threads N`. The machine was busy with other work during the run, so the true speedups are a little higher.

| workload | serial | 2 thr | 4 thr | 8 thr |
|---|--:|--:|--:|--:|
| Softmax [4096×4096] | 23.6 ms | 1.70× | 2.37× | 3.01× |
| element_wise chain [8192×2048] | 29.9 ms | 1.72× | 2.79× | 3.79× |

These ops are limited by memory speed, not by core count. On servers with slower memory (for example AWS Graviton), the speedups should be even bigger than the Apple numbers above.

### Scope and boundaries
- tract-core stays rayon-free and unsafe-free; all rayon and the (pre-existing, unchanged) unsafe stay in tract-linalg. The primitive's signature is un-gated, so core compiles identically with or without the `multithread-mm` feature.
- A one-thread `MultiThread` executor now runs these ops inline (serial) instead of falling through to the global pool, so `--threads 1` is genuinely serial.
- Residual Add / gated Mul that lower to `OptBin*`, and matmul input packing, are deliberately left out of this PR to keep it focused; they can be parallelized the same way in follow-ups.

🍍


kali:
Hey ! thanks for your interest in tract. This looks pretty good to me. Out of curiosity, can you tell us what machine you benched on ? You just imply it's not an AppleSilicon system.

kali:
Please ignore the LLM reds, these ones come from main.

simon0191:
> Hey ! thanks for your interest in tract. This looks pretty good to me. Out of curiosity, can you tell us what machine you benched on ? You just imply it's not an AppleSilicon system.

Hey!

It was actually an Apple Silicon machine: an M4 Pro with 14 cores. All the numbers in the table came from there.

Sorry for the confusing wording. In the line that mentioned "Apple Silicon" and "Graviton" I meant these ops are limited by memory speed, not by core count, so on a server with slower memory the speedups should be even bigger than the numbers I posted. I'll reword that line so it's clearer. The machine was also busy with other work during the run, so if anything the real speedups are a bit higher.

Thanks for the quick review!

PR #2526, opened 2026-07-27
Multi-thread the OptBinUnicast / OptBinByScalar elementwise fast paths

simon0191:
Multi-threads the two optimised elementwise binary ops, `OptBinUnicast` and `OptBinByScalar`, which until now always ran on one thread. On a transformer encoder these are the biggest group of nodes after `Const` and about 7.7% of single-threaded runtime, so they cap how far the model can scale once the matmuls are already threaded.

**This PR builds on #2523 and cannot compile without it.** See "Relationship to #2523" at the bottom.

---

## Background: what was single-threaded

tract multi-threads the matmul microkernel. Everything else in a graph runs on the calling thread. `codegen` lowers most elementwise binary ops to one of two fast-path ops:

- **`OptBinUnicast`** — `b` matches `a` except for some leading size-1 axes. The classic case is a residual add, `[1, seq, hidden] + [1, 1, hidden]`.
- **`OptBinByScalar`** — `b` has trailing size-1 axes, so one `b` value applies to a whole slab of `a`. Example: `[1, seq, hidden] * [1, seq, 1]`.

Both had an `eval` that looped over the outer axes and called the linalg kernel once per iteration, on one thread.

I profiled a real encoder to check this was worth doing at all (a GLiNER PII model, ModernBERT backbone, sequence length 384, Apple Silicon). Every binary op in it lowers to one of these two — there are **zero** generic `Add`/`Mul` nodes left:

| op | nodes | share of single-threaded runtime |
|---|--:|--:|
| `OptMulUnicast` | 153 | 2.9% |
| `OptMulByScalar` | 77 | 1.2% |
| `OptAddUnicast` | 57 | 2.4% |
| `OptSubByScalar` | 39 | 0.9% |
| `OptAddByScalar` | 19 | 0.3% |
| `OptSubFByScalar` | 1 | 0.0% |
| **total** | **346** | **7.7%** |

## What this PR does

Adds `tract_linalg::multithread::par_bin` and routes both `eval` methods through it. Each one collapses to a single call — their old two-branch outer loop turns out to be just "one block" versus "many blocks".

`par_bin` treats `a` as **blocks** of `period` elements, where `period` is how many elements of `a` a single kernel call covers, and hands blocks to the executor that is already installed during eval (the same one the matmul dispatcher reads, so there is no new plumbing or configuration). A `BShare` argument says how `b` lines up with the blocks:

- `Lockstep` — `b` is one `period`-element row that every block consumes in step. This is what the unicast kernels want.
- `PerBlock` — `b` holds one scalar per block. This is what the by-scalar kernels want.

### Why blocks, and why the split also goes *inside* a block

Two properties of the existing kernels constrain this, and both are easy to get wrong:

**1. A block is the largest region one kernel call may cover.** `UnicastKer::bin` walks `b` in lockstep with `a`, indexing `b[num_element_processed..]`. It does not wrap around. If you hand it more than one period of `a` against a single `b`, it indexes past the end of `b` and panics. So the old per-iteration loop was doing real work, not just bookkeeping, and `par_bin` keeps that loop rather than merging iterations. It is deliberately **not** behind the `multithread-mm` feature flag for that reason: a build without the feature needs it too.

**2. Splits inside a block have to preserve `a`/`b` alignment.** `unicast_with_alignment` asserts that `a` and `b` have the same `align_offset`. So chunk boundaries inside a block land on `vector_size()`-element multiples, which keeps the two offsets congruent. `OptBinUnicast::check_b_alignement` already guarantees `period` itself is such a multiple whenever there is more than one block, which is what makes this work out.

Splitting inside a block is not an optimisation — it is required. At batch size 1 the outer axis has length 1, so a block-only split would give exactly one task and no parallelism at all, on the shape that matters most.

## Correctness

These kernels are pure elementwise, so **no chunk boundary can change a result**. Output is bit-identical to the single-threaded path at any thread count. Nothing here splits a reduction, which is the case where a split would reorder float accumulation.

Verification, in order of how much I trust it:

1. **Bit-exact A/B on the real encoder.** Ran it single-threaded, saved the outputs, then re-ran at 2, 4, 8 and 14 threads asserting `--approx exact` (atol = rtol = 0, so any difference at all fails). All exact, at both the default threshold and with the threshold forced to 0 so even small tensors take the parallel path.
2. **Repeated under a debug build.** The `unsafe` here is byte-offset arithmetic, so this is about memory safety, not just values. Debug asserts and overflow checks on, same result.
3. **New harness case** `harness/nnef-test-cases/parallel-bin`. One graph covering every split shape at two block/period ratios, so both `BShare` variants run at a nonzero offset inside a block. It generates its own single-threaded reference on the same binary and platform, then asserts the threaded run matches exactly — so it does not depend on which SIMD kernel the host happens to register.
4. **Existing suites**: `test-unit-core` 783 pass, `test-nnef-cycle` 2392 pass, `tract-core` 262 pass. `cargo clippy --workspace` and `cargo clippy -p tract-linalg --features multithread-mm` clean. Formatted with the pinned stable rustfmt.

I mutation-tested the new harness case rather than assume it works, and it is worth knowing what it does and does not catch. It compares two runs of the *same* binary, so it detects a parallel-vs-serial divergence but not a bug both paths share. Forcing the intra-block `b` offset to 0 (code only the parallel path runs) fails the case. Forcing the per-block `b` offset to 0 (code both paths run) passes it — and instead fails `test-unit-core` and three `test-nnef-cycle` layer-norm tests, which compare against a real reference. Both layers are needed; neither is sufficient alone.

## About the `unsafe`

`par_bin` contains one `unsafe` block, which builds sub-`TensorView`s over disjoint byte ranges via `TensorView::from_bytes`. It lives in linalg so that `tract-core` gains no `unsafe` and no rayon dependency from this change.

Flagging it explicitly because `linalg/src/multithread.rs` is a dispatcher rather than a kernel: the safety argument is that `offset + len <= period` and blocks are disjoint, so no two tasks ever touch the same byte. That invariant is spelled out in a comment next to the block. Happy to restructure if you would rather this be expressed differently.

## Performance

Isolated the three code paths in a graph of just three `OptBin*` ops over 4.19M f32 each, minimum of 5 runs, compared against this PR's own merge base:

| threads | before | after | speedup |
|--:|--:|--:|--:|
| 1 | 1.32 ms | 1.30 ms | 1.01× |
| 2 | 1.40 ms | 0.88 ms | 1.60× |
| 4 | 1.30 ms | 0.67 ms | 1.94× |
| 8 | 1.38 ms | 0.67 ms | 2.07× |

"Before" is flat across thread counts, as expected for ops that ignored the executor entirely. One thread is unchanged, so there is no cost to the single-threaded path.

The plateau at 4 threads is memory bandwidth, not a threading problem: roughly 150 MB of traffic in 0.67 ms is about 224 GB/s, which is this machine's ceiling. These ops are bandwidth-bound by nature, so the ceiling will differ on other hardware — lower headroom on a bandwidth-rich desktop part, more on a bandwidth-constrained server CPU.

**I am not claiming an end-to-end win on the full encoder from this PR.** I measured 1.08× at 4 threads and 1.00× at 8, but the machine was heavily loaded during benchmarking (load average 8–27 from unrelated work) and that spread is inside the noise. A trustworthy whole-model number needs a quiet machine, and I would rather say so than present a figure I do not believe. The op-level numbers above are robust because they use a minimum over repeats.

## Relationship to #2523

This is stacked directly on #2523, and depends on it in two concrete ways:

- `par_bin` calls `current_threading_element_threshold()`, the size gate #2523 introduces. Below that threshold, dispatch overhead costs more than the parallelism gains.
- The new harness case uses the `--threading-threshold` CLI flag from #2523 to force the parallel path on.

Neither exists on `main` today, so this does not build standalone. **The diff you see currently includes #2523's commit**; once #2523 lands it will shrink to just the second commit. Reviewing that second commit alone is enough.

If you would prefer this squashed into #2523, or split differently, say so and I will restructure — I kept them separate because #2523 is safe code and this one carries the `unsafe`, so they seemed worth reviewing independently.

Happy to rebase onto current `main` whenever suits; `main` has moved 4 commits ahead of #2523's base, though the merge is clean today.

🍍

### DIFF
--- a/core/src/ops/binary.rs
+++ b/core/src/ops/binary.rs
@@ -1,10 +1,10 @@
 use crate::internal::*;
-use crate::ndarray::Dimension;
 use downcast_rs::Downcast;
 use dyn_eq::DynEq;
 use std::fmt::{self, Debug};
 use tract_data::itertools::izip;
 use tract_itertools::Itertools;
+use tract_linalg::multithread::BShare;
 use tract_linalg::{BinOp, LinalgFn};
 
 use super::math::{Add, Max, Min, Mul, Sub};
@@ -679,9 +679,7 @@ impl EvalOp for OptBinByScalar {
             return Ok(tvec!(self.binop.eval(a, b, c_dt)?.into_tvalue()));
         }
 
-        // Not a requirement as TensorView doesn't require a owned tensor but in reality
-        // "a "should be mutable (it's omitted here as Rust compiler advise to remove it)
-        let a = a.into_tensor();
+        let mut a = a.into_tensor();
         let b_shape = b.shape();
 
         let first_unary_axis = b_shape
@@ -693,20 +691,12 @@ impl EvalOp for OptBinByScalar {
             .last()
             .context("Cannot use by_scalar when no trailing dimensions are unary")?;
 
-        let iterating_shape = &a.shape()[..first_unary_axis];
-        if !iterating_shape.is_empty() {
-            for it_coords in tract_ndarray::indices(iterating_shape) {
-                let mut view = TensorView::at_prefix(&a, it_coords.slice())?;
-                let b_view = TensorView::at_prefix(&b, it_coords.slice())?;
-                debug_assert_eq!(b_view.shape().iter().product::<usize>(), 1);
-                (self.eval_fn)(&mut view, &b_view)?;
-            }
-        } else {
-            let mut view = a.view();
-            let b_view = b.view();
-            debug_assert_eq!(b_view.shape().iter().product::<usize>(), 1);
-            (self.eval_fn)(&mut view, &b_view)?;
-        }
+        // b carries one scalar per block of a, the blocks being a's axes before
+        // first_unary_axis; check_input_shapes makes b's match a's there.
+        let n_blocks: usize = a.shape()[..first_unary_axis].iter().product();
+        // A zero-sized dim zeroes n_blocks, and par_bin no-ops on an empty a.
+        let period = a.len().checked_div(n_blocks).unwrap_or(0);
+        tract_linalg::multithread::par_bin(&*self.eval_fn, &mut a, &b, period, BShare::PerBlock)?;
         Ok(tvec!(a.into_tvalue()))
     }
 }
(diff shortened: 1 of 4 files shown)


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali:
Thanks !

PR #2530, opened 2026-07-27
onnx: move NonZero to onnx-opl with NNEF serialization

simon0191:
NonZero can be evaluated but not saved: the op lives in the `onnx` crate with no OPL dumper, so dumping any model containing it fails with `No serializer found for node ... NonZero`. This moves the typed op into `onnx-opl` behind a `tract_onnx_non_zero` primitive with a dumper and loader, following the `NonMaxSuppression` layout; the `onnx` crate keeps an inference op that wires it.

The data-dependent output count stays a free symbol resolved from the real output shape at eval time. Its name travels in a `count_symbol` attribute, which keeps round trips idempotent and keeps two NonZero nodes on distinct symbols since their counts are unrelated. A fresh symbol is minted when the attribute is absent, so hand-written NNEF still loads.

While moving the eval, the unchecked tensor views became plain views and `output_facts` now rejects datum types the op cannot handle: block-quant storage used to panic rather than error, a quantized input tested the raw byte and ignored the zero point, and a TDim input aborted the whole declutter pass.

`test_nonzero_example` comes off the skip lists in `test-nnef-cycle` and `test-f16`, so those suites now prove the round trip. `harness/nnef-test-cases/non_zero_two_nodes/` loads a two-node graph, asserts both outputs, and checks the dumped attributes are distinct.

🍍


kali (approved):

kali:
Thank you !

PR #2531, opened 2026-07-28
core: add a last-axis fast path to GatherElements

simon0191:
`GatherElements` reads every output element through two dynamic-rank multi-dimensional index
operations, each one a runtime stride dot-product with bounds checks. That costs about 9 ns
per element. This adds a fast path for the case where the gather axis is the tensor's last
axis, which cuts the op by 9x to 13x, and leaves every other case on the existing code.

## What changed

When the gather axis is the last axis, and the other dimensions of `data` and `indices` are
equal, each leading coordinate addresses one whole row of both operands. The gather is then a
flat lookup inside that row and needs none of the per-element index arithmetic. Any other
shape falls through to the generic path, unchanged.

Two behavior changes worth calling out:

- `output_facts` now rejects an out-of-range `axis`, and a `data` / `indices` rank mismatch.
  Both used to build a model fine and then panic during eval, which unwinds through `api/c`
  and `api/py`. For example `tract_core_gather_elements(..., axis = -1)` arrives as
  `usize::MAX` through `usize::coerce` in `nnef/src/deser.rs`, builds, and panics at run.
  `Gather` already validates this way in `compute_output_shape`. All 4428 ONNX conformance
  tests still pass, so this rejects nothing that used to work.
- On the fast path an out-of-range index value returns an error. On the generic path it still
  panics inside `ndarray`. More on that below.

## Where the win comes from

DeBERTa-v2 and v3 use relative-position attention, and computing it runs 24 of these gathers
per inference, pulling `[heads, seq, seq]` out of `[heads, seq, 2 * buckets]`. They do no
arithmetic at all, so they were pure overhead and the largest cost after matmul.

The op on its own, on those shapes with 12 heads and 512 buckets (Apple M4 Pro):

| output shape | before | after | faster by |
| --- | --: | --: | --: |
| `[12, 128, 128]` | 1.682 ms | 0.128 ms | 13.1x |
| `[12, 256, 256]` | 7.124 ms | 0.823 ms | 8.7x |
| `[12, 512, 512]` | 26.664 ms | 2.233 ms | 11.9x |

End to end on a DeBERTa-v3-base classifier (86M params, fp32, frozen shapes, 1 thread, same
machine):

| seq | total before | total after | this op before | this op after |
| --: | --: | --: | --: | --: |
| 128 | 149.4 ms | 132.2 ms | 46.8 ms (27.8%) | 14.4 ms (9.8%) |
| 256 | 576.7 ms | 344.4 ms | 319.2 ms (42.0%) | 72.3 ms (18.5%) |

The op is serial, so the gain grows with thread count. On an AWS Graviton3 (`c7g.2xlarge`,
8 vCPU) the same change measured 1.50x total at seq 128 on 8 threads, and 1.75x at seq 256 on
8 threads.

## How often the fast path fires

Narrow, but it covers the shape that costs. I instrumented both branches and counted every
`GatherElements` call:

| | on fast path, by node | on fast path, by element |
| --- | --: | --: |
| the DeBERTa-v3 model above, per inference | 24 of 24 (100%) | 100% |
| ONNX conformance suite | 11 of 82 (13%) | 0.1% |
| the new `suite-unit` proptest | 101 of 256 (39%) | |

Every conformance miss was a gather on axis 1 of an `[N, C, d1...]` tensor, coming from the
loss-function expansions (`NegativeLogLikelihoodLoss` and friends). That is the most common
`GatherElements` shape in your test suite and it keeps the generic path.

Two caveats on reading those numbers. I have one real model that uses this op, so 100% is a
single data point. And the conformance suite is a set of small shape fixtures, so its 13% and
0.1% say nothing about real inference time either. The guard is three cheap comparisons before
any work, so shapes that miss pay almost nothing for it.

## Testing

- New `gather_elements` suite in `suite-unit`: a proptest over ranks 1 to 4 that reaches both
  paths, plus explicit cases for last axis, negative indices, rank 1, non-last axis,
  mismatched dimensions, and two zero-element shapes. 9 cases against each of the raw,
  decluttered and optimized runtimes. `data` is filled with its own flat offsets, so the
  expected answer comes from the shape alone rather than from re-running the op's own logic.
- New harness case `harness/nnef-test-cases/gather-elements-last-axis`: gathers the same
  elements twice in one graph, once on the fast path and once transposed onto the generic
  path, then asserts the difference is 0. `--assert-op-count` keeps a future pass from folding
  one leg into the other and making the check vacuous.
- Mutation tested. Deleting the leading-dimension check, the negative-index handling, or the
  row offset each fails both the unit suite and the harness case.
- The model output is bit-identical before and after at both sequence lengths
  (`--approx exact`).
- Four tests in `core` for the out-of-range and empty-axis errors. Those are only reachable on
  the fast path, and the suite framework compares outputs rather than errors, so I could not
  express them there. `core/src/ops/array/gather.rs` already tests op `eval` this way. Happy
  to move them if you would rather have the harness extended instead.
- `test-unit-core` (810), `test-nnef-cycle` (2412), `test-tflite` (1369) and `test-onnx-core`
  (4428) all pass. `cargo clippy --workspace` and `cargo fmt --all` clean.

## Two things I need your eyes on

**A CUDA skip I could not verify.** The two new zero-element cases put a zero-byte constant
through `CudaTensor::from_tensor`, which calls `clone_htod` with an empty slice.
`metal/src/context.rs` guards exactly this by substituting a 1-byte buffer, and CUDA has no
equivalent, so I added those two cases to `ignore_unit` in `test-rt/test-cuda/suite.rs` with a
comment. I have no NVIDIA hardware, so this is reasoning from the code, not a reproduction. If
your GPU CI is fine with them, drop the skip. `gather_elements` is also skipped in
`test-tflite`, which has no serializer for the op.

**Two pre-existing issues this made visible**, both left alone:

1. On 32-bit targets (`armv7-unknown-linux-musleabihf` is a release target) the generic path's
   `... as usize` truncates an out-of-range index into an in-range one and silently returns
   the wrong element. The fast path's `usize::try_from` reports it. Worth a separate fix, and
   it is the reason the two paths now disagree on bad input.
2. When the indices reach the op as `TDim`, `eval` casts the whole tensor to `i64` on every
   inference. In the model above that is a 786k-element cast costing roughly 2.2 ms of the
   3.0 ms per node at seq 256, and the same cause makes two `MultiBroadcastTo` nodes 9.3 ms
   each. That is now the dominant cost in this op, not the gather, and it is why the op is
   still second in the profile at seq 256.

## Deliberately not done

- **Writing straight into the output tensor.** The fast path builds a `Vec` and then pays a
  copy in `ArrayD::into_tensor()`. Writing into the tensor directly is 1.4x to 1.8x faster in
  a microbenchmark, but `Tensor::zero_dt` does not cover `String` or `Blob`, so it needs a
  datum-type gate, and I did not want to narrow the fast path for a gain that the `TDim` cast
  above dwarfs on real models. `Gather::eval_t` does this with `Tensor::uninitialized`; I left
  it alone rather than add `unsafe` here.
- **Threading the loop.** It scales only 1.6x to 2.3x on 8 threads at these shapes, so it is
  bandwidth-bound, and the op is a small share of runtime now.
- **Handling any axis, not just the last one.** The general outer/axis/inner form needs a
  multiply per element, which would slow down the case this is for.

🍍

### DIFF
--- a/core/src/ops/array/gather_elements.rs
+++ b/core/src/ops/array/gather_elements.rs
@@ -1,6 +1,10 @@
 use crate::internal::*;
 use ndarray::*;
 
+/// For every coordinate of `indices`, reads `data` at that same coordinate with
+/// `axis` replaced by the index value found there. The output has `indices`'
+/// shape. A negative index counts from the end of `axis`. `data` and `indices`
+/// must have the same rank; off `axis`, `indices` may be smaller than `data`.
 #[derive(Debug, Clone, new, Hash, PartialEq, Eq)]
 pub struct GatherElements {
     pub axis: usize,
@@ -15,6 +19,57 @@ impl Op for GatherElements {
 }
 
 impl GatherElements {
+    /// Gathers when `axis` is the last one and the leading dimensions of both
+    /// operands are identical. Every leading coordinate then addresses one whole
+    /// row of both operands, so the op is a flat per-row lookup and none of the
+    /// per-element dynamic-rank stride arithmetic of the generic path is needed.
+    /// `Ok(None)` means the operands do not qualify and the caller must use the
+    /// generic path.
+    ///
+    /// Index resolution is identical to the generic path, but an out-of-range
+    /// index is reported as an error instead of panicking inside `ndarray`.
+    fn eval_contiguous_last_axis<T: Datum>(
+        &self,
+        data: &ArrayViewD<T>,
+        indices: &ArrayViewD<i64>,
+    ) -> TractResult<Option<ArrayD<T>>> {
+        let rank = data.ndim();
+        let Some(last_axis) = rank.checked_sub(1) else { return Ok(None) };
+        if self.axis != last_axis
+            || indices.ndim() != rank
+            || data.shape()[..last_axis] != indices.shape()[..last_axis]
+        {
+            return Ok(None);
+        }
+        // Both views come from a plain tract tensor, so they are contiguous and
+        // this never declines; it is handled rather than asserted.
+        let (Some(data_slice), Some(index_slice)) = (data.as_slice(), indices.as_slice()) else {
+            return Ok(None);
+        };
+        let row_len = data.shape()[last_axis];
+        let gathered_len = indices.shape()[last_axis];
+        // A zero gathered length still has leading coordinates to walk, and
+        // walking them would be pure waste for an empty output.
+        let rows = if indices.is_empty() { 0 } else { indices.len() / gathered_len };
+        let mut output = Vec::with_capacity(indices.len());
+        for row in 0..rows {
+            let data_row = &data_slice[row * row_len..][..row_len];
+            for &index in &index_slice[row * gathered_len..][..gathered_len] {
+                let resolved = if index < 0 { index + row_len as i64 } else { index };
+                let value = usize::try_from(resolved)
+                    .ok()
+                    .and_then(|resolved| data_row.get(resolved))
+                    .with_context(|| {
+                        format!(
+                            "Invalid GatherElements index {index} in row {row} on axis of len {row_len}"
+                        )
+                    })?;
+                output.push(value.clone());
+            }
+        }
+        Ok(Some(ArrayD::from_shape_vec(indices.shape(), output)?))
+    }
+
     unsafe fn eval_t<T: Datum>(
         &self,
         data: TValue,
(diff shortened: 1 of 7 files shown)


kali:
Apparently the skip is not required, many locations assume empty allocation of cuda tensors are valid.

PR #2532, opened 2026-07-29
Bound eager const-eval by memory rather than element count

simon0191:
Follow-up to #2445. Limiting eager const-eval to tensors under 16 elements stopped it partway through shape-derived index arithmetic, which `prop_const` then declines to finish because its own fallback gates a *shared* constant input on 1024 elements. Both are element counts sitting next to a guard that already works in bytes, so this replaces them with one shared byte budget and bounds the inference-time fold by the memory it materializes, the way `prop_const` already does.

The model that surfaced it is Llama Prompt Guard 2 86M, a DeBERTa-v3: with input shapes frozen, its whole disentangled-attention relative-position table is constant, and it stopped being folded. Prompt Guard 2 at sequence length 128, single-threaded, optimized graph:

| | ms/inference | nodes | non-const nodes | `OptMatMul` |
|---|--:|--:|--:|--:|
| before #2445 | 99.3 | 908 | 651 | 122 |
| main today | 143.2 | 1030 | 760 | 146 |
| this branch | 102.2 | 908 | 651 | 122 |

The graph is back to exactly the shape it had before #2445, and output is bit-identical to that build (`--assert-output-bundle --approx exact`). The 24 extra matmuls on main are the relative-position embedding projection, about 1.5 MiB, being redone per inference instead of once at load. The same regression reproduces on a Graviton3 `c7g.2xlarge`, where it is 0.68x at sequence length 128 and 0.81x at 256.

The intent of #2445 is kept: a decoded weight tensor is orders of magnitude past the budget, and the new output-size check means an op whose eval expands its inputs is no longer folded during inference even when the inputs themselves are small.

`cargo fmt --all --check` and `cargo clippy --workspace` clean. `tract-core` 268, `tract-hir` 63, `test-unit-core` 810, `test-onnx-core` 4428, `test-nnef-cycle` 2412, all passing. Two tests added next to the pass; the first fails on main.

I left the `volume() < 16` fast path in `TypedModel::wire_node` alone, since `prop_const` is the fallback there and this asymmetry is the whole reason inference needed a different bound.

🍍

### DIFF
--- a/core/src/optim/mod.rs
+++ b/core/src/optim/mod.rs
@@ -22,6 +22,15 @@ use self::slice::PushSliceUp;
 use self::uniform_mask::FoldUniformMask;
 use op_optim::OpOptim;
 
+/// Memory a single constant fold may materialize.
+///
+/// Bounds every eager evaluation of a constant subgraph, both during inference
+/// and in `prop_const`: a fold under the budget is cheap enough to pay for at
+/// load time, and one over it is left in the graph rather than risk turning a
+/// large weight into several copies. Shape and index arithmetic sits far below
+/// it, decoded weight tensors far above.
+pub const CONST_FOLD_MEM_BUDGET: u64 = 8 << 20;
+
 pub trait TypedPass: Debug + Send + Sync + dyn_clone::DynClone {
     fn reset(&mut self) -> TractResult<()>;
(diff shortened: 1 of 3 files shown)


kali:
Very nice !

PR #2534, opened 2026-07-29
linalg/mmm: decouple parallel chunking from cache blocking

simon0191:
The rayon MMM dispatcher used one number, the chunk size, for two jobs that want
different values: splitting work across threads, and keeping a worker's packed
operands cache-resident. This gives every dispatch chunk the two-tier
cache-blocked tile walk that only the single-thread path had, which frees
`chunk_grid` to track the thread count and to shape chunks for minimal operand
re-reading.

## What was wrong

A chunk covering `dr_m x dr_n` panels reads `dr_m*mr*k` elements of the packed A
and `dr_n*nr*k` of the packed B. Over the whole grid A is therefore read once per
column of chunks and B once per row, so both the number of chunks and their shape
decide how much memory traffic a split costs.

`chunk_grid` had three problems:

1. **Chunk size was the only cache-blocking mechanism.** Chunks were capped at 16
   panels per side so a worker's A and B sub-panels would stay in cache. On a grid
   that is large in both dimensions that produces far more chunks than threads: a
   4096-cube f32 GEMM on 8 threads got 1024 chunks, reading A and B 32 times each.
   Inside a chunk the tiles were walked with a plain nested loop, so the only
   locality came from the chunk being small.

2. **Grids below that cap collapsed to a band.** When the 16-panel grid yielded
   fewer than `4*nth` chunks, the whole grid became exactly `nth` chunks along the
   longer axis. Exactly `nth` chunks leaves rayon nothing to steal, so one slow
   worker (a contended core, or an E-core on a big.LITTLE phone) sets the wall
   clock for the whole GEMM. A band is also the worst shape for re-reads on a
   square-ish grid: at 512x512 panels on 8 threads it costs 1.5x the traffic of a
   2x4 grid with the same chunk count.

3. **Chunk counts could name an empty chunk.** `dr = panels.div_ceil(nchunks)` was
   used together with the *requested* `nchunks`, so the last chunk of an axis could
   start past the end of the grid. A 1x1 panel grid on 2 threads produced two
   chunks, one of them empty.

## What changed

- `for_each_blocked_tile` takes panel ranges instead of counts, and the blocked
  walk (`run_blocked`, previously `run_single_thread_blocked`) runs over one
  rectangle: the whole grid when single-threaded, one chunk per rayon work item.
  Both blocking tiers stay gated on the rectangle's own extents, so a chunk whose
  streamed operand already fits cache walks exactly the naive order, as before.
- `chunk_grid` aims for `CHUNKS_PER_THREAD * nth` chunks and shapes them to
  minimise `nchunks_n*m + nchunks_m*n`, which puts
  `nchunks_m = sqrt(chunks * m / n)`: chunks as square as the operands' extents
  rather than a band across one axis. Chunk counts are recomputed from the edges,
  so no chunk is ever empty.

  `CHUNKS_PER_THREAD` is 4, which is the same granularity the rule it replaces
  already treated as sufficient (its re-plan test was `nchunks < 4*nth`). Going
  lower trades a linear loss of load-balance slack for only a `sqrt` gain in
  re-reads, so it is not worth it. At 8 threads and an 8x8 kernel, in units of
  panel-rows re-read (`nchunks_n*m + nchunks_m*n`):

  | grid (panels) | old | old re-reads | new | new re-reads | new chunks |
  |---|---|--:|---|--:|--:|
  | 16x96 (GLiNER qkv, seq 128) | 1x8 | 224 | 2x16 | 448 | 32 |
  | 32x384 (FFN up, seq 256) | 2x24 | 1536 | 1x32 | 1408 | 32 |
  | 64x64 (512-cube) | 1x8 | 576 | 5x6 | 704 | 30 |
  | 128x128 (1024-cube) | 8x8 | 2048 | 5x6 | 1408 | 30 |
  | 512x512 (4096-cube) | 32x32 | 32768 | 5x6 | 5632 | 30 |

  The two rows that get more re-reads are the ones where the old grid had a single
  chunk per thread; there the extra traffic buys 4x the slack, and it lands on the
  small operand (A is `m*k`, 393 KB for the GLiNER row). The large grids get both
  less traffic and still ~4 chunks per thread.
- The L3 super-block is sized against the share of the last-level cache a
  concurrent walker actually gets, not the whole thing. Without that, N chunks
  each blocking for the full LLC would evict each other's super-blocks.
- `run_with_scratch_space_col_outer` and `_row_outer` were identical apart from one
  boolean, so they merge into `run_with_scratch_space_2d(.., col_outer, ..)`.
  Adding the blocked walk to both would otherwise have duplicated it four times.

## Results are unchanged

Chunking and both blocking tiers only reorder independent tiles, and every output
tile still accumulates the whole of k in a single kernel call, so the output is
bit-identical at any thread count. New harness case
`harness/nnef-test-cases/parallel-matmul` asserts that: a serial reference run
versus 2, 3 and 8 threads at `--approx exact`, over a skewed, a square, a
tall-thin and a batched matmul.

New unit tests cover the grid invariants directly: that the four numbers
`chunk_grid` returns tile the panel grid with no empty, overlapping or missing
chunk; that there are at least as many chunks as threads whenever the grid has
that many panels; and that the chosen shape never costs more re-reading than a
band along either axis at the same chunk count. Each is checked across 14 grid
shapes, 4 kernel aspect ratios and 8 thread counts.

`cargo test -p tract-linalg --features multithread-mm --lib` passes 4406 tests.
Four `arm64::sme::test_sme_qmmm_i32_32x32::store_i8::add_unicast_*` failures and
the `memory-arena` harness case fail identically on this machine with the diff
reverted, so they are pre-existing here and unrelated.

## Measurements, and what I could not measure

I could not get trustworthy multithreaded wall-clock numbers for this. The machine
I have is a 10P+4E Apple M4 Pro, whose heterogeneous cores confound thread scaling
on their own, and it sat at a load average between 30 and 90 throughout. That is
far too noisy for parallel timing: two configurations that were bit-for-bit
identical measured 28% apart.

What I could measure, because it needs only one core and so survives that load, is
the cost of replaying a chunk decomposition sequentially, which captures the
operand re-reading a chunking causes. Against the plain unchunked single-thread
walk, on the NEON 8x8 f32 kernel, the new decomposition costs 1.00x to 1.01x on
GLiNER encoder shapes (128 or 256 x 768 x 768/3072) and 0.68x to 0.90x on
512-cube and 256x768x3072. So it is neutral where the old decomposition was
already fine and cheaper where it was not. The large-square case stayed
inconclusive under the load.

The case above is arithmetic about re-read counts, not a measured speedup, so
wall-clock validation on a quiet many-core box is the one thing I would like
checked before this lands. A `/ci bench` run would cover it, and the multi-threaded
shapes are where I would look first.

## One limitation I want your read on

The L3 tier divides its budget by the number of concurrent walkers, but the L2 tier
does not: `inner_block_edge` budgets the whole detected L2 per rectangle. That is
right where L2 is private per core (Neoverse V1's 1 MB, most current server and
mobile Arm), and wrong where L2 is shared across a cluster, since sibling chunks
would then each budget the same bytes and could evict one another. `CacheInfo::l2`
reports either and cannot tell them apart, and its own doc says "per perf-core /
cluster", so I could not gate this on detection.

I left the behaviour undivided and documented the assumption rather than guessing,
because dividing unconditionally would under-block the private-L2 parts this change
is aimed at. Cortex-A7/A9 in the nightly bench matrix are the cluster-shared-L2
cases, and `bench-suite` does not enable `multithread-mm`, so those runs would not
catch it either way. You know that hardware matrix far better than I do, so tell me
if you would rather divide L2 too, or gate it on a new topology probe.

## Found on the way

`for_each_blocked_tile` overflowed when a range did not start at zero and blocking
was off (`blk == usize::MAX`), because `jb + blk` wrapped. Only the new per-chunk
caller can reach that (the single-thread walk always started at zero), so nothing
shipped with it, but the walk tests now cover `usize::MAX` edges on offset
rectangles.

Separately, and not part of this PR: the SVE kernels are gated on FEAT_SVE2, but
the f32 GEMM and GEMV kernels use only base SVE intrinsics (`svld1_f32`,
`svmla_n_f32_x`, `svst1_f32`, `svwhilelt_b32`, `svcntw`). Neoverse V1 implements
SVE but not SVE2, so it runs plain NEON today. That looks worth its own issue,
though V1's 2x256-bit SVE and 4x128-bit NEON have the same peak FMA throughput, so
it would be an instruction-count win rather than a throughput one.

🍍

### DIFF
--- a/harness/nnef-test-cases/parallel-matmul/graph.nnef
+++ b/harness/nnef-test-cases/parallel-matmul/graph.nnef
@@ -1,0 +1,30 @@
+version 1.0;
+
+# Exercises the multithreaded MMM chunked dispatch at shapes that make it pick
+# different chunk grids: a skewed grid (n much longer than m, so the chunks band
+# along n), a square one (a genuinely 2D grid), a tall-thin one, and a batched
+# matmul (one dispatch per prefix coordinate). Every shape stays above the default
+# 64-panel threading threshold even on a 32x32-tile kernel (the widest tile any
+# f32 mmm kernel here uses, e.g. Apple AMX and SME), so all four reach the
+# parallel path without tuning the threshold: 96, 256, 128 and 96 panels
+# respectively.
+#
+# runme.sh asserts the threaded output is bit-identical to the same-platform
+# single-threaded run: chunking and the in-chunk cache-blocked walk only reorder
+# independent tiles, and each output tile still accumulates the whole k in one
+# kernel call.
+graph parallel_matmul(skew_a, skew_b, sq_a, sq_b, thin_a, thin_b, batch_a, batch_b) -> (skew, sq, thin, batch)
+{
+    skew_a = external<scalar>(shape = [128, 768]);
+    skew_b = external<scalar>(shape = [768, 768]);
+    sq_a = external<scalar>(shape = [512, 512]);
+    sq_b = external<scalar>(shape = [512, 512]);
+    thin_a = external<scalar>(shape = [1024, 256]);
+    thin_b = external<scalar>(shape = [256, 128]);
+    batch_a = external<scalar>(shape = [4, 256, 256]);
+    batch_b = external<scalar>(shape = [4, 256, 384]);
+    skew = matmul(skew_a, skew_b);
+    sq = matmul(sq_a, sq_b);
+    thin = matmul(thin_a, thin_b);
+    batch = matmul(batch_a, batch_b);
+}
--- a/harness/nnef-test-cases/parallel-matmul/runme.sh
+++ b/harness/nnef-test-cases/parallel-matmul/runme.sh
@@ -1,0 +1,33 @@
+#!/bin/sh
+
+cd `dirname $0`
+set -ex
+
+: ${TRACT_RUN:=cargo run -p tract-cli $CARGO_OPTS --}
+
+trap 'rm -f ref.npz' EXIT
+
+# --allow-random-input is seeded from a fixed constant, so both runs below get
+# byte-identical input without committing a bundle.
+#
+# Serial reference on this same binary/platform (the default executor is serial).
+$TRACT_RUN . -O run --allow-random-input --save-outputs ref.npz
+
+# The chunked MMM dispatch splits the panel grid across threads and walks each
+# chunk cache-blocked; both only reorder independent tiles, so the threaded
+# result must be bit-identical to the serial one (--approx exact = atol=rtol=0).
+# Comparing against a same-platform serial run keeps this independent of which
+# kernel the host registers.
+#
+# A default build (including the CI cli-tests harness) has no multithread-mm, so
+# `--threads` bails; drive this case with a multithread-mm TRACT_RUN (the plan's
+# verification step does) to exercise the parallel leg. We warn rather than skip
+# silently so a serial-only run is never mistaken for threaded coverage.
+if $TRACT_RUN --threads 2 . dump -q >/dev/null 2>&1; then
+    for threads in 2 3 8; do
+        $TRACT_RUN --threads $threads . -O run \
+            --allow-random-input --assert-output-bundle ref.npz --approx exact
+    done
+else
+    echo "WARNING: TRACT_RUN has no multithread-mm; parallel bit-identity leg NOT run." >&2
+fi
--- a/linalg/Cargo.toml
+++ b/linalg/Cargo.toml
@@ -84,6 +84,11 @@ name = "mm_for_asr_am"
 harness = false
 
 [[bench]]
+name = "mm_threads"
+harness = false
+required-features = ["multithread-mm"]
+
+[[bench]]
 name = "qmmm_i8"
 harness = false
 
--- a/linalg/benches/mm_threads.rs
+++ b/linalg/benches/mm_threads.rs
@@ -1,0 +1,111 @@
+//! Wall-clock scaling of the parallel MMM dispatcher across thread counts.
+//!
+//! Exercises the chunk-grid / cache-blocking split: each shape is run under a
+//! rayon pool of 1, 2, 4 and 8 threads. t1 is the regression guard (the serial
+//! path must stay neutral); the larger squares are where re-read traffic and
+//! load-balance slack decide multi-thread scaling.
+//!
+//! Compare two revisions with criterion baselines:
+//!   cargo bench -p tract-linalg --features multithread-mm --bench mm_threads -- --save-baseline main
+//!   # switch revision
+//!   cargo bench -p tract-linalg --features multithread-mm --bench mm_threads -- --baseline main
+//!
+//! `MM_THREADS` overrides the thread sweep (comma list), `MM_DT` picks `f32`
+//! (default) or `i8`.
+
+use criterion::*;
+use tract_data::internal::*;
+use tract_linalg::mmm::{AsInputValue, FusedSpec};
+use tract_linalg::multithread::{Executor, multithread_tract_scope};
+
+use DatumType::*;
+
+/// (label, m, k, n). Element dims of the shapes in the PR's re-read table,
+/// plus a tall-thin and a batched-ish wide case mirroring the harness cases.
+const SHAPES: &[(&str, usize, usize, usize)] = &[
+    ("gliner_qkv_128", 128, 768, 768), // 16x96 panels @ 8x8
+    ("ffn_up_256", 256, 768, 3072),    // 32x384 panels
+    ("sq512", 512, 512, 512),          // 64x64
+    ("sq1024", 1024, 1024, 1024),      // 128x128
+    ("sq2048", 2048, 2048, 2048),      // interpolation point
+    ("sq4096", 4096, 4096, 4096),      // 512x512 — headline re-read case
+    ("tall_thin", 4096, 512, 128),     // skewed
+    ("wide", 128, 512, 4096),          // skewed the other way
+];
+
+fn threads() -> Vec<usize> {
+    match std::env::var("MM_THREADS") {
+        Ok(s) => s.split(',').filter_map(|t| t.trim().parse().ok()).collect(),
+        Err(_) => vec![1, 2, 4, 8],
+    }
+}
+
+fn dt() -> DatumType {
+    match std::env::var("MM_DT").as_deref() {
+        Ok("i8") => I8,
+        _ => F32,
+    }
+}
+
+fn bench_shape(c: &mut Criterion, dt: DatumType, label: &str, m: usize, k: usize, n: usize) {
+    let mut group = c.benchmark_group(format!("{label}/{m}x{k}x{n}"));
+    group.throughput(Throughput::Elements((m * k * n) as u64));
+
+    let mmm = tract_linalg::ops().mmm(dt, Some(m), Some(k), Some(n)).unwrap();
+    let a = Tensor::zero_dt(dt, &[m, k]).unwrap();
+    let b = Tensor::zero_dt(dt, &[k, n]).unwrap();
+    let packing = &mmm.packings()[0];
+    let pa = packing.0.prepare_one(&a, 1, 0).unwrap();
+    let pb = packing.1.prepare_one(&b, 0, 1).unwrap();
+    let mut cbuf = Tensor::zero_dt(dt, &[m, n]).unwrap();
+
+    for &nth in &threads() {
+        // A tract pool with <= 1 worker is treated by the dispatcher as "use the
+        // global rayon pool" (mod.rs), so `multithread(1)` would silently run on
+        // every core. The genuine serial path is `SingleThread`.
+        let pool = if nth <= 1 { Executor::SingleThread } else { Executor::multithread(nth) };
+        group.bench_function(BenchmarkId::from_parameter(format!("t{nth}")), |be| {
+            let mut scratch = unsafe { mmm.allocate_scratch_space() };
+            be.iter(|| {
+                multithread_tract_scope(pool.clone(), || unsafe {
+                    mmm.run_with_scratch_space(
+                        m,
+                        n,
+                        scratch.as_mut(),
+                        &[
+                            FusedSpec::AddMatMul {
+                                a: AsInputValue::Borrowed(&*pa),
+                                b: AsInputValue::Borrowed(&*pb),
+                                packing: 0,
+                            },
+                            FusedSpec::Store(mmm.c_view(Some(0), Some(1)).wrap(&cbuf.view_mut())),
+                        ],
+                    )
+                    .unwrap()
+                })
+            })
+        });
+    }
+    group.finish();
+}
+
+fn benches(c: &mut Criterion) {
+    let dt = dt();
+    for &(label, m, k, n) in SHAPES {
+        bench_shape(c, dt, label, m, k, n);
+    }
+}
+
+fn config() -> Criterion {
+    Criterion::default()
+        .sample_size(10)
+        .warm_up_time(std::time::Duration::from_millis(800))
+        .measurement_time(std::time::Duration::from_secs(4))
+}
+
+criterion::criterion_group! {
+    name = benches_group;
+    config = config();
+    targets = benches
+}
+criterion::criterion_main!(benches_group);
--- a/linalg/src/cache.rs
+++ b/linalg/src/cache.rs
@@ -25,6 +25,12 @@ pub struct CacheInfo {
     pub l2: usize,
     /// L3 / last-level cache, bytes. `0` if unknown.
     pub l3: usize,
+    /// How many physical cores share one L2 (1 == private per core, as on most
+    /// server/mobile Arm and x86). Greater than 1 on cluster-shared-L2 parts
+    /// (Cortex-A9/A53). `0` if the topology could not be read — callers treat
+    /// that as private. SMT siblings do not count: they already share the core's
+    /// L2, so a 2-thread core with a private L2 reports 1, not 2.
+    pub l2_sharers: usize,
 }
 
 impl CacheInfo {
@@ -43,6 +49,12 @@ impl CacheInfo {
     /// L2 cache, or a conservative 256 KiB guess when undetected.
     pub fn l2_or_default(&self) -> usize {
         if self.l2 > 0 { self.l2 } else { 256 * 1024 }
+    }
+
+    /// Physical cores sharing one L2, at least 1 — unknown topology (`0`) reads
+    /// as private, the regression-safe assumption (no shared-cache division).
+    pub fn l2_sharers_or_one(&self) -> usize {
+        self.l2_sharers.max(1)
     }
 }
 
@@ -239,27 +251,54 @@ fn detect() -> CacheInfo {
         l3: sysctl_usize("hw.perflevel0.l3cachesize")
             .or_else(|| sysctl_usize("hw.l3cachesize"))
             .unwrap_or(0),
+        // Apple L2 is per-cluster (shared across a perflevel's cores), but sysctl
+        // does not expose the sharing degree; report unknown (treated as private).
+        l2_sharers: 0,
     }
+}
+
+/// Count the CPUs named by a Linux cpu-list string (`"0-3"`, `"0,8"`,
+/// `"0-3,8-11"`). Malformed fields are skipped, so a garbled file counts 0.
+#[cfg_attr(not(any(target_os = "linux", target_os = "android")), allow(dead_code))]
+fn count_cpu_list(s: &str) -> usize {
+    s.split(',')
+        .filter_map(|part| {
+            let part = part.trim();
+            if part.is_empty() {
+                return None;
+            }
+            match part.split_once('-') {
+                Some((a, b)) => {
+                    let a: usize = a.trim().parse().ok()?;
+                    let b: usize = b.trim().parse().ok()?;
+                    (b >= a).then_some(b - a + 1)
+                }
+                None => part.parse::<usize>().ok().map(|_| 1),
+            }
+        })
+        .sum()
 }
 
 #[cfg(any(target_os = "linux", target_os = "android"))]
 fn detect() -> CacheInfo {
     // Walk /sys/.../cache/indexN, keying off the reported level+type rather than
     // assuming a fixed index layout (it varies: SMT, unified vs split L2, …).
+    let read = |p: String| std::fs::read_to_string(p).ok();
     let mut ci = CacheInfo::default();
+    // SMT siblings share the core's L2 already; only cores beyond that set count
+    // as L2-sharing. Absent topology ⇒ assume no SMT (1).
+    let smt = read("/sys/devices/system/cpu/cpu0/topology/thread_siblings_list".to_string())
+        .map(|s| count_cpu_list(&s))
+        .filter(|&n| n > 0)
+        .unwrap_or(1);
     for idx in 0..16 {
         let base = format!("/sys/devices/system/cpu/cpu0/cache/index{idx}/");
-        let Ok(level) = std::fs::read_to_string(format!("{base}level")) else {
+        let Some(level) = read(format!("{base}level")) else {
             continue;
         };
         let level: usize = level.trim().parse().unwrap_or(0);
-        let ctype = std::fs::read_to_string(format!("{base}type"))
-            .unwrap_or_default()
-            .trim()
-            .to_ascii_lowercase();
-        let size = std::fs::read_to_string(format!("{base}size"))
-            .map(|s| parse_cache_size(&s))
-            .unwrap_or(0);
+        let ctype = read(format!("{base}type")).unwrap_or_default().trim().to_ascii_lowercase();
+        let size = read(format!("{base}size")).map(|s| parse_cache_size(&s)).unwrap_or(0);
         if size == 0 {
             continue;
         }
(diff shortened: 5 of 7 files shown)


kali:
You're right about MT not being exercised by the bench suite at all. MT in tract is a recent addition, and it has often been a underwhelming. And our on)device production settings are all ST for tract.

Maybe it would be time to introduce a MT bench though. 

@czoli1976 any insights on this ?

czoli1976:
a few things from benching MT on tract over the past months, native and WASM.

The bench binary is ST by construction. bench.yml builds --features bench-suite; multithread-mm is a separate feature and --threads bails without it. So an MT-touching PR is dormant in the bench today and "neutral" is guaranteed rather than measured. (Same failure mode as #2253, which benched neutral because the op was wasm-gated and the bench ran native.) An MT bench needs its own build leg, not just a new [[bench]] entry.

Scaling and E2E are very different stories, and only the second one is a bench.

WASM, 4 threads (linalg/MULTITHREAD_BENCHMARKS.md): dense 1024³ 3.34× Chromium / 3.12× WebKit / 3.40× Firefox; DFN3 full streaming 1.03–1.08×, i.e. noise. Only ~25% of DFN3 is in MMMs above the panel gate; the rest is FFT and small RNN matmuls that stay serial by design.
Native M1 Pro, parakeet encoder S=512: --threads 6 = 1.64× E2E (0.335 → 0.204 s), out of ~5× op-level scaling. Plateau at 6 P-cores; E-cores add ~0 and thread 7 dips.
So an MT bench made of DFN3-class models will report noise forever. It needs at least one dense-dominated case (parakeet encoder at S≥512, or an LLM PP) to have signal at all.

Heterogeneous cores are the main noise source, and that part is fixable by policy. --threads 0 guesses physical cores; on a 10P+4E box that puts E-core workers in the same rayon pool and the numbers move a lot. Every MT case should carry an explicit --threads N with N = that host's P-core count, recorded in the metric label — otherwise the reference series stops being comparable the moment a runner changes.

4Record the ST/MT ratio for the same binary, not just the MT absolute. The ratio is what actually regresses when chunking breaks, and it is far more robust to load and thermals than the absolute. Related: bench-thresholds.toml's k×historical-noise will auto-widen MT series — right in principle, but it means small MT regressions will never flag. A scaling-ratio metric with a tighter floor gates better than a wall-clock one.

 Include a case where the correct answer is "don't thread". I chased an isolated GEMV under-threading finding (2.5–4.8× apparently available at m=2048, gated off by THREADING_PANEL_THRESHOLD = 64) all the way to e2e and it was a clean negative: lowering the gate regressed every real decode workload by 5–78%. Real decode is many tiny per-layer GEMVs; ~5 µs of rayon dispatch per op swamps µs-scale compute, and threading f16 GEMV additionally serialises on the shared AMX unit. The gate at 64 is correctly calibrated. An MT bench without a decode/streaming case would happily score a gate-lowering change as a win.


Bit-exactness across thread counts is cheap and worth gating separately from timing. I have run into issues with SVT-AV1 (Another Repo) with outputs not bite identical with MT so it is worth double checking (that is a video codec but I think it is worth checking)  In the WASM run I hashed the output tensor (FNV-1a) across 60 cells × 1–4 threads, all identical. This PR's parallel-matmul harness case does the same job at --approx exact, which I think is the better home for it: cheaper and more sensitive than any timing bench.

WASM MT, if it comes up: not cheap to put in CI. It needs +atomics,+bulk-memory,+mutable-globals with -Z build-std, wasm-bindgen-rayon, COOP/COEP headers and a browser driver — rayon::ThreadPoolBuilder::build() cannot work on wasm32-unknown-unknown, which is why Executor::RayonGlobal exists in the first place. I would keep an MT bench native-only.

czoli1976:
Reviewing and Confirming on my machines ....

czoli1976:
You asked for wall-clock validation on a quiet box before this lands, so here it is — with the caveat up front that I could only half-deliver on "many-core".

**Machines.** Timing is all from a MacBook Air M4 (`Mac16,12`, 4P+6E, macOS 26.5.1, on AC), load average 1.5–2.1 throughout — quiet, but 4 P-cores is not the many-core box you wanted, so this doesn't fully discharge the ask. My M1 Pro was sitting at load 143 while I did this, so I used it only for correctness and didn't time anything on it; your problem exactly.

Method: same packed operands, same NEON 8x8 f32 kernel, only the thread count varies. Min of 9 timed batches, two independent passes, `tract_linalg::ops().mmm()` driven directly so nothing but the dispatcher is in the measurement. Absolute ms, PR vs `7ff6ecfbe`:

| shape (m×k×n) | t1 | t4 | t6 | t8 | t10 |
|---|--:|--:|--:|--:|--:|
| 128×768×768 (GLiNER qkv s128) | 0.113 → 0.075 | −19% | −12% | **−32%** | −19% |
| 256×768×768 (qkv s256) | 0.150 → 0.150 | −15% | −17% | −4% | **−28%** |
| 256×768×3072 (FFN up) | 0.653 → 0.650 | −8% | −8% | — | −3% |
| 512 cube | 0.134 → 0.134 | −16% | −21% | +2% | **−29%** |
| 1024 cube | 1.069 → 1.069 | −4% | −2% | −6% | −7% |
| 2048 cube | 17.78 → **15.93** | −0.1% | −3% | −0.7% | **+5.3%** |
| 64×4096×4096 (tall-thin) | 2.02 → 1.99 | **−22%** | −20% | **−23%** | −21% |
| 4096×4096×64 (wide-flat) | 2.01 → 1.99 | −4% | −6% | — | −1% |

Net: faster on 7 of 8 shapes at essentially every thread count. Two things stand out.

**The band fix is the big win, and it's exactly your problem #2.** tall-thin is −20 to −23% across the board. Replaying both `chunk_grid`s at 8×512 panels: old gives `1×10` = 10 chunks for 10 threads (1.0 per thread, nothing to steal), new gives `1×40` (4.0 per thread). That's the mechanism you described, and it converts to wall clock cleanly.

**The 2048 cube regresses ~5% at full thread count, reproducibly** (both passes, +5.3% and +4.9%). Same arithmetic explains it, running the other way:

| grid | nth | old chunks | new chunks |
|---|--:|--:|--:|
| 256×256 panels (2048 cube) | 10 | 16×16 = **256** (25.6/thread) | 6×6 = **36** (3.6/thread) |
| 256×256 panels | 4 | 256 (64/thread) | 4×4 = 16 (4.0/thread) |

Your trade-off argument — linear slack against a `sqrt` traffic gain — runs symmetrically. For a grid this large the old cap wasn't giving 4 chunks per thread, it was giving 25; dropping to 3.6 on a 4P+6E box means the tail is set by whichever E-core draws one of only 36 chunks. Note it's specifically t10 (all E-cores engaged): at t4, where all workers are P-cores, the PR is at parity. So `CHUNKS_PER_THREAD = 4` reads to me as the right *floor* but too coarse a *ceiling* for grids that are large in both dimensions — where the extra chunks are cheap precisely because the `sqrt` shaping already keeps re-reads down.

Worth noting the other side of that row: single-thread 2048 cube is **17.78 → 15.93 ms, −10%**, from the blocked walk now covering the whole grid. That's a real win the PR description undersells.

**On your L2 question — Apple Silicon is the cluster-shared case, and there is a probe.** `hw.perflevel0.cpusperl2` reports 4 on this M4 (four P-cores sharing one 16 MB L2) and 3 on my M1 Pro (two clusters of three). So macOS exposes cores-per-L2 directly; Linux has the same via `/sys/devices/system/cpu/cpuN/cache/indexN/shared_cpu_list`. Empirically your call to leave it undivided looks fine: at t4 — all four workers on one shared L2, so budgeting 4× the real share — the PR is at parity or better on every shape, and it's only at t10 that anything regresses, which the chunk arithmetic above already accounts for. I'd keep it undivided and not add a probe on this evidence, but the probe exists if you ever want it.

**Bit-exactness** — I checked it independently rather than trusting the harness case, since I've been bitten by MT non-determinism elsewhere. Real non-zero data (sin/cos-filled A and B, not zeros), full output tensor compared byte-for-byte against the 1-thread run: 7 shapes × {2,3,4,6,8,10} threads, all identical. Your `parallel-matmul` case is the right home for this, and it holds.

**Also confirmed:** `cargo test -p tract-linalg --features multithread-mm --lib` gives 4410 passed / 0 failed on the M1 Pro (= your 4406 plus the four `arm64::sme::…::add_unicast_*` that only appear on SME silicon — I reproduced those separately on the M4 as pre-existing on `main` too, same four, see #2537). The three new `chunk_grid` invariant tests pass.

One caveat on all of the above: MT scaling on this hardware is poor in absolute terms — best case 2.5× on 10 threads, and several shapes are *slower* threaded than serial on both branches. That's pre-existing and matches what I posted earlier in this thread about E-cores. The PR consistently reduces that penalty (256×768×768 at t10 goes from 0.65× to 0.90× of serial), which is arguably the more important result than the raw wins.

I'll re-run the whole set on the M1 Pro (6P+2E) once it's quiet — I've got rather too much going on there at the moment. That's a different big.LITTLE split, 6 P-cores instead of 4, so it should say whether the 2048-cube tail effect is really about E-cores joining the pool or just about this particular machine. Happy to add other shapes or thread counts to that run if there are any you'd want covered.

---

<details>
<summary><b>Full review brief</b> — structured notes if you want to hand this to an assistant: measurement method, the chunk-count arithmetic, and a list of things I checked that should <i>not</i> be changed.</summary>

# PR #2534 — independent review brief

Context for an assistant working on `sonos/tract` PR #2534 (`linalg/mmm: decouple
parallel chunking from cache blocking`), head `a39848cad`, base `7ff6ecfbe`.

Independent replication of the PR's claims, plus the multithreaded wall-clock
measurement the PR description explicitly asks for. Background evidence, not
instructions from the maintainer. Nothing has been merged or pushed.

**There is one substantive finding (§3) and one answered open question (§4).**
Everything else confirms the PR is correct. §6 lists things that were checked and
should *not* be changed.

## Hardware, and why it matters here

* **Timing — MacBook Air M4** (`Mac16,12`), 4 P + 6 E cores, macOS 26.5.1, AC power,
  load average 1.5–2.1 throughout. Quiet, but **only 4 P-cores** — this is not the
  "quiet many-core box" the PR asks for, and does not fully discharge that request.
* **Correctness only — MacBook Pro 16"** (`MacBookPro18,3`), M1 Pro, 6 P + 2 E. It
  was at load average 143 during this work, so **nothing was timed on it**. Same
  failure mode the PR description reports.

Both machines are heterogeneous (big.LITTLE). That is directly relevant to §3: a
regression that only appears once E-cores join the pool.

Measurement method: `tract_linalg::ops().mmm()` driven directly with pre-packed
operands, so only the dispatcher varies. Min of 9 timed batches, two independent
passes. f32, NEON 8x8 kernel.

---

## 1. Verified — the PR's claims hold

| claim | result |
| --- | --- |
| `cargo test -p tract-linalg --features multithread-mm --lib` passes | 4410 passed, 0 failed on M1 Pro (= the PR's 4406 + the 4 SME tests that only run on SME silicon) |
| new `chunk_grid` invariant tests | all 3 pass (`tiles_the_panel_grid`, `feeds_every_thread`, `shape_beats_a_band_on_operand_traffic`) |
| 4 `arm64::sme::…::add_unicast_*` failures are pre-existing | confirmed on an M4 — identical 4 fail on `main` and on the PR (see PR #2537 for the same check) |
| output bit-identical at any thread count | **confirmed independently**, see §2 |
| chunking fixes the degenerate band | **confirmed in wall clock**, see §3 |

`chunk_grid` was also read closely. The `clamp(1, n_panels_m)` cannot panic: the only
caller (`chunked_dispatch_rayon`) returns early when either panel count is 0, before
the call. The recount-from-edges at the end is sound — `dr = ceil(p/c)` implies
`ceil(p/dr) <= c`, so the returned counts never name an empty chunk and never exceed
the requested count.

## 2. Bit-exactness — independently confirmed

Not taken on trust from the `parallel-matmul` harness case. A separate check ran the
MMM with **real non-zero data** (sin/cos-filled A and B — zero-filled operands would
make this test vacuous) and compared the **full output tensor byte-for-byte** against
the single-threaded run:

7 shapes × {2, 3, 4, 6, 8, 10} threads → **all bit-identical**.

Shapes: 256×768×768, 256×768×3072, 512³, 1024³, 2048³, 64×4096×4096, 4096×4096×64.

## 3. FINDING — `CHUNKS_PER_THREAD = 4` is too coarse for large grids

### The measurement

Absolute ms, PR vs base, min of 9 × 2 passes:

| shape | t1 | t4 | t6 | t8 | t10 |
|---|--:|--:|--:|--:|--:|
| 128×768×768 | 0.113 → 0.075 | −19% | −12% | −32% | −19% |
| 256×768×768 | 0.150 → 0.150 | −15% | −17% | −4% | −28% |
| 256×768×3072 | 0.653 → 0.650 | −8% | −8% | — | −3% |
| 512 cube | 0.134 → 0.134 | −16% | −21% | +2% | −29% |
| 1024 cube | 1.069 → 1.069 | −4% | −2% | −6% | −7% |
| **2048 cube** | 17.78 → **15.93** | −0.1% | −3% | −0.7% | **+5.3%** |
| 64×4096×4096 | 2.02 → 1.99 | −22% | −20% | −23% | −21% |
| 4096×4096×64 | 2.01 → 1.99 | −4% | −6% | — | −1% |

Net strongly positive: 7 of 8 shapes faster at essentially every thread count. Two
results deserve attention.

**The band fix converts.** 64×4096×4096 is −20 to −23% everywhere. Replaying both
`chunk_grid` implementations at 8×512 panels, 10 threads: old → `1×10` = 10 chunks
(1.0 per thread, nothing for rayon to steal); new → `1×40` (4.0 per thread). Exactly
the PR's stated problem #2, and it shows up in wall clock.

**The 2048 cube regresses at full thread count**, +5.3% and +4.9% across two passes,
so not noise. Same arithmetic, running the other way:

| grid | nth | old | new |
|---|--:|--:|--:|
| 256×256 panels | 10 | 16×16 = 256 chunks (25.6/thread) | 6×6 = 36 chunks (3.6/thread) |
| 256×256 panels | 4 | 256 chunks (64/thread) | 4×4 = 16 chunks (4.0/thread) |

The PR justifies `CHUNKS_PER_THREAD = 4` as "the same granularity the rule it
replaces already treated as sufficient (its re-plan test was `nchunks < 4*nth`)".
That holds for grids at or below the old 16-panel cap, but for a grid this large the
old code was not producing 4 chunks per thread — it was producing 25. The trade the
PR states ("a linear loss of load-balance slack for only a `sqrt` gain in re-reads")
runs symmetrically: on grids that are large in both dimensions, *more* chunks buy
linear slack for only `sqrt` more traffic.

The regression is specific to **t10**, where the 6 E-cores join. At t4 (all workers
on P-cores) the PR is at parity. That points at load-balance tail, not memory
traffic: with 36 chunks and 10 heterogeneous workers, whichever E-core draws a late
chunk sets the wall clock.

### Suggested direction — not a patch to apply blindly

`CHUNKS_PER_THREAD = 4` is a sound *floor*. What is missing is a ceiling that lets
large grids keep more slack, e.g. scaling the target chunk count with the panel grid
size rather than with `nth` alone, so a 256×256 grid gets meaningfully more than 36
chunks at 10 threads while a 16×96 grid keeps today's behaviour. The `sqrt` shaping
already keeps the re-read cost of extra chunks sub-linear, which is what makes this
cheap.

Do **not** simply raise the constant — that would re-coarsen the small and banded
grids this PR fixes, which are its main win. Any change here needs re-measuring at
both ends (64×4096×4096 and 2048³) at several thread counts.

Caveat: measured on 4P+6E only. A homogeneous many-core box may not show this at
all, since it is a heterogeneity/tail effect. Worth confirming before acting.

## 4. ANSWERED — the L2 shared-vs-private question the PR raises

The PR says it left `inner_block_edge` budgeting the whole detected L2 per rectangle,
undivided, because `CacheInfo::l2` cannot distinguish private from cluster-shared,
and asks for a read on it.

**Apple Silicon is the cluster-shared case, and a topology probe does exist:**

* macOS: `sysctl hw.perflevel0.cpusperl2` → **4** on the M4 (four P-cores sharing one
  16 MB L2), **3** on the M1 Pro (two clusters of three). `hw.perflevel1.cpusperl2`
  gives the E-cluster.
* Linux: `/sys/devices/system/cpu/cpuN/cache/indexN/shared_cpu_list` gives the same.

**Empirically the undivided behaviour is fine on exactly that hardware.** At t4 every
worker is a P-core sharing one L2, so each chunk budgets 4× its true share — the
worst case for this assumption — and the PR is at parity or better on all 8 shapes
there. The only regression is at t10 and is explained by chunk count (§3), not by L2.

Recommendation: **keep L2 undivided**. The evidence does not support adding a probe,
and dividing would under-block the private-L2 targets the change is aimed at. The
probe is available if it is ever wanted.

## 5. Not measured — what this review does not cover

* **A genuinely many-core box.** 4 P-cores is the ceiling here. Thread scaling above
  ~4 real cores, and the behaviour of §3 on a homogeneous 32-core machine, are open.
* **End-to-end model timing.** All numbers are dispatcher-level. Note that
  `bench-suite` does not build `multithread-mm`, so the nightly bench cannot see this
  PR at all — "no speed regressions" in the bot comment is guaranteed, not measured.
* **Non-f32 datatypes and non-NEON kernels.** f32 / arm64 8x8 only.
* **WASM.** `rayon::ThreadPoolBuilder::build()` does not work on
  `wasm32-unknown-unknown`; that path goes through `Executor::RayonGlobal` and was
  not exercised.

## 6. DO NOT CHANGE — checked and correct

1. **`chunk_grid`'s `clamp(1, n_panels_m)`.** Cannot panic; the caller guards both
   panel counts against 0 first. Do not add a redundant check.
2. **The recount-from-edges at the end of `chunk_grid`.** Correct, and it is what
   fixes the empty-chunk bug the PR describes as problem #3.
3. **Leaving the L2 tier undivided.** See §4 — supported by measurement on
   shared-L2 hardware.
4. **Bit-exactness.** Independently verified with real data across 7 shapes and 6
   thread counts. Do not add defensive reordering or accumulation changes.
5. **`CHUNKS_PER_THREAD = 4` as a floor.** The small and banded grids are where this
   PR wins biggest (−20 to −23% on 64×4096×4096). §3 is about adding a ceiling for
   large grids, not about raising this number.

## 7. Reproduction notes

* Build: `cargo build -p tract-linalg --release --features multithread-mm`. The
  feature is **not** on by default and `--threads` is a no-op without it.
* Drive threads from a test binary with
  `tract_linalg::multithread::multithread_tract_scope(Executor::multithread(n), || …)`;
  `Executor::SingleThread` for the 1-thread reference.
* Pre-pack operands once outside the timed loop via
  `mm.packings()[0].0.prepare_one(&a, 1, 0)` / `.1.prepare_one(&b, 0, 1)`, then call
  `mm.run_with_scratch_space(m, n, scratch, &[FusedSpec::AddMatMul{..}])`. Otherwise
  packing dominates and the dispatcher change is invisible.
* Use **non-zero** operands for any bit-exactness check — zero-filled tensors make it
  vacuous.
* Report the ST/MT ratio for the same binary alongside absolute ms. Note the ratio can
  move the *wrong* way when the PR is better: 2048³ single-thread improves 10%, which
  compresses its apparent scaling.
* On heterogeneous Apple hardware, pin `--threads` to the P-core count
  (`sysctl hw.perflevel0.physicalcpu`) for the least noisy signal, and record it.

</details>

kali:
```
i9-11900KB, 8 P-cores, private 1.25 MB L2, shared 24 MB L3, powersave. Δ = PR vs
  pre-#2534, + = slower. Squares show run1(run2); control = base-vs-base noise floor.

  ┌────────────────────────┬──────┬──────┬─────────────┬─────────────┬────────────┐
  │     shape (M×K×N)      │  t1  │  t2  │     t4      │     t8      │ control t8 │
  ├────────────────────────┼──────┼──────┼─────────────┼─────────────┼────────────┤
  │ gliner_qkv 128×768×768 │ +0.0 │ +0.0 │        +4.0 │        +4.8 │          — │
  ├────────────────────────┼──────┼──────┼─────────────┼─────────────┼────────────┤
  │ ffn_up 256×768×3072    │ +1.1 │ +0.4 │        +0.0 │        +6.4 │          — │
  ├────────────────────────┼──────┼──────┼─────────────┼─────────────┼────────────┤
  │ sq512                  │ +1.9 │ +0.0 │        +5.0 │        +0.0 │          — │
  ├────────────────────────┼──────┼──────┼─────────────┼─────────────┼────────────┤
  │ sq1024                 │ +0.1 │ +0.0 │   +5.7(5.2) │   +5.9(5.9) │       +0.0 │
  ├────────────────────────┼──────┼──────┼─────────────┼─────────────┼────────────┤
  │ sq2048                 │ +0.0 │ +0.0 │  +8.9(11.1) │  +9.6(11.7) │       +0.0 │
  ├────────────────────────┼──────┼──────┼─────────────┼─────────────┼────────────┤
  │ sq4096                 │ +0.0 │ +0.0 │ +17.2(17.9) │ +20.1(19.1) │       +0.4 │
  ├────────────────────────┼──────┼──────┼─────────────┼─────────────┼────────────┤
  │ tall_thin 4096×512×128 │ +1.5 │ +0.0 │        +0.2 │        +0.0 │          — │
  ├────────────────────────┼──────┼──────┼─────────────┼─────────────┼────────────┤
  │ wide 128×512×4096      │ +1.9 │ +0.0 │        +0.0 │        +4.7 │          — │
  └────────────────────────┴──────┴──────┴─────────────┴─────────────┴────────────┘
```

kali:
I'm going to rebase over main to take advantage (maybe) of the branch new multithread testing.

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali:
The multithread bench bit was not merged. I'll re-rebase tomorrow, hopefully we will have a reference build then.

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali:
Apologies for the noise, but benches over gha are a bit tricky to setup right.

kali:
I gave a shot at detecting how L2 is shared, see how the small device react.

kali:
The A53 is a harder one. Let's call it a day and merge this one.

PR #2537, opened 2026-07-30
linalg/pack: transpose k-contiguous blocks instead of one store per element

simon0191:
Packing an operand whose `k` axis is contiguous, which is the shape an activation arrives in,
stored one element per call at a destination stride of `r`. This replaces that with a 4x4 blocked
transpose, vectorised on aarch64, which roughly halves `OptMatMulPack` on a DeBERTa-v3
classifier. The packed layout does not change, so model output is bit-identical.

## What changed

`PackedFormat::pack_t` has an arm for `k_stride == 1`. It fed `KInWriter` one element at a time,
and each `write` did a single store plus the writer's per-element state machine (decrement,
compare, branch, pointer fixup). The destination stride is `r`, so that is one scalar store per
element with nothing amortised. For a 256x768 f32 activation it is 196k of them.

The new `pack_k_major` computes the destination address in closed form instead of walking a state
machine:

```
dst[(mn / r) * panel_len + k * r + mn % r] = src[mn * mn_stride + k]
```

and moves the block as 4x4 tiles. Two things matter for speed:

* **The tile.** Four contiguous loads, a register transpose, four contiguous stores. On aarch64
  that is `ld1` / `trn` / `st1` for 32-bit and 16-bit lanes. Everywhere else it is a portable tile;
  on aarch64 LLVM compiles that one into paired 32-bit loads and stores rather than a vector
  transpose, which is why the intrinsics are there. I have no x86 machine, so I have not checked
  what it becomes on SSE.
* **The loop order.** `k` outer, `mn` inner, so each panel is filled strictly front to back.
  mn-outer instead sweeps the whole panel once per group of 4 columns, and the write traffic that
  wastes grows with `r`. Going k-outer is what makes r=24 as fast as r=8: with mn-outer, r=24 was
  1.5x slower than r=8 on the same byte count.

The `k % 4` and `panel_width % 4` tails go one element at a time. `KInWriter` is untouched and is
now the reference the tests compare against.

## Numbers

Llama Prompt Guard 2 86M (mDeBERTa-v3-base, fp32), frozen shapes, 1 thread, Apple M4 Pro. Median
of 3 interleaved runs:

| seq | pack before | pack after | total before | total after |
| --: | --: | --: | --: | --: |
| 128 | 13.3 ms (13.1%) | 6.9 ms (7.3%) | 101.9 ms | 94.7 ms |
| 256 | 29.2 ms (9.8%) | 15.0 ms (5.3%) | 298.2 ms | 282.8 ms |

The two nodes that dominate, summed over the 12 layers at seq 256:

| node | before | after |
| --- | --: | --: |
| `layer.N/output/dense/MatMul.pack_b` | 10.89 ms | 4.72 ms |
| `layer.N/attention/self/MatMul_3.pack_a` | 7.74 ms | 4.45 ms |

All 110 pack nodes got faster. The pack is serial, so the share it takes back should grow with
thread count, but I cannot time multi-thread runs reliably on this machine, so I have not measured
that.

Packing a single block in isolation, same machine:

| block | before | after |
| --- | --: | --: |
| 256x768 f32, r=8 | 55.0 us | 16.2 us |
| 256x768 f32, r=24 | 59.9 us | 17.4 us |
| 256x768 f16, r=32 | 58.7 us | 13.1 us |

Those came from a throwaway criterion bench that is not in this PR. Happy to add
`linalg/benches/pack.rs` if you want it in tree, which would also be useful for the follow-up that
threads this pack.

## Testing

* New `PackKMajorProblem` in `pack.rs`: packs through `pack_segment` and compares byte for byte
  against feeding `KInWriter` element by element, which is the code this replaces. Run for a
  4-byte, a 2-byte and a 1-byte element type so every tile path is covered, over a proptest of
  `r` in {1, 2, 3, 4, 5, 8, 12, 16, 24, 32}, `k` and `mn` up to 40, cropped `k_range` and
  `mn_range`, and varying panel alignment. Plus explicit cases for exact tiles, each tail, `r`
  narrower than one tile, and an `mn_range` running past `mn` (those columns must stay untouched,
  as before).
* Mutation tested. Swapping two stores in the 32-bit tile, the same in the 16-bit tile, reversing
  the portable tile, and an off-by-one in each of the two tail loops: all five are caught.
* The portable path was exercised on its own by building with the aarch64 branches cfg'd out. All
  32 pack tests pass that way too.
* Model output bit-identical before and after, at seq 128 and seq 256.
* `tract-linalg` (4408), `tract-core` (269) and `test-unit-core` (816) pass, and `tract-linalg`
  also passes with `--features complex`. `cargo clippy --workspace` and `cargo fmt --all` clean.

Pre-existing on my machine and unrelated: 4 `arm64::sme::test_sme_qmmm_i32_32x32::store_i8::add_unicast_*`
failures, which fail the same way on `main`.

One detail worth flagging in review: the vector tiles are selected on `size_of` **and** `align_of`
of the element type. Size alone would route `Complex<i16>` and `Complex<f16>`, which are 4 bytes but
2-aligned under the `complex` feature, into a 32-bit-lane load.

## Deliberately not done

* **8x8 tiles.** 4x4 already tops out around 40 GiB/s on the isolated pack here, and 8x8 costs 16
  registers and twice the permutes.
* **A vector tile for 1-byte elements.** It needs 8x8 or 16x16 to pay off, and the hot int8 path is
  `PackedI8K4`, not `PackedFormat`.
* **Not packing the activation at all.** ONNX Runtime's NEON sgemm reads A straight out of its
  register tiling and never packs it. That is a kernel design property, not something to graft on
  here.
* **Threading the pack.** Separate change, and the two gains compose.

🍍

### DIFF
--- a/linalg/src/arm32.rs
+++ b/linalg/src/arm32.rs
@@ -30,11 +30,12 @@ fn cpu_part() -> Option<usize> {
     })
 }
 
-fn has_neon() -> bool {
+pub(crate) fn has_neon() -> bool {
     if let Some(forced) = crate::knobs::TRACT_CPU_ARM32_NEON.get() {
         return forced;
     }
-    has_neon_cpuinfo().unwrap_or(false)
+    static NEON: std::sync::OnceLock<bool> = std::sync::OnceLock::new();
+    *NEON.get_or_init(|| has_neon_cpuinfo().unwrap_or(false))
 }
 
 pub fn plug(ops: &mut Ops) {
--- a/linalg/src/frame/pack.rs
+++ b/linalg/src/frame/pack.rs
@@ -721,7 +721,14 @@ unsafe fn pack_k_major<T: Copy>(
             let panel_width = r.min(mn_len - panel_mn);
             let src = b.offset(panel_mn as isize * mn_stride);
             let dst = packed.add(panel * panel_len);
-            let tiled_mn = panel_width / 4 * 4;
+            // 32-bit arm has no vector transpose to lower a tile to, and 16 live
+            // 32-bit values overflow its register file, so a tile there is a round
+            // trip through the stack. One store per element is cheaper.
+            let tiled_mn = if cfg!(target_arch = "arm") && std::mem::size_of::<T>() == 4 {
+                0
+            } else {
+                panel_width / 4 * 4
+            };
             let tiled_k = k_len / 4 * 4;
             for k in (0..tiled_k).step_by(4) {
                 for x in (0..tiled_mn).step_by(4) {
--- a/linalg/src/frame/pack.rs
+++ b/linalg/src/frame/pack.rs
@@ -326,14 +326,19 @@ impl PackedFormat {
                 }
             }
         } else if k_stride == 1 {
-            let mut packer = self.write_with_k_inner(pb, k_range.len(), mn);
-            let mn_valid_end = mn_range.end.min(mn);
-            for x in mn_range.start..mn_valid_end {
-                for k in k_range.clone() {
-                    packer.write(*b.offset(x as isize * mn_stride + k as isize))
-                }
-            }
             // just ignore invalid mn_range
+            let mn_valid_end = mn_range.end.min(mn);
+            if mn_valid_end > mn_range.start {
+                pack_k_major(
+                    b.offset(mn_range.start as isize * mn_stride + k_range.start as isize),
+                    pb,
+                    self.single_panel_len(k_range.len()),
+                    self.r,
+                    mn_stride,
+                    k_range.len(),
+                    mn_valid_end - mn_range.start,
+                )
+            }
         } else {
             let mut packer = self.write_with_k_outer(pb, k_range.len(), mn);
             let mn_valid_end = mn_range.end.min(mn);
@@ -689,6 +694,141 @@ unsafe fn pack_mn_major<Chunk: Copy>(
                 p_row.copy_from_nonoverlapping(b_row, partial_pane);
             }
         }
+    }
+}
+
+/// Pack a k-contiguous source block: transpose it into the k-inner packed
+/// layout, where source element `(mn, k)` of the block lands at
+/// `(mn / r) * panel_len + k * r + mn % r`. `b` points at element `(0, 0)`, and
+/// `mn_len` counts valid mn columns only: nothing outside the block is read.
+///
+/// The result must stay byte-identical to feeding [`KInWriter`] mn-outer /
+/// k-inner. Stores are strided by `r`, so the block moves as 4x4 tiles, k-outer
+/// so that each panel is filled front to back; the tails go element by element.
+#[inline(never)]
+unsafe fn pack_k_major<T: Copy>(
+    b: *const T,
+    packed: *mut T,
+    panel_len: usize,
+    r: usize,
+    mn_stride: isize,
+    k_len: usize,
+    mn_len: usize,
+) {
+    unsafe {
+        for panel in 0..mn_len.divceil(r) {
+            let panel_mn = panel * r;
+            let panel_width = r.min(mn_len - panel_mn);
+            let src = b.offset(panel_mn as isize * mn_stride);
+            let dst = packed.add(panel * panel_len);
+            let tiled_mn = panel_width / 4 * 4;
+            let tiled_k = k_len / 4 * 4;
+            for k in (0..tiled_k).step_by(4) {
+                for x in (0..tiled_mn).step_by(4) {
+                    transpose_4x4(
+                        src.offset(x as isize * mn_stride + k as isize),
+                        mn_stride,
+                        dst.add(k * r + x),
+                        r,
+                    );
+                }
+            }
+            for k in tiled_k..k_len {
+                for x in 0..tiled_mn {
+                    *dst.add(k * r + x) = *src.offset(x as isize * mn_stride + k as isize);
+                }
+            }
+            for x in tiled_mn..panel_width {
+                let row = src.offset(x as isize * mn_stride);
+                for k in 0..k_len {
+                    *dst.add(k * r + x) = *row.add(k);
+                }
+            }
+        }
+    }
+}
+
+/// Transpose a 4x4 tile: `src` rows are `src_stride` apart with contiguous
+/// elements, `dst` rows are `dst_stride` apart with contiguous elements. Both
+/// strides count elements and may leave the tiles unaligned. Specialised by
+/// element width where a vector transpose exists, portable everywhere else.
+#[inline(always)]
+unsafe fn transpose_4x4<T: Copy>(src: *const T, src_stride: isize, dst: *mut T, dst_stride: usize) {
+    unsafe {
+        // Alignment is part of the test: a 4-byte T of alignment 2 (Complex<i16>)
+        // must not be moved through a lane type it cannot be aligned for.
+        #[cfg(target_arch = "aarch64")]
+        if std::mem::size_of::<T>() == 4 && std::mem::align_of::<T>() == 4 {
+            transpose_4x4_neon_32(src as _, src_stride, dst as _, dst_stride);
+            return;
+        }
+        #[cfg(target_arch = "aarch64")]
+        if std::mem::size_of::<T>() == 2 && std::mem::align_of::<T>() == 2 {
+            transpose_4x4_neon_16(src as _, src_stride, dst as _, dst_stride);
+            return;
+        }
+        let tile: [[T; 4]; 4] = std::array::from_fn(|i| {
+            let row = src.offset(i as isize * src_stride);
+            std::array::from_fn(|j| *row.add(j))
+        });
+        for j in 0..4 {
+            let out = dst.add(j * dst_stride);
+            for (i, row) in tile.iter().enumerate() {
+                *out.add(i) = row[j];
+            }
+        }
+    }
+}
+
+/// 4x4 transpose of 32-bit lanes: four `ld1`, eight `trn`, four `st1`.
+#[cfg(target_arch = "aarch64")]
+#[target_feature(enable = "neon")]
+unsafe fn transpose_4x4_neon_32(
+    src: *const u32,
+    src_stride: isize,
+    dst: *mut u32,
+    dst_stride: usize,
+) {
+    use std::arch::aarch64::*;
+    unsafe {
+        let a = vld1q_u32(src);
+        let b = vld1q_u32(src.offset(src_stride));
+        let c = vld1q_u32(src.offset(2 * src_stride));
+        let d = vld1q_u32(src.offset(3 * src_stride));
+        let ab_even = vreinterpretq_u64_u32(vtrn1q_u32(a, b));
+        let ab_odd = vreinterpretq_u64_u32(vtrn2q_u32(a, b));
+        let cd_even = vreinterpretq_u64_u32(vtrn1q_u32(c, d));
+        let cd_odd = vreinterpretq_u64_u32(vtrn2q_u32(c, d));
+        vst1q_u32(dst, vreinterpretq_u32_u64(vtrn1q_u64(ab_even, cd_even)));
+        vst1q_u32(dst.add(dst_stride), vreinterpretq_u32_u64(vtrn1q_u64(ab_odd, cd_odd)));
+        vst1q_u32(dst.add(2 * dst_stride), vreinterpretq_u32_u64(vtrn2q_u64(ab_even, cd_even)));
+        vst1q_u32(dst.add(3 * dst_stride), vreinterpretq_u32_u64(vtrn2q_u64(ab_odd, cd_odd)));
+    }
+}
+
+/// 4x4 transpose of 16-bit lanes, on 64-bit halves of the vector registers.
+#[cfg(target_arch = "aarch64")]
+#[target_feature(enable = "neon")]
+unsafe fn transpose_4x4_neon_16(
+    src: *const u16,
+    src_stride: isize,
+    dst: *mut u16,
+    dst_stride: usize,
+) {
+    use std::arch::aarch64::*;
+    unsafe {
+        let a = vld1_u16(src);
+        let b = vld1_u16(src.offset(src_stride));
+        let c = vld1_u16(src.offset(2 * src_stride));
+        let d = vld1_u16(src.offset(3 * src_stride));
+        let ab_even = vreinterpret_u32_u16(vtrn1_u16(a, b));
+        let ab_odd = vreinterpret_u32_u16(vtrn2_u16(a, b));
+        let cd_even = vreinterpret_u32_u16(vtrn1_u16(c, d));
+        let cd_odd = vreinterpret_u32_u16(vtrn2_u16(c, d));
+        vst1_u16(dst, vreinterpret_u16_u32(vtrn1_u32(ab_even, cd_even)));
+        vst1_u16(dst.add(dst_stride), vreinterpret_u16_u32(vtrn1_u32(ab_odd, cd_odd)));
+        vst1_u16(dst.add(2 * dst_stride), vreinterpret_u16_u32(vtrn2_u32(ab_even, cd_even)));
+        vst1_u16(dst.add(3 * dst_stride), vreinterpret_u16_u32(vtrn2_u32(ab_odd, cd_odd)));
     }
 }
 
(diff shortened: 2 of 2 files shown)


czoli1976:
I am looking at this, running it locally ...

kali:
I think we have positive signal here. Should we try and generalize the approach, make t4x4_f32 and t4x4_f16 new primitives in linalg ?

czoli1976:
Really nice piece of work — thank you for it, and especially for the write-up. The k-outer ordering rationale and the mutation testing made this unusually pleasant to review; being able to check your reasoning against `KInWriter` as the reference is exactly what one wants and rarely gets.

I was curious enough to replicate it independently against `main` @ `88663ad52`, on two machines:

| | machine | chip | SME |
| --- | --- | --- | --- |
| **A** | MacBook Pro 16" (`MacBookPro18,3`) | Apple **M1 Pro**, 6P+2E, 16 GB | no (`FEAT_SME=0`) |
| **B** | MacBook Air 13" (`Mac16,12`) | Apple **M4**, 10 cores, macOS 26.5.1 | yes (`FEAT_SME=1`, `SME2`, `SME_I8I32`) |

Machine A carried the end-to-end work — it's the same generation as the bot's `apple-m1-max` reference device, so those deltas should be close to like-for-like. Machine B is your own chip generation, so I used it for the block benchmarks and to settle the SME footnote. Everything checks out: correctness, the speed claims, and bit-identical packing. One thing I think is worth raising, plus a few notes that are just information rather than requests.

Your isolated block numbers reproduce on a different M4 within ~2%:

| block | yours (M4 Pro) | here (MacBook Air M4) |
| --- | --: | --: |
| 256×768 f32 r=8 | 3.40× | 3.33× |
| 256×768 f32 r=24 | 3.44× | 3.42× |
| 256×768 f16 r=32 | 4.48× | 4.38× |

Also, your SME footnote checks out — on the MacBook Air M4, `main` gives 4403 passed / 4 failed and this PR gives 4410 passed / 4 failed, the identical `test_sme_qmmm_i32_32x32::store_i8::add_unicast_*` four. Pre-existing and not yours. (It does reproduce on more than one M4, so it may be worth its own issue at some point.)

**Reproduced:** `tract-linalg` 4412/0 on the M1 Pro (`--features complex`; that box has no SME, hence no sign of the four above), `tract-core` 269, `test-unit-core` 816, `fmt` clean, portable path green with the aarch64 arms `#[cfg]`'d out, and all five of your mutants caught (each failed 6–8 of the `k_major` tests). Model output bit-identical on `hey_snips_v1` and `hey_snips_v4_model17` with seeded inputs. End-to-end, `-O`, min of 5 interleaved rounds:

| model | main | PR | here | bot (m1-max) |
| --- | --: | --: | --: | --: |
| hey_snips_v1 | 0.147 ms | 0.138 ms | −6.1% | −6.9% |
| hey_snips_v4_model17 | 1.380 ms | 1.222 ms | −11.4% | −11.7% |

### The one thing I'd raise: the cortex-a9 red looks real, and specific to armv7 + f32

`bench.yml` maps cortex-a9 and cortex-a7 to `armv7-unknown-linux-gnueabihf`, so those two boards take the portable tile rather than the intrinsics. Emitting assembly for that triple, per single 4×4 tile:

| target | tile compiles to | stack spills |
| --- | --- | --: |
| aarch64 f32/f16 | NEON `trn` | 0 |
| aarch64 i8 (portable) | scalar, register-resident, 25 insns | 0 |
| x86_64 SSE2/AVX2 | scalar, no shuffles, 84 insns (main: 129/155) | 0 |
| **armv7, 4-byte T** | scalar, no NEON even with `+neon` | **8**, plus 9 callee-saved pushes |
| armv7, 2-byte T | scalar, no NEON | 0 |

armv7 has ~14 usable GPRs, so 16 live 32-bit values don't fit and LLVM spills — against code that previously did one store per element with no frame at all. That fits the bot exactly: the one red is `hey_snips_v1` on cortex-a9, `hey_snips_v1` is f32, and the same model is −6.9% on m1-max.

Setting `tiled_mn = 0` there is enough on its own — the tile and k-tail loops go empty and the existing mn-tail loop already covers the panel at the same addresses:

```rust
// A 4x4 tile needs 16 live values. armv7 has ~14 usable GPRs, so a 32-bit tile spills.
let tiled_mn = if cfg!(target_arch = "arm") && std::mem::size_of::<T>() == 4 {
    0
} else {
    panel_width / 4 * 4
};
```

Caveat: I have no armv7 board or emulator, so this is emitted assembly plus the bot, not a measurement — only the dinghy boards can confirm the +6.1% actually goes away. And it's entirely reasonable to just accept it, given what the change buys everywhere else.

Incidentally this also answers your open question about SSE: it doesn't become a vector transpose on SSE2 or AVX2 either, but it spills nothing there and comes out shorter than the code it replaces, so x86 should be fine.

### If `linalg/benches/pack.rs` lands, worth labelling the cache assumption

The single-buffer block bench is fully cache-resident, so it's an upper bound. Rotating through a ~192 MB working set so each pack starts cold:

| block | M4 Air hot | M4 Air cold | M1 Pro hot | M1 Pro cold |
| --- | --: | --: | --: | --: |
| 256×768 f32 r=8 | 3.33× | **1.30×** | 6.4× | 2.30× |
| 256×768 f32 r=24 | 3.42× | 1.57× | 5.8× | 3.2× |
| 256×768 f16 r=32 | 4.38× | 2.32× | 7.5× | 4.6× |

Neither column is "the" answer — a real model is partly resident and lands between them. Reassuringly your own `output/dense/MatMul.pack_b` figure (10.89/4.72 = 2.31×) sits inside that bracket on both machines. It just means the hot number shouldn't be read as an end-to-end expectation when the model-level effect is ~2× at the node and 6–11% overall.

### One small regression: `mn < 4` with small `k`

Below a full tile there's no tiling to do, so the panel goes through the scalar tail and the `#[inline(never)]` call plus per-panel setup costs more than the `KInWriter` state machine did. Consistent across six rounds:

| shape | MacBook Pro M1 Pro | MacBook Air M4 |
| --- | --- | --- |
| `1×64` f32 r=8 | 34.9 → 35.6 ns (0.97×) | 19.0 → 22.0 ns (**0.86×**) |
| `1×256` f32 r=8 | 153.3 → 106.9 ns (1.43×) | 72.7 → 74.5 ns (0.98×) |
| `4×64` f32 r=8 | 162.1 → 33.2 ns (4.88×) | 80.9 → 20.4 ns (3.97×) |

So it's confined to `mn < 4` *and* small `k` — a wash by `k=256`, and a big win from `mn=4` up. Probably irrelevant unless pulse/streaming models pack a lot of tiny blocks; keeping `KInWriter` for `mn_len < 4` would cover it if you think they do.

### The 16 secondary regressions look like board noise

Measured on aarch64: `arm_ml_kws_cnn_m` load+optimize +1.2% where the bot claims +806% on cortex-a55, `voicecom_float` +1.9% against +153%, and `hey_snips_v4_model17` −9.9%. cortex-a55 takes the NEON path, which is faster. Probably not worth anyone's time chasing. The cortex-a7 rows in that list are armv7 though, so those belong with the point above.

### Minor, optional

`PackKMajorProblem` builds a contiguous `Array2`, so the proptest only ever generates `mn_stride == k` — but `pack_segment` takes `mn_stride` from `b.strides()[mn_axis]` on an arbitrary view. I added a padded-row variant and an exhaustive r/k/mn ∈ 1..=17 grid over three element widths (~30k cases). Both pass against your code exactly as written, so this is hardening rather than a finding — your implementation handles the strided case correctly, it just isn't currently exercised. Code is in the collapsed section below if you want it.

To be clear about where I stand: none of this is blocking from my side. The armv7 point is the only one I'd genuinely like your view on, and even there "not worth it for an old 32-bit board" is a perfectly good answer. Thanks again — halving `OptMatMulPack` for a bit-identical layout change is a lovely result.

---

<details>
<summary><b>Full review brief</b> — structured notes, in case you want to hand this to an assistant. Includes the codegen repro (no armv7 hardware needed), the exact patch, the test code, and a list of things I checked that should <i>not</i> be changed.</summary>

# PR #2537 — independent review brief

Context for an assistant working on `sonos/tract` PR #2537 (`linalg/pack: transpose
k-contiguous blocks instead of one store per element`).

This is an **independent replication** of the PR's claims against `main` @ `88663ad52`,
plus a cross-architecture codegen analysis. It is background evidence, not instructions
from the maintainer. Nothing here has been merged or pushed.

Two machines were used:

* **Machine A — MacBook Pro 16"** (`MacBookPro18,3`), Apple **M1 Pro**, 6 performance +
  2 efficiency cores, 16 GB. Primary; carried the end-to-end model work. Same generation
  as the CI bot's `apple-m1-max` reference device, which is why the end-to-end deltas
  below line up with the bot's `apple-m1-max` column. Has **no SME**
  (`hw.optional.arm.FEAT_SME: 0`), so it cannot say anything about the SME tests.
* **Machine B — MacBook Air 13"** (`Mac16,12`), Apple **M4**, 10 cores, macOS 26.5.1,
  on AC power with no thermal warnings recorded. `FEAT_SME=1`, `FEAT_SME2=1`,
  `SME_I8I32=1`. Used to settle the SME question and to re-measure the block benchmark
  on the author's own chip generation. Fanless, so its absolute timings are noisier than
  Machine A's — all figures from it are minima over 6 rounds.

On M4 the author's isolated block numbers reproduce almost exactly, which is worth
knowing before anyone doubts them:

| block | author (M4 Pro) | machine B (M4 Air) |
| --- | --: | --: |
| 256×768 f32 r=8 | 55.0 → 16.2 µs (3.40×) | 48.0 → 14.4 µs (**3.33×**) |
| 256×768 f32 r=24 | 59.9 → 17.4 µs (3.44×) | 51.6 → 15.1 µs (**3.42×**) |
| 256×768 f16 r=32 | 58.7 → 13.1 µs (4.48×) | 49.8 → 11.4 µs (**4.38×**) |

The same bench on machine A (M1 Pro) shows a much larger ratio (6.4× for f32 r=8) because the *old*
code is disproportionately slow there — 132 µs vs 48 µs on M4 — while the new code is
close on both. Quote the M4 ratios, not the M1 Pro ones.

There is exactly **one requested change** (§2). Everything else is either confirmation
that the PR is right, or an explicit "do not change this" (§5).

---

## 1. What was verified (all of the PR's claims hold)

| PR claim | replication result |
| --- | --- |
| `tract-linalg` passes (~4408) | 4412 passed, 0 failed on machine A (M1 Pro), `--features complex` (that box has no SME, so the four failures in §5.6 cannot appear there) |
| `tract-core` 269 | 268 + 1 = 269 |
| `test-unit-core` 816 | 816 |
| portable path green with aarch64 arms `#[cfg]`'d out | green |
| 5 mutants caught | 5/5 reproduced; each failed 6–8 `k_major` tests |
| `cargo fmt --all` clean | clean |
| clippy clean | no PR-attributable warnings |
| model output bit-identical | confirmed, 2 models, seeded input, `array_equal == True` |
| 4 SME failures pre-existing & unrelated | confirmed on machine B (MacBook Air M4) — see §5.6 |
| isolated block speedups (M4 Pro) | reproduced within 2% on machine B — see table above |

The address math was also reviewed by hand: the tile / k-tail / mn-tail loops exactly
partition each panel with no gap or overlap; the change from passing the full-tensor
`mn` to `mn_len` is address-equivalent because `KInWriter` only used `mn` to decide
panel transitions; `end_padding_record` and `align_panel` flow through the same
`single_panel_len` on both paths.

End-to-end (`-O`, min of 5 interleaved rounds):

| model | main | PR | delta | CI bot (apple-m1-max) |
| --- | --: | --: | --: | --: |
| hey_snips_v1 (f32) | 0.147 ms | 0.138 ms | −6.1% | −6.9% |
| hey_snips_v4_model17 | 1.380 ms | 1.222 ms | −11.4% | −11.7% |
| inceptionv1q | 52.09 ms | 50.29 ms | −3.5% (noisy) | −8.2% |

---

## 2. REQUESTED CHANGE — armv7 falls back to scalar for 4-byte types

### Why

`.github/workflows/bench.yml` lines 299–307 map the `cortex-a9` and `cortex-a7` bench
devices to `armv7-unknown-linux-gnueabihf`. That is `target_arch = "arm"`, so those
boards take the **portable** tile in `transpose_4x4`, not the NEON intrinsics.

Compiled for that triple, the portable tile behaves very differently by element width.
Per single 4×4 tile:

| target | tile compiles to | stack spills |
| --- | --- | --: |
| aarch64 f32 / f16 | explicit NEON `trn` | 0 |
| aarch64 i8 (portable tile) | scalar, register-resident, 25 insns | 0 |
| x86_64 SSE2 / AVX2 | scalar, no shuffles, 84 insns (old code: 129 / 155) | 0 |
| **armv7, 32-bit T** | scalar, **no NEON even with `-C target-feature=+neon`** | **8**, plus 9 callee-saved registers pushed |
| armv7, 16-bit T | scalar, no NEON | 0 |

armv7 has ~14 usable general-purpose registers, so 16 live 32-bit values cannot stay in
registers and LLVM spills. The code being replaced was one store per element with no
spill and no frame, so on armv7 the tile is a net loss for 4-byte types.

This matches the bot exactly: the only red speed regression is `hey_snips_v1` on
`cortex-a9` at +6.1%, and `hey_snips_v1` is an f32 model. The *same model* is −6.9% on
apple-m1-max. Same code path, opposite sign, explained by register file size.

### Reproduce the codegen claim (no armv7 hardware needed)

```
rustup target add armv7-unknown-linux-gnueabihf
```

Put the portable tile in a standalone file and inspect the emitted assembly:

```rust
#![crate_type = "lib"]
#[inline(always)]
pub unsafe fn transpose_4x4<T: Copy>(src: *const T, src_stride: isize, dst: *mut T, dst_stride: usize) {
    unsafe {
        let tile: [[T; 4]; 4] = std::array::from_fn(|i| {
            let row = src.offset(i as isize * src_stride);
            std::array::from_fn(|j| *row.add(j))
        });
        for j in 0..4 {
            let out = dst.add(j * dst_stride);
            for (i, row) in tile.iter().enumerate() { *out.add(i) = row[j]; }
        }
    }
}
#[no_mangle] pub unsafe extern "C" fn tile_f32(s: *const f32, ss: isize, d: *mut f32, ds: usize) { unsafe { transpose_4x4::<f32>(s, ss, d, ds) } }
#[no_mangle] pub unsafe extern "C" fn tile_f16(s: *const u16, ss: isize, d: *mut u16, ds: usize) { unsafe { transpose_4x4::<u16>(s, ss, d, ds) } }
```

```
rustc --target armv7-unknown-linux-gnueabihf -O -C target-feature=+neon --emit asm tile.rs -o tile.s
```

In `tile_f32` you will see `sub sp, sp, #16`, `push {r4..r11, lr}`, eight `[sp]`
accesses, and zero `vtrn`/`vld1`/`vst1`. In `tile_f16` there are no `[sp]` accesses.
Swap the target to `x86_64-unknown-linux-gnu` or `aarch64-apple-darwin` and the `[sp]`
count is zero in both.

### The change

In `pack_k_major` (`linalg/src/frame/pack.rs`), gate the tiled path off for 4-byte
types on armv7. Setting `tiled_mn = 0` is sufficient and needs no other edit: the tile
loop and the k-tail loop both become empty, and the existing mn-tail loop already walks
the whole panel element by element at the same destination addresses.

```rust
// A 4x4 tile needs 16 live values. armv7 has ~14 usable GPRs, so a 32-bit tile spills.
let tiled_mn = if cfg!(target_arch = "arm") && std::mem::size_of::<T>() == 4 {
    0
} else {
    panel_width / 4 * 4
};
```

Replaces the current:

```rust
let tiled_mn = panel_width / 4 * 4;
```

### Verify after changing

* `cargo test -p tract-linalg --release pack` — still green (the fallback path is
  already covered; `k_major_narrower_than_a_tile` exercises `tiled_mn == 0`).
* Re-emit the armv7 asm for `pack_k_major` and confirm the tile is gone.
* aarch64 and x86_64 codegen must be **unchanged** — the `cfg!` is a compile-time
  constant and folds away on every non-arm target.

### Important caveat

This was **not measured on armv7 hardware**. The reviewer had no armv7 board, no QEMU
and no Docker. The evidence is emitted assembly plus the CI bot. Only the dinghy boards
can confirm the +6.1% actually goes away. It is also a legitimate maintainer decision to
accept a 6% loss on an old 32-bit board in exchange for 6–11% on aarch64 — do not treat
this change as mandatory, and do not claim it is verified.

---

## 3. Benchmark caveat worth writing down

The PR offers to add `linalg/benches/pack.rs`. If that lands, the numbers should be
labelled with their cache assumption, because a single-buffer block bench is fully
cache-resident and is an upper bound, not an expected value.

Same pack call, same machine; the only difference is whether the source is already in
cache. "Cold" rotates through a ~192 MB working set of source/destination pairs so each
pack starts with nothing in L2.

| block | A (M1 Pro) hot | A cold | B (M4 Air) hot | B cold |
| --- | --: | --: | --: | --: |
| 256×768 f32 r=8 | 6.4× | 2.30× | 3.33× | **1.30×** |
| 256×768 f32 r=24 | 5.8× | 3.2× | 3.42× | 1.57× |
| 256×768 f16 r=32 | 7.5× | 4.6× | 4.38× | 2.32× |

Read these as a **bracket**, not as "cold is the truth": a real model is partially
cache-resident, so its gain sits somewhere between the two columns. The PR's own
measured node speedup for `layer.N/output/dense/MatMul.pack_b` — 10.89 / 4.72 = **2.31×**
— falls inside that bracket on both machines, which is a good sign the author's model
measurements are sound.

The point for the bench file is narrower: do not let a hot 3.3–6.4× be read as an
end-to-end expectation when the model-level effect is ~2× at the node and 6–11% overall.

---

## 3b. Minor measured regression: `mn < 4` with small `k`

Not a correctness issue and arguably not worth acting on, but it is real and it
reproduces on two machines, so it should not come as a surprise later.

When `panel_width < 4` no tiling is possible, `tiled_mn` is 0, and the whole panel goes
through the scalar mn-tail loop. That path is fine for larger `k`, but at very small
total work the `#[inline(never)]` call plus the per-panel setup costs more than the
`KInWriter` state machine it replaced.

| shape | A (M1 Pro) main → PR | B (M4 Air) main → PR |
| --- | --- | --- |
| `1×64` f32 r=8 | 34.9 → 35.6 ns (0.97×) | 19.0 → 22.0 ns (**0.86×**) |
| `1×256` f32 r=8 | 153.3 → 106.9 ns (1.43×) | 72.7 → 74.5 ns (0.98×) |
| `4×64` f32 r=8 | 162.1 → 33.2 ns (4.88×) | 80.9 → 20.4 ns (3.97×) |

So it is confined to `mn < 4` **and** small `k`; by `k = 256` it is a wash, and from
`mn = 4` up the change is a large win everywhere. The M4 figure is a consistent −14%
across six rounds, not noise.

Only worth a fix if streaming/pulse models pack many tiny blocks in practice. If so, the
cheapest option is to keep `KInWriter` when `mn_len < 4`. Do not restructure
`pack_k_major` for this.

---

## 4. Optional: two test-coverage additions

Neither found a bug. Both pass against the PR as written. Purely hardening — take or
leave.

`PackKMajorProblem` builds `Array2::from_shape_fn((mn, k))`, which is contiguous, so the
proptest only ever generates `mn_stride == k`. But `pack_segment` takes `mn_stride` from
`b.strides()[mn_axis]` on an arbitrary `TensorView`, which can be larger. Nothing
currently exercises that.

Add to `mod test` in `linalg/src/frame/pack.rs`:

```rust
    // pack_segment takes mn_stride from an arbitrary TensorView, so rows need not be
    // laid end to end. The proptest builds a contiguous [mn, k] array, which only ever
    // produces mn_stride == k.
    fn check_k_major_strided<T: Datum + Copy + num_traits::Zero>(
        k: usize,
        mn: usize,
        r: usize,
        pad: usize,
        value: impl Fn(usize) -> T,
    ) {
        let packer = super::PackedFormat::new(T::datum_type(), r, 1);
        let stride = k + pad;
        let src: Vec<T> = (0..mn * stride).map(&value).collect();
        let len = packer.len(k, mn);
        let mut packed = Tensor::zero::<T>(&[len]).unwrap();
        let mut reference = Tensor::zero::<T>(&[len]).unwrap();
        unsafe {
            packer.pack_t::<T>(
                packed.as_ptr_mut_unchecked::<T>(),
                src.as_ptr(),
                mn,
                1,
                stride as isize,
                0..k,
                0..mn,
            );
            let mut w = packer.write_with_k_inner(reference.as_ptr_mut_unchecked::<T>(), k, mn);
            for x in 0..mn {
                for kk in 0..k {
                    super::PackingWriter::write(&mut w, src[x * stride + kk]);
                }
            }
        }
        assert_eq!(packed, reference, "k={k} mn={mn} r={r} pad={pad} dt={:?}", T::datum_type());
    }

    #[test]
    fn k_major_padded_rows() {
        for (k, mn, r, pad) in
            [(16usize, 16usize, 8usize, 5usize), (9, 7, 8, 1), (768, 32, 8, 3), (4, 4, 4, 7)]
        {
            check_k_major_strided::<u32>(k, mn, r, pad, |i| i as u32 * 7 + 1);
            check_k_major_strided::<f16>(k, mn, r, pad, |i| f16::from_f32(i as f32));
            check_k_major_strided::<u8>(k, mn, r, pad, |i| i as u8);
        }
    }

    #[test]
    fn k_major_exhaustive_small() {
        for r in 1..=17usize {
            for k in 1..=17usize {
                for mn in 1..=17usize {
                    for pad in [0usize, 3] {
                        check_k_major_strided::<u32>(k, mn, r, pad, |i| i as u32 * 7 + 1);
                        check_k_major_strided::<f16>(k, mn, r, pad, |i| f16::from_f32(i as f32));
                        check_k_major_strided::<u8>(k, mn, r, pad, |i| i as u8);
                    }
                }
            }
        }
    }
```

`k_major_exhaustive_small` is ~30k cases across three element widths and runs in well
under a second in release.

---

## 5. DO NOT CHANGE — things that were checked and are correct

An assistant reading this should not "fix" any of the following. Each was investigated
and found to be correct as written.

1. **The `align_of::<T>() == 4` / `== 2` guards in `transpose_4x4`.** Correct and worth
   keeping. Deleting them does not fail any test, and cannot: the transpose is
   byte-preserving, so results are identical either way, and tract allocates every
   tensor at `vector_size()` alignment (`data/src/tensor.rs:264`), so a 4-byte element is
   always 4-aligned in practice regardless of the type's declared alignment. The guard is
   defensive against a future non-`Tensor` caller. Do not remove it, and do not add a
   test claiming to cover it.
2. **The 16 "secondary" load / load+optimize regressions in the bot comment.** Not caused
   by this diff. Measured on aarch64: `arm_ml_kws_cnn_m` load+optimize +1.2% (bot claimed
   +806% on cortex-a55), `voicecom_float` +1.9% (bot: +153%), `hey_snips_v4_model17`
   −9.9%. cortex-a55 takes the NEON path, which is measurably faster. Treat as board
   noise. The `cortex-a7` rows in that list are armv7 and belong to §2 instead.
3. **The `// ---- k-contiguous packing ----` banner comment in the tests.** CLAUDE.md
   discourages section banners, but the file already uses one for `PackedI8K4`, so this
   matches local style.
4. **`#[target_feature(enable = "neon")]` without runtime feature detection.** Matches
   existing repo convention, e.g. `linalg/src/arm64/arm64simd/rms_norm.rs:16`.
5. **The small-block / GEMV regime** — see §3b, which reports the one measured
   regression there. Everything from 4×64 upward is 2–6× faster on both machines; only
   `mn < 4` with small `k` regresses, and that is a judgement call, not a defect.
6. **The 4 SME test failures mentioned in the PR description.** The author's assessment
   is **confirmed** on independent hardware — they are pre-existing and unrelated to this
   PR. Verified on machine B, an unrelated MacBook Air M4 (Mac16,12, `FEAT_SME=1`, `FEAT_SME2=1`,
   `SME_I8I32=1`), running the compiled `tract-linalg` test binary from both worktrees:

   | build | result |
   | --- | --- |
   | `main` @ 88663ad52 | 4403 passed, **4 failed** |
   | PR #2537 | 4410 passed, **4 failed** |

   Identical failures in both: `arm64::sme::test_sme_qmmm_i32_32x32::store_i8::add_unicast_`
   `{arbitrary,col_major,dt,row_major}`. The PR adds 7 tests and zero new failures; its
   own pack tests are 1255/0 on real SME silicon. Do not attempt to "fix" these as part
   of this PR — but note they are a genuine pre-existing bug in tract main on M4-class
   hardware, reproducible on more than one machine, which may deserve its own issue.

---

## 6. Reproduction notes

* Baseline used: `main` @ `88663ad52`. PR head: `5528aea5e`.
* Models come from the manifest in `.travis/benches.toml`, base URL
  `https://tract-ci-builds.s3.amazonaws.com/model`.
* `-O` is a **global** flag and must precede the model path:
  `tract -O <model> -i 80,40,f32 bench -R`. Passing `-O` after `bench` is rejected.
* Bit-identity check: run both binaries with
  `run --input-from-npz <fixed.npz> --save-outputs-npz <out.npz>` and compare with
  `numpy.array_equal`. Input node names are `inputs` for `hey_snips_v1.pb` and
  `input_node` for `hey_snips_v4_model17.pb`.
* All A/B timings here are interleaved (main, PR, main, PR, …) and reported as the
  minimum across rounds, to keep thermal drift out of the comparison.

</details>

czoli1976:
@kali — from the cross-arch codegen in my comment above, I'd say yes, and the primitive boundary lands exactly where the per-arch behaviour diverges today:

| target | what the portable tile compiles to |
| --- | --- |
| aarch64 | scalar but register-resident — fine as-is |
| x86_64 SSE2 / AVX2 | scalar, **zero shuffles** — LLVM never vectorises it |
| armv7, 32-bit | scalar, **8 stack spills** per tile |

aarch64 is already covered by the intrinsics in this PR, so the case for real `t4x4_f32` / `t4x4_f16` primitives is the other two: x86 has unclaimed headroom (an explicit `unpcklps`/`shufps` tile should beat the scalar code LLVM emits today), and armv7 gets a principled place to opt out instead of a `cfg!` inside `pack_k_major`.

Two caveats worth pricing in. At model scale the pack is memory-bound — cold-cache speedups are 1.30–2.32× against 3.3–4.4× hot — so the end-to-end return on a faster transpose is a good deal smaller than the block bench suggests; the win here is arch coverage more than raw ratio. And a quick look through core (einsum, change_axes) didn't turn up an obvious second consumer, so the justification would be per-arch dispatch rather than reuse.

kali:
neon-less armv7 is of very little interest to tract, but that's the target rust will pick (and I don't want to change that). so all neon usage on arm7 have to be dispatched at runtime (so it belongs in linalg). It is possible we have a real performance boost in there, I will give it a shot, I have devices.

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali:
I think I'm ready to merge this one now. Any last thoughts ?

PR #2538, opened 2026-07-31
bench: compile bench-suite with multithread-mm

kali:
Pulls the rayon multithread-mm feature into the bench-suite build so CI benches exercise the multithread codegen. Runtime stays single-threaded (the suite never passes --threads), so this measures the compile-in cost of the feature; a local i9 A/B pass showed it in the noise floor.


kali:
relates to #2534's convo.

PR #2539, opened 2026-07-31
Add tools/asset-mirror, a LAN caching reverse-proxy for model assets

kali:
Serves the R2 origin over HTTP on the LAN and caches immutable model blobs on first miss, so fleet/dev boxes stop each keeping a multi-GB copy. Kept out of the workspace (root exclude) so its TLS/HTTP deps stay off the shared lock and CI.


kali:
Unknown `/ci` target: `help`.

Usage: `/ci full`, `/ci large-models` (aliases: `large`, `llm`), `/ci wheels` (aliases: `python`, `py`), `/ci examples` (alias: `ex`), or `/ci bench` (alias: `benches`).

PR #2544, opened 2026-08-01
metal: split multi-axis Reduce into single-axis reduces

czoli1976:
Ports the `split_multi_axis_reduce` rule from #2077 (cuda) to the Metal transform: a `Reduce` over several axes is rewritten into a chain of single-axis reduces before translation, so it stops falling back to CPU. `GpuReduce` only ever dispatches `axes[0]`, so a multi-axis reduce could not translate and the graph had to round-trip through the host in the middle.

`MeanOfSquares` is excluded, as in the cuda rule, since chaining it per axis is not the multi-axis result.

### Effect on MobileNet v2 (`--metal -O`, `dump --audit-json`)

The global average pool (`Reduce<Sum>` over two axes) was the only op left on host:

| | nodes | sync nodes | ops still on CPU |
|---|---|---|---|
| before | 376 | 4 | `Reduce<Sum>` |
| after | 375 | 2 | none |

The two remaining syncs are the model's own input and output.

Same seeded input through both binaries: max absolute difference `9.5e-07` over the 1000 logits, identical argmax.

Interleaved A/B, 5 rounds each, 300 loops per run, M-series:

```
before  13.119  12.629  11.616  12.842  13.183   best 11.616  mean 12.68
after   11.417  11.335  11.340  11.499  11.671   best 11.335  mean 11.45
```

Best-to-best that is 2.4%; the wider mean gap is mostly the run-to-run jitter the mid-graph flush was adding (before spans 1.57 ms, after 0.34 ms). The point of the change is removing the fallback rather than the timing.

### Checks

`cargo fmt --all` clean, `cargo clippy --workspace` clean, `tract-core` 268/268.

Two failures in `tract-gpu` / `tract-metal` reproduce identically on `main` with this commit stashed out, so they are not from this change: `tensor::tests::test_device_tensor` ("GPU Context not initialized") and `kernels::matmul::mfa::tests::test_mfa_attention_causal_const_is_noop` (the causal probe returns `full=0.08356` where it expects the triangular-const-alone value). Happy to open a separate issue for the latter if it is not already known.

🍍

### DIFF
--- a/metal/src/transform.rs
+++ b/metal/src/transform.rs
@@ -14,6 +14,7 @@ use tract_core::ops::cnn::conv::rewrite_kernel_conv_in_oihw;
 use tract_core::ops::cnn::{Conv, rewrite_conv_with_n_axis};
 use tract_core::ops::einsum::prefix_matmul::{PrefixMatMul, rewrite_einsum_to_prefix_matmul};
 use tract_core::ops::konst::Const;
+use tract_core::ops::nn::Reduce;
 use tract_core::tract_linalg::block_quant::Q4_0;
 use tract_core::transform::ModelTransform;
 use tract_gpu::fact::{DeviceFact, DeviceTypedFactExt};
@@ -150,6 +151,7 @@ impl MetalTransform {
             .with_rule_for("rewrite_kernel_conv_in_oihw", rewrite_kernel_conv_in_oihw)
             .with_rule_for("rewrite_conv_with_n_axis", rewrite_conv_with_n_axis)
             .with_rule_for("remove_rms_norm_cast", remove_rms_norm_cast)
+            .with_rule_for("split_multi_axis_reduce", split_multi_axis_reduce)
             .rewrite(&(), model)?;
 
         if stop_at_phase == 1 {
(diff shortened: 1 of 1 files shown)


kali:
> Two failures in `tract-gpu` / `tract-metal` reproduce identically on `main` with this commit stashed out, so they are not from this change: `tensor::tests::test_device_tensor` ("GPU Context not initialized") and `kernels::matmul::mfa::tests::test_mfa_attention_causal_const_is_noop` (the causal probe returns `full=0.08356` where it expects the triangular-const-alone value). Happy to open a separate issue for the latter if it is not already known.

I'm not aware of these. Do we have a gap in the CI coverage ?

metal/src/transform.rs
```diff
@@ -503,3 +505,28 @@ fn convert_const(op: &Const) -> TractResult<Const> {
     let metal_const = op.val().clone().into_device()?.into_tensor().into_arc_tensor();
     Const::new_with_exotic_fact(metal_const, Box::new(metal_fact))
 }
+
+/// Rewrites a `Reduce` over several axes into a chain of single-axis reduces, which is
+/// what `GpuReduce` accepts. Only reducers that compose associatively per axis qualify:
+/// `MeanOfSquares` over a chain is not the multi-axis result.
+fn split_multi_axis_reduce(
```
kali, line 512: is this a function that already exists in cuda ? (just guessing from the commit message). Can we unify them in tract-gpu ?

czoli1976: Yes — it's the rule from #2077; I mirrored it rather than sharing it because `add_matmul_broadcast` / `fuse_axis_op` / `untranspose_matmul_output` are all per-backend duplicates today. Unified now: it lives in `tract_gpu::rewrite_rules::split_multi_axis_reduce` and both transforms import it, so the cuda copy goes away too and the PR is net -24 lines. Same graph and same outputs on MobileNet as before the hoist.

czoli1976:
On the two failures — they turn out to be different stories.

The MFA one is not a CI gap: `crates.yml`'s `metal` job runs `cargo test -p tract-metal -p test-metal` on every PR, and `metal (1.91)` passed on this PR's commit. So it passes on your runner and fails only on mine. This machine is on a macOS beta, so my guess is a driver/OS difference in the causal path rather than a code bug — happy to dig into it separately and open an issue if it turns out to be real.

`tract-gpu` does look like a genuine gap though: nothing in `.github/workflows` references it, so `cargo test -p tract-gpu` never runs anywhere. Standalone it cannot pass — `tensor::tests::test_device_tensor` fails on "GPU Context not initialized" because no backend has registered a context in that test binary.

Unrelated, for the record: the bench bot's 4 secondary regressions are all `arm_ml_kws_cnn_m` load / load+optimize on cortex-a7 and cortex-a9, where `tract-metal` isn't even a dependency (`cli/Cargo.toml` gates it behind `cfg(any(target_os = "macos", target_os = "ios"))`), so they can't come from this change.

metal/src/transform.rs
czoli1976: moved to tract-gpu

kali:
About the load, I tried to load in two times to split away the i/o from the actual loading. We'll see. I acknowledge it's borderline useless today.

kali:
@czoli1976 i'm going to remove my two commits from there. it was opportunistic, and it failed spectacularly. Moving them to a different PR

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2546, opened 2026-08-01
metal: drop the MFA triangular-constant probe test- fails on M1 Pro (macOS 26.5.2) and an M4 (macOS 26.6).

czoli1976:
`test_mfa_attention_causal_const_is_noop` fails on macOS 26.5 — the probe it
locks down (MFA's `triangular` function constant alone does not mask) no longer
holds there, so `cargo test -p tract-metal` is red on that OS:

```
causal probe: lower(j<=i)=0.51085 upper(j>=i)=0.49934 full=0.08356
Error: expected triangular-const-alone == full attn, got full=0.08356253
```

Nothing in tract depends on it: `MetalMfaSdpa` passes the constant as `false`
and realizes causal as a bottom-right aligned `[Sq,Sk]` additive mask, so the
test only pinned a driver detail that Apple has since changed. This drops it and
spends the coverage where it matters instead — `test_metal_mfa_sdpa_op` gains
causal with `Sq != Sk` and a decode step against a cache, both of which the
suite was missing. Both are bit-exact against the host reference here
(`max_abs=0.000000`), i.e. the fused path itself is fine on macOS 26.5.

Not one machine's quirk: the probe fails with the same numbers on an M1 Pro
(macOS 26.5.2) and an M4 (macOS 26.6).

No behaviour change. `cargo test -p tract-metal --release` on the M1 Pro:
76 passed, 0 failed (was 1 failed); fmt and clippy clean.

🍍

### DIFF
--- a/metal/src/kernels/matmul/mfa/mod.rs
+++ b/metal/src/kernels/matmul/mfa/mod.rs
@@ -454,7 +454,8 @@ pub fn mfa_attention_head_major(
 
 /// Metal device op: fused SDPA over `[B,H,Sq,D]` Q / `[B,H,Sk,D]` K,V via the
 /// vendored MFA kernel. `B>1` is folded to `B*H` heads; causal is realized as a
-/// `[Sq,Sk]` additive mask (the `triangular` constant is a no-op).
+/// bottom-right aligned `[Sq,Sk]` additive mask, never the kernel's `triangular`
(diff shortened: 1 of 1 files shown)


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali (approved):

PR #2551, opened 2026-08-01
metal: mlx's preferred implicit-GEMM conv kernel (1.2-75x)

czoli1976:
Stacked on #2549 and #2550 — their commits are the first two here.

#2549 ported mlx's *general* implicit-GEMM convolution. That is mlx's fallback:
its own routing prefers a second kernel whenever the channel counts are aligned
(`C ≤ 4 or C%16 == 0`, and `O ≤ 16 or O%16 == 0`), which is most real shapes.
This ports that one too and routes the way mlx does, so the general kernel now
only takes what the specialised one declines.

The specialised kernel drops the per-output-position jump tables and is
instantiated on small channel counts (1–4) and short filters, which is where the
difference mostly comes from.

### Numbers

`bench_conv_specialized` (added here), M1 Pro, ms per dispatch, best of 10.
Both kernels are already checked against the CPU op by the existing tests.

| shape | general | specialised | |
|---|---|---|---|
| 299×299×3 → 32, 3×3 s2 | 0.950 | 0.213 | **4.46×** |
| 147×147×32 → 64, 3×3 | 1.564 | 0.618 | **2.53×** |
| 73×73×80 → 192, 3×3 | 1.229 | 0.601 | **2.05×** |
| 112×112×32 → 64, 3×3 | 0.384 | 0.218 | **1.76×** |
| 8×8×384 → 384, 3×3 | 0.367 | 0.255 | **1.44×** |
| 35×35×64 → 96, 3×3 | 0.154 | 0.128 | **1.20×** |

f16 tracks within a few percent. Every shape measured improves; the first-layer
case gains most, since three input channels hit the channel specialisation.

End to end on Inception v3 (TF, NHWC, 299×299), `--metal`, M1 Pro:

| | ms/inference |
|---|---|
| main | 477.7 |
| #2549 + #2550 | 44.8 |
| with this | 40.6 |

So **1.10×** on top of the stack, **11.8×** against main. Same predicted class as
the CPU path.

### Grouped convolutions — withdrawn

**This PR no longer takes them.** It did, and it was wrong: the rewrite relabelled a moved kernel as `OHWI`, and for `group > 1` that label does not describe the tensor. tract's `OHWI` puts the group on the **input** axis, so its grouped kernel is `[O/group, kH, kW, C]`; the ported kernel wants each group's weights contiguous, which is `[O, kH, kW, C/group]` — the group on the **output** axis, as `B += tid.z * N * K` assumes. The two interleave the group differently, so the relabelled tensor claimed a layout it did not have and grouped convolutions came out wrong.

Caught by @kali against `cargo test -p test-metal`: 35 conv proptests failing, all with `group > 1`. With grouped convolutions back on the direct kernel, **23622 passed, 0 failed**.

I looked at repairing it rather than withdrawing it, and did not like any of the three options. A relabel cannot work, since the layouts are permutations of each other rather than views. Crafted `wt_strides` cannot work either: the group offset would need a `(o / O_per_group) * C_per_group` term, which is not linear in the output channel. Per-group dispatches do not work because the kernel derives its h/w step from `C / groups`, so each group would still need its weights materialised in the kernel's order. That leaves teaching the kernel tract's layout, which means changing weight addressing in three loader variants inside a 130 KB verbatim port of mlx's kernel — forking it from upstream, and losing the ability to pull their fixes, which is much of why porting it was attractive.

So the numbers below are gone from this PR. If you think grouped convolutions are worth forking the port for, say so and I will do it as its own change with its own justification.

The grouped test cases stay. They now cover the direct kernel serving those shapes, which is the only grouped-conv coverage metal has.

### Validation

`cargo test -p tract-metal --release`: 98 passed, one failure —
`test_mfa_attention_causal_const_is_noop`, pre-existing on main and unrelated
(#2546). The existing conv tests now exercise both kernels: aligned channels go
to the specialised one, and the unaligned case (C=5, O=7) still covers the
general path, plus grouped cases (2/4/8 groups, stride 2, 1×1, f16) and a
gate test for the grouped shapes mlx would decline. A large-activation case
(1024×1024×32 → 64) covers the shape arithmetic mlx has been fixing upstream
(ml-explore/mlx#3938, #3893). fmt and clippy clean.

M1 Pro only, like #2550 — my M4 was asleep. Happy to add second-device numbers
before this lands if you would rather have them.

the source header, flattened from the same pinned commit as #2549.

🍍


czoli1976:
Added a large-activation case (1024x1024x32 -> 64) to the conv tests. mlx has been fixing 32-bit shape arithmetic in this area upstream (ml-explore/mlx#3938, #3893); this port shares that 32-bit arithmetic, and the headroom is wide — the worst realistic shape I tried gives ~131k tiles against i32's 2.1e9 — but the case is covered now rather than assumed.

kali:
Hey, we have a mismatch in axes convention...

rewrite_conv_kernel_metal (metal/src/transform.rs:122) relabels the kernel as KernelFormat::OHWI after a plain move_axis:

- OIHW grouped is [O, C/group, H, W] → move_axis(1,3) = [O, H, W, C/group]
- HWIO grouped is [H, W, C/group, O] → move_axis(3,0) = [O, H, W, C/group]

That is what the mlx kernel indexes, but it is not what tract's OHWI means: for OHWI the group sits on the I axis, so the legal grouped shape is [O/group, H, W, C] 

Mayve the freestanding routines in core/src/ops/cnn/conv/mod.rs can help.

It's caught by `cargo test -p test-metal`.

Are you ok taking this one @czoli1976 ?

czoli1976:
ok .... the fix here is to stop sending all grouped-conv as will be too much to deal with and I guess not worth it and keep the single-group implicit-GEMM kernel

czoli1976:
Taken — thank you, that was a real bug and a good catch.

You are right about the convention. `OHWI` puts the group on the **I** axis, so its grouped kernel is `[O/group, kH, kW, C]`; the ported kernel wants each group's weights contiguous, `[O, kH, kW, C/group]`, which is what `B += tid.z * N * K` assumes. The rewrite produced the second and labelled it the first, so for `group > 1` the tensor did not mean what the op said it meant. `cargo test -p test-metal` reproduced it exactly as you said: 35 conv proptests, all `group > 1`.

**Grouped convolutions now stay on the direct kernel** — `mlx_conv_eligible` and `mlx_conv_dispatchable` both reject `group != 1`, so they are neither relabelled nor dispatched. `test-metal` is **23622 passed, 0 failed**, against 23587/35 before. The single-group path, which is the point of the PR, is untouched. Title and description updated; the grouped numbers are withdrawn rather than left standing.

I did look at keeping them before giving up on it. The freestanding routines you pointed at describe the layouts precisely, and that is what convinced me the two cannot be bridged cheaply: they are permutations of each other, not views, so no relabel works; crafted `wt_strides` cannot express the group offset either, since it needs a `(o / O_per_group) * C_per_group` term that is not linear in the output channel; and per-group dispatches still need each group's weights materialised in the kernel's order, because the kernel derives its h/w step from `C / groups`. What is left is teaching the kernel tract's layout — three loader variants inside a 130 KB verbatim port. That forks it from mlx upstream and gives up pulling their fixes, which was much of the appeal. Happy to do it as its own change if you think grouped is worth that, but it did not seem right to bundle.

The grouped test cases stay; they now cover the direct kernel on those shapes.

kali:
@czoli1976 missed a spot ?

PR #2555, opened 2026-08-02
Resize PR 2: Complete ONNX Resize for opset 18/19, and make it fast

czoli1976:
Implements the ONNX Resize attributes tract was missing — `antialias`, `axes`, `keep_aspect_ratio_policy`, `half_pixel_symmetric`, `tf_crop_and_resize` + `roi`/`extrapolation_value` — and replaces the per-element evaluator with a precomputed per-axis tap/weight plan, which is 2.7× faster on the op and 1.45× end-to-end on a Resize-heavy model.

**Stacked on #2554** — its commit is the first of the two here, so please merge that one first; this PR's own diff is the second commit.

### Conformance

Enabling the whole ONNX Resize node set gives the before/after (each counted across both opset checkouts and both runtimes):

| | before | after |
|---|---|---|
| Resize node tests | 84 passed / **64 failed** | **148 passed / 0 failed** |

The 18 distinct failures fell into five groups, and the failure modes were not equally visible:

- `antialias` — parsed nowhere, so it was **silently dropped**: downsampling models got aliased output with no error.
- `keep_aspect_ratio_policy` — same, not parsed at all.
- `axes` — parsed into a field nothing read, then failed shape inference.
- `half_pixel_symmetric` (opset 19) — rejected at parse.
- `tf_crop_and_resize` / `roi` / `extrapolation_value` — rejected at parse.

Semantics were taken from the ONNX reference implementation (`onnx/reference/ops/op_resize.py`), including the `ratio = 1` convention at integer source coordinates, edge-padded neighbour selection, `exclude_outside` renormalisation, and the antialias footprint widening to `1/scale`.

### Evaluator

`eval` used `ArrayD::from_shape_fn` with a dynamic-rank index per output element, recomputing the coordinate transform (and a division) per pixel. It now builds one `AxisPlan` per axis — clamped tap indices plus weights — and resamples over contiguous inner runs, which is also what makes the widened antialias windows affordable. Both the core and the ONNX op share the plan builder; each keeps its own kernel and tie-break set, so the core/ONNX split is unchanged.

`axes` is spread to full rank at declutter time when the aux input is constant, so the core op still takes over those nodes.

### Numbers

u2netp (rembg, 38 bilinear Resize nodes), 1×3×320×320, release, interleaved A/B, min of 5 rounds:

| | Resize op | whole network |
|---|---|---|
| main | 302.4 ms | 605.4 ms |
| this branch | 107.6 ms | **419.1 ms** |
| | **2.7×** | **1.45×** |

All seven model outputs are bit-identical to main (max abs diff 0.0), so this is overhead removal, not a numerics change. Resize was 49% of optimised runtime on this model before, 26% after.

### Testing

`onnx-tests.sh` on `1_13_0` and `1_19_1`, `test-unit-core`, `test-f16`, `test-metal`, and the crate unit tests all green. `cargo fmt --all` and `cargo clippy --workspace` clean. Note that `cargo check --all-targets` fails on aarch64 for an unrelated pre-existing reason (`linalg/benches/avx512_zombies.rs` is x86-only).

🍍

### DIFF
--- a/core/src/ops/nn/resize.rs
+++ b/core/src/ops/nn/resize.rs
@@ -1,7 +1,7 @@
 use crate::internal::*;
 use crate::ops::array::Tile;
 
-/// Maps an output coordinate back to the input axis. The four ONNX coordinate
+/// Maps an output coordinate back to the input axis. The ONNX coordinate
 /// transformation modes that have a well-defined inverse without an input ROI.
 #[derive(Clone, Debug, Hash, PartialEq, Eq)]
 pub enum CoordTransformer {
@@ -9,6 +9,7 @@ pub enum CoordTransformer {
     AlignCorners,
     Asymmetric,
     PytorchHalfPixel,
+    HalfPixelSymmetric,
 }
 
 impl CoordTransformer {
@@ -28,8 +29,13 @@ impl CoordTransformer {
                 if len_out > 1 {
                     (x_out as f32 + 0.5) / scale - 0.5
                 } else {
-                    0.0
+                    -0.5
                 }
+            }
+            CoordTransformer::HalfPixelSymmetric => {
+                let adjustment = len_out as f32 / (scale * len_in as f32);
+                let offset = len_in as f32 / 2.0 * (1.0 - adjustment);
+                offset + (x_out as f32 + 0.5) / scale - 0.5
             }
         }
     }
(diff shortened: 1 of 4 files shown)


kali:
Rebased after the merge of the first resize PR.

kali (approved):

kali:
Thanks a lot for working on the Resize.

PR #2556, opened 2026-08-02
Resize PR 3: Run Resize on the Metal GPU

czoli1976:
Resize had no Metal kernel, so in a GPU graph every node round-tripped its tensor through the host. This adds a resample-one-axis kernel that consumes the same per-axis tap/weight plan the CPU op builds, which makes one kernel enough for nearest, linear, cubic and their antialiased variants — they differ only in window size and weights.

Needed the `AxisPlan` from #2554 and #2555; now that both are merged this is rebased onto main and stands alone as a single commit.

### Wiring

The plan depends only on shapes and attributes, so it is baked at translation time; the node then keeps just its data input. That is also why this is wired in `translate_node` rather than through `register_metal_op!`: the macro rewires the *same* inputs, and Resize's `scales`/`sizes` constant is a `TDim`, whose datum type fails the device-dtype gate in `try_make_metal_op` — the node was rejected before any translator ran. Nodes with a symbolic shape or a non-constant scales input aren't translated and stay on CPU.

### Effect

u2netp (rembg, 38 bilinear Resize nodes), 1×3×320×320, `--metal -O`, min of 3 interleaved rounds:

| | before | after |
|---|---|---|
| Resize nodes on GPU | 0 / 38 | **38 / 38** |
| host round-trips (`DeviceSync*`) | 150 | **74** |
| end-to-end | 560.7 ms | **460.7 ms** (1.22×) |

Outputs match the CPU runtime to 6e-7 max abs (f32 accumulation order).

Two caveats I'd rather state than have you find:

- **This does not make Metal beat CPU on this model** — 461 ms vs 419 ms on the same machine. The remaining `DeviceSyncToHost` is 91% of the Metal profile because 33 `OptMaxPool` nodes still have no Metal kernel and force 40 host round-trips per inference. Resize was necessary but not sufficient; MaxPool is the next one that matters.
- Per-op Metal timings in `dump --profile` are encode time, not GPU execution (the sync nodes absorb that), so I'm quoting only the end-to-end figure.

`GpuResize` is backend-agnostic and CUDA would need only a launch function, but there's no NVIDIA GPU on my machine, so I haven't written code I can't run. Happy for someone with the hardware to take it.

### Testing

`test-metal` 22762 passed / 0 failed (630 of them Resize node tests, which exercise the kernel — I checked the nodes actually land on `MetalResize` rather than silently falling back). `onnx-tests.sh 1_19_1` green. `cargo fmt --all --check` and `cargo clippy` clean.

🍍

### DIFF
--- a/core/src/ops/nn/resize.rs
+++ b/core/src/ops/nn/resize.rs
@@ -356,7 +356,8 @@ impl Resize {
         );
     }
 
-    fn plan_axis(&self, scale: f32, len_in: usize, len_out: usize) -> AxisPlan {
+    /// The resampling plan for one axis under this op's interpolator.
+    pub fn plan_axis(&self, scale: f32, len_in: usize, len_out: usize) -> AxisPlan {
         let window = window_size(&self.interpolator, false, scale);
         let coord = |x| Some(self.coord_transformer.transform(x, scale, len_in, len_out));
         match self.interpolator {
--- a/gpu/src/ops/mod.rs
+++ b/gpu/src/ops/mod.rs
@@ -17,6 +17,7 @@ pub mod leaky_relu;
 pub mod pad;
 pub mod pulse;
 pub mod reduce;
+pub mod resize;
 pub mod rms_norm;
 pub mod rotate_half;
 pub mod scaled_masked_softmax;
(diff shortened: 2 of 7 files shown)


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali (approved):

PR #2557, opened 2026-08-02
Run 2D MaxPool on the Metal GPU

czoli1976:
MaxPool had no Metal kernel, so it round-tripped its tensor through the host in the middle of a GPU graph. This adds a kernel that pools two spatial axes with every axis stride passed explicitly, so NCHW and NHWC share one path.

**Stacked on #2554, #2555 and #2556** — this branch carries their commits and this PR's own change is the last one. It only depends on #2556 for the benchmark story, not for the code, so it can also be taken on its own if that's easier.

### Effect

The point of this one is not the kernel's own throughput, it's what it removes. u2netp (rembg), 1×3×320×320, `--metal -O`, min of 3 interleaved rounds:

| | host round-trips (`DeviceSync*`) | end-to-end |
|---|---|---|
| main | 150 | 560.7 ms |
| + Metal Resize (#2556) | 74 | 460.7 ms |
| + Metal MaxPool (this) | **8** | **359.8 ms** |

The graph is now essentially GPU-resident, and that is what makes Metal win on this model: **359.8 ms vs 419.1 ms on CPU**. It lost before either kernel (560.7) and still lost with Resize alone (460.7) — the residual host transfers dominated. Resize was necessary, MaxPool was what flipped it. Together, 1.56× on the Metal path.

Outputs match the CPU runtime to 5.96e-7 max abs across all seven model outputs.

### Notes

Padding positions are skipped rather than compared, so a window lying entirely in the padding keeps the CPU op's `min_value()` result.

The translator registers for **both `MaxPool` and `OptMaxPool`**. The Metal transform runs before the optimize pass, so the node is still `MaxPool` at translation time even though a `-O` dump shows `OptMaxPool` — registering only for the optimized op silently never fires. `OptMaxPool` is added to `core::ops::cnn`'s public re-exports for this.

Two limits worth knowing:

- Scoped to **rank 4, two spatial axes, no index outputs, concrete shapes**. 1D/3D pooling, `with_index_outputs` and symbolic shapes aren't translated and stay on CPU.
- One thread per output element with a 1×1×1 threadgroup, matching the existing Gather/DiagGather kernels. A tuned threadgroup size probably has more in it; I left that alone because the transfer elimination is the dominant term here and I didn't want to tune on a single model.

### Testing

`test-metal` 22762 passed / 0 failed (558 of them pooling node tests; I checked the nodes actually land on `MetalMaxPool` rather than silently falling back to CPU). `onnx-tests.sh 1_19_1` and `test-unit-core` green. `cargo fmt --all` and `cargo clippy` clean.

🍍

### DIFF
--- a/core/src/ops/cnn/mod.rs
+++ b/core/src/ops/cnn/mod.rs
@@ -11,7 +11,7 @@ mod sumpool;
 
 pub use self::conv::{Conv, KernelFormat};
 pub use self::deconv::Deconv;
-pub use self::maxpool::MaxPool;
+pub use self::maxpool::{MaxPool, OptMaxPool};
 pub use self::padding::PaddingSpec;
 pub use self::patch_axis::PatchAxis;
 pub use self::patches::{Patch, PatchSpec};
--- a/core/src/ops/nn/resize.rs
+++ b/core/src/ops/nn/resize.rs
@@ -352,7 +352,8 @@ impl Resize {
         );
     }
 
-    fn plan_axis(&self, scale: f32, len_in: usize, len_out: usize) -> AxisPlan {
+    /// The resampling plan for one axis under this op's interpolator.
+    pub fn plan_axis(&self, scale: f32, len_in: usize, len_out: usize) -> AxisPlan {
         let window = window_size(&self.interpolator, false, scale);
         let coord = |x| Some(self.coord_transformer.transform(x, scale, len_in, len_out));
         match self.interpolator {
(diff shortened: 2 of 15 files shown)


czoli1976:
Marking this draft: it overlaps #2553, which I opened earlier and had not checked before writing this. #2553 already adds Metal max and sum pooling, with a better-coalesced NHWC dispatch.

The two are not identical, though — #2553 deliberately leaves NCHW on the CPU path, and this one covers NCHW, which is what an ONNX/PyTorch export like u2netp uses. I will rework the useful part of this as NCHW support on top of #2553 rather than ask anyone to review two pooling kernels.

czoli1976:
Superseded by #2553, which now carries channels-first max pooling as a second commit. Closing this rather than asking anyone to review two pooling kernels.

PR #2558, opened 2026-08-03
cli: add /ci bench-mt-ladder thread-scaling run

kali:
Dispatch bench.yml with --threads-ladder from a /ci bench-mt-ladder [ladder] comment (default 2,4,0) and render a self-contained bench-mt-report table (evaltime + speedup vs the serial base per thread count) instead of the vs-main comparison.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -15,6 +15,11 @@ on:
         description: "Optional PR number to bench (from fork ok). Leave empty to run on selected branch."
         required: false
         type: number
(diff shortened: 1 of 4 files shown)


kali:
Unknown `/ci` target: `bench-mt-ladder`.

Usage: `/ci full`, `/ci large-models` (aliases: `large`, `llm`), `/ci wheels` (aliases: `python`, `py`), `/ci examples` (alias: `ex`), or `/ci bench` (alias: `benches`).

PR #2560, opened 2026-08-03
feat(linalg): add AVX2+FMA SiLU f32 kernel

cverrier:
x86 hosts with AVX2+FMA but no AVX-512 ran SiLU through the generic scalar kernel; they now get a vectorized one, which improves the throughput. Accuracy follows the FMA sigmoid f32 kernel.

Also, I benched the different SiLU f32 kernels on an AVX-512 machine:

```bash
cargo bench -p tract-linalg --bench silu -- silu_f32
```

Throughput results:

```
silu_f32/linalg         thrpt:  [1.8691 Gelem/s 1.8697 Gelem/s 1.8703 Gelem/s]

silu_f32/linalg-asm-fused
                        thrpt:  [3.0071 Gelem/s 3.0179 Gelem/s 3.0314 Gelem/s]
```

which gives a nice ~61% increase in throughput: this was expected because the `x86_64_avx512_silu_f32_16n` kernel is composed, not fused — it calls `avx512_sigmoid_f32` directly. Therefore, on AVX-512 — which is backward compatible with AVX2+FMA — we can fall back to this new AVX2+FMA SiLU kernel for now; see #2567, where this change will be implemented.

Later, I will also implement a fused version for AVX-512.

### DIFF
--- a/linalg/benches/activations_avx512.rs
+++ b/linalg/benches/activations_avx512.rs
@@ -2,9 +2,10 @@
 // their x86 predecessor.
 //
 //   sigmoid, tanh              : predecessor = FMA (256-bit, 8-wide) kernel
-//   hardswish, leaky_relu,
-//   silu, gelu                 : predecessor = generic scalar kernel
+//   hardswish, leaky_relu, gelu: predecessor = generic scalar kernel
 //                                (no FMA kernel exists on x86)
+//   silu                       : predecessor = generic scalar kernel
+//                                (the FMA kernel is benched in silu.rs)
 //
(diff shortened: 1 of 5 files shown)


cverrier:
@kali A few checks got cancelled, looks unrelated to the PR itself. Could you re-run them when you get a chance?

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

cverrier:
@kali Is this work ready to be merged?

kali (approved):

PR #2564, opened 2026-08-03
WASM PR 2: move the WASM microbenches out of src and into the bench target

czoli1976:
Stacked on #2563 — its two commits are the first two here.

The wasm backend kept three timing modules inside `src/` behind `#[cfg(test)]` + `#[ignore]`, which no other backend does. `benches/wasm.rs` already supersedes two of them, and says so itself: its header records that the in-`src` version looped all four GEMV kernels back-to-back and so biased whichever ran last. Those two go; the activation bench moves across to join it.

`numerical_consistency_16x1_vs_32x1` stays in `src`, in `wasm/dispatch_tests.rs` — it is an assertion rather than a timer, and the only numerical cross-check between two kernels.

The four copies of the kernel-lookup helper in `benches/wasm.rs` collapse to one. I left the timing loops alone: they use 50 versus 200 warmup calls and only two of them do min-of-N reporting, and those differences are deliberate, so a shared helper would have quietly changed what the numbers mean.

Net: 335 lines out of `src`, and `wasm.rs` is down to 78 lines.

Nothing here ships — test and bench code only, and the kernel disassembly is unchanged. The bench runs: on an M1 Pro the relocated activation bench reports sigmoid 1.49–1.54x and tanh 1.40–1.62x over the scalar path, matching what the module documented.

🍍


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali (approved):

PR #2565, opened 2026-08-03
WASM PR 3: test the invariants the WASM kernels rely on, and run CI against relaxed-simd

czoli1976:
Stacked on #2564.

Two gaps in how the wasm backend is covered, both of which matter before anything touches the kernels.

**CI only builds one of the two configurations.** The backend chooses its multiply-add form, its int8 packing and its sigmoid/tanh kernels at compile time on `+relaxed-simd`, so the single `+simd128` leg in `.travis/cross.sh` leaves roughly half of it never executed. This adds a second leg. I ran it: `tract-linalg` 2208, `tract-core` 816 + 267, `test-unit-core` 9 + 1, all green. It does cost a second wasm build and about 2.5 minutes of test time per run — the one thing in this series that costs you something, so say if you would rather not.

**Nothing checked that `AddRowColProducts` and `AddMatMul` agree on fusion.** With `k = 1` they compute the same `c[i][j] += a[i] * b[j]`, so within a kernel they have to make the same choice about the fused multiply-add — and today they do, in both directions: 32x1 and 8x8 fuse both arms, the other four fuse neither. Nothing held that in place. The existing `mmm_kernel_fuse_tests!` cannot: it uses small integers whose products are exact, where fused and separate agree trivially. The new test picks `a = b = 1 + 2^-12` against `c = -1`, so the product needs 25 significand bits and the two forms differ by one ulp, then compares the two arms bit-for-bit. Under `+simd128` it is vacuous but still true; under `+relaxed-simd` it bites. I confirmed it bites by flipping 32x1's `AddRowColProducts` to the non-fused macro — it fails on exactly that kernel.

Also records why `wasm_f32_4x4` sits at `TargetOptimized`. It is the only kernel not at `ManuallyOptimized`, so `strategize`'s `retain()` drops it and neither `mmm_f32` nor `mmv_f32` names it — it is unreachable through dispatch and reached only by name from its generated tests. It is worth keeping there as the second f32 kernel with a genuinely two-dimensional C tile, which is what gives the generated store and packing tests a second layout to cover, so this documents the arrangement and asserts dispatch never returns it rather than changing anything.

🍍


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2566, opened 2026-08-03
WASM PR 4: fold the repeated fused-op arms into macros

czoli1976:
Stacked on #2565. This is the point of the series.

Each of the six f32 kernels spelled the same 24 `FusedKerSpec` arms out once per accumulator, so 8x8 needed 535 lines to say what 4x1 says in 86, and widening a kernel meant copying every arm again. Of 2342 kernel lines, 2005 were those arms and only 337 were the `AddMatMul` loops that actually differ.

The arms that apply one operation uniformly across the accumulators now expand from `wasm/fuse.rs`. Each macro emits the same statements in the same order, so this is a source-level change only. Per kernel the 24 arms become 24 lines.

The two layouts need different primitives, and which one an arm reaches for is visible on the arm's own line rather than hidden behind a flag: in the GEMV kernels a per-row operand is a vector load and a per-column one is a splat, and in the GEMM kernels it is the other way round. 8x8 holds each row as a low/high pair that takes one splat between them.

Five arms stay written out per kernel. `AddMatMul` and `AddRowColProducts` both carry the fused-versus-separate multiply-add choice, and keeping them adjacent and explicit is what makes that choice checkable by eye — #2565 added the test that enforces it. `AddUnicast` and `Store` walk the C tile by stride differently in each kernel. `LoadTile` differs too: 4x1 uses an unaligned `v128_load` where the others dereference, so a shared macro would have changed the alignment immediate for no gain.

The last commit corrects documentation that had drifted: `WASM_RELAXED_SIMD.md` said all six f32 kernels switch to fused multiply-add under `+relaxed-simd`, when only 32x1 and 8x8 do and the quoted 1.40–1.55x applies only to those; the header on `wasm_i32_4x4` claimed bit-identity with `generic_i32_4x4` without qualification, which holds only *without* `+relaxed-simd`, since the relaxed dot takes an i7 second operand; and the module gate moves from `target_family = "wasm"` to `target_arch = "wasm32"`, matching what the kernels require.

## Evidence

Same gate as #2563, and it is what makes this reviewable: **the emitted body of all eight kernels is byte-identical**, under `+simd128`, under `+simd128,+relaxed-simd`, and in the shipped `release` profile with `lto = true`. So the 1153 deleted lines can be read as deleted repetition rather than as a rewrite of six hand-tuned SIMD kernels.

Benchmarks are a guard here rather than the claim, since the instructions are unchanged. Best-of-two runs of `benches/wasm.rs` across 24 shapes: median +0.5%, range −1.4% to +2.1%.

Tests: 2195 under `+simd128`, 2208 under `+simd128,+relaxed-simd`, with `tract-core` and `test-unit-core` green in both. No new clippy lints.

`wasm_i32_4x4` is deliberately left alone — it already loops over a scalar `[[i32; 4]; 4]` at about 7 lines an arm, so it never had this problem, and collapsing it would mean a second macro family for scalar accumulators.

Across the four PRs `wasm.rs` goes from 3159 lines in one file to 2196 across eight.

🍍


czoli1976:
@kali do you think you will have time on Tuesday for this so to finish the  WASM Refactor ? If so I can rebase on monday evening

kali:
missed the call-out, dealing with it :)

czoli1976:
Mercie !

PR #2567, opened 2026-08-03
perf(linalg): run SiLU through the fused FMA kernel on AVX-512 hosts

cverrier:
AVX-512 hosts ran silu_f32 and silu_f16 through a kernel composed over the zmm sigmoid, which pays a scratch copy and an extra traversal that the fused FMA kernel avoids. They now use the fused kernel at both widths. The composed kernel has no remaining users, and is removed.


cverrier:
@kali Have you seen this one?

cverrier:
Just rebased; merge conflicts fixed

kali (approved):

PR #2570, opened 2026-08-03
linalg/mmm: take the chunking slack only where the cache absorbs it

czoli1976:
Follow-up to #2534: the a53 `inceptionv3 · pass_mt` regression looks like the
chunk *count*, and this gates that count on whether the machine has a cache that
can absorb the re-reads it costs.

## Why a53 and not the rest of the fleet

The a53 number is identical to three digits across the 09:12 and 11:02 runs on
#2534 — 1.12 s → 1.26 s both times. The L2-sharers commit only does anything when
`l2_sharers > 1`, so it was a no-op there. If `cache_info().l2 == 0` on that
runner, `inner_tier_pays` short-circuits, `blk` and `blk_outer` are both
`usize::MAX`, and `run_blocked` is a plain naive walk — which leaves `chunk_grid`
as the only thing that changed on that device.

Replaying old vs merged over InceptionV3's 94 threaded convs at `nth = 4`,
`arm64simd_mmm_f32_12x8_a53` (m = Cout, k = Cin·kh·kw, n = Hout·Wout):

```
    m     k     n |  panels  |   old  #  |   new   #  |  oldMB  newMB    ×
  384  4032    64 |  32x8    |  4x1   4  |  16x1  16  |   10.3   22.7  2.20
  384  2592   289 |  32x37   |  1x4   4  |   4x4  16  |   18.9   27.9  1.48
  320  2048    64 |  27x8    |  4x1   4  |  14x1  14  |    4.7   10.0  2.11
  192  1344   289 |  16x37   |  1x4   4  |   3x5  15  |    5.7    9.8  1.73
   96   864  1225 |   8x154  |  1x4   4  |  1x16  16  |    5.6    9.5  1.72

packed-operand traffic per inference: 458 MB → 707 MB (1.54×), +248 MB
```

+248 MB at 1.5–2.5 GB/s of effective DRAM bandwidth is +100 to +165 ms against a
measured +140 ms.

The cause is two rules meeting rather than the `sqrt` shaping. The old
`nchunks < 4·nth` was a *re-plan trigger*, and what it re-planned *to* was `nth`.
On grids below the old 16-panel cap — nearly all of InceptionV3 — the merged code
therefore asks for 4× the chunks the old code chose, and on a part with a 512 KB
cluster L2 and no L3 each extra chunk is a full re-read of a packed operand from
DRAM. That is also why a53 and not a55: a55 sits in a DynamIQ cluster with a DSU
L3 behind it, so the same re-reads hit cache.

## The change

`chunks_per_thread()` gates `CHUNKS_PER_THREAD` on the detected last-level cache —
falling back to L2 where no LLC is exposed, which is the case on Apple — and takes
one chunk per thread below 2 MB. `TRACT_MMM_CHUNKS_PER_THREAD` overrides it,
resolved once like the cache probe it reads, so the fleet can sweep 1/2/4 without a
rebuild.

Above 2 MB of last-level cache the chunk grid is bit-for-bit what it is today, so
this cannot move the m1-max, i9 or Orin numbers. Below it, the grid returns to the
old chunk count:

```
default (cpt=4)                      forced cpt=1 (what an a53 gets)
384x4032x64:  16x1 = 16 ch, 22.7 MB   4x1 = 4 ch, 10.3 MB
384x2592x289:  4x4 = 16 ch, 27.9 MB   2x2 = 4 ch, 14.0 MB
192x1344x289:  3x5 = 15 ch,  9.8 MB   1x4 = 4 ch,  5.7 MB
  total 79.9 MB                         total 40.2 MB
```

It covers the a53 whichever way the cache probe goes there: if L2 is undetected,
`cache_info().l2 == 0` drives the count to 1 anyway.

## Tests

`cargo test -p tract-linalg --features multithread-mm` passes 4420 tests across all
targets. `harness/nnef-test-cases/parallel-matmul` passes at `--approx exact` for 2,
3 and 8 threads, run both on the default path and under
`TRACT_MMM_CHUNKS_PER_THREAD=1` — the second is the one that matters, since forcing
the constant is the only way to exercise the a53 chunk counts on Apple hardware.

`cargo test --workspace` gives 29 suites clean plus test-f16's own 4747, with one
failure — `nnef_f16::onnx::node::v1_19_1::test_averagepool_2d_ceil_last_window_starts_on_pad`
— which fails identically on `a6c4243c6` with the diff reverted, so it is
pre-existing and unrelated. `test-tflite`, `tract-proxy`, `tract-proxy-sys` and
`test-cuda` could not build or run on this host (no OpenSSL, no FFI dylib, no CUDA
driver) and were excluded. Nothing outside `linalg` can see this change in any case:
it is entirely inside `#[cfg(feature = "multithread-mm")]`, which is off by default.

## What I could not check

No wall clock. On a 12 MB-L2 M1 Pro `chunks_per_thread()` resolves to 4, so this
diff is inert on every machine I have; the traffic numbers above are arithmetic
over canonical InceptionV3 conv geometry, not measurements. A bench run is the
thing that would settle it, and `inceptionv3 · pass_mt` on cortex-a53 is the metric
to watch.

`CHUNK_SLACK_LLC_BYTES = 2 MB` is a calibration guess, chosen to put Orin's 2 MB L3
on the slack side and A7/A9/A53 off it. The env knob is there so a sweep can answer
where that boundary belongs rather than a rebuild per value — and if the constant
survives its bench run it should become a proper `declare_knob!`.

🍍

### DIFF
--- a/linalg/src/frame/mmm/mod.rs
+++ b/linalg/src/frame/mmm/mod.rs
@@ -655,18 +655,52 @@ unsafe fn run_with_scratch_space_2d<K: MatMatMulKer>(
     }
 }
 
-/// Chunks per thread the 2D dispatch aims for. Above one, rayon can steal work
-/// when threads progress unevenly — a contended core, an E-core on a big.LITTLE
-/// part, a chunk carrying more border tiles — so a straggler costs at most its
-/// share rather than the whole grid's tail. Each extra chunk also re-reads a band
-/// of the packed operands, and that cost grows as `sqrt(chunks)` against a linear
-/// gain in slack, which is what keeps this small.
+/// Chunks per thread the 2D dispatch aims for on a part whose last-level cache
+/// can absorb the re-reads (see [`chunks_per_thread`]). Above one, rayon can steal
+/// work when threads progress unevenly — a contended core, an E-core on a
+/// big.LITTLE part, a chunk carrying more border tiles — so a straggler costs at
+/// most its share rather than the whole grid's tail. Each extra chunk also
+/// re-reads a band of the packed operands, and that cost grows as `sqrt(chunks)`
+/// against a linear gain in slack, which is what keeps this small.
 #[cfg(feature = "multithread-mm")]
 const CHUNKS_PER_THREAD: usize = 4;
 
+/// Smallest last-level cache that makes the extra chunks worth their re-reads.
+/// Below this a part has only a small cluster L2 and no LLC behind it
+/// (Cortex-A7/A9/A53), so the band a chunk re-reads comes from DRAM every time.
+#[cfg(feature = "multithread-mm")]
+const CHUNK_SLACK_LLC_BYTES: usize = 2 * 1024 * 1024;
+
+/// Chunks per thread [`chunk_grid`] aims for on this machine.
+///
+/// An extra chunk buys load-balance slack and costs one more pass over a packed
+/// operand. That pass is a cache hit on a part with a last-level cache big enough
+/// to hold the band and a DRAM round trip on one without, so the slack is only
+/// worth taking above [`CHUNK_SLACK_LLC_BYTES`]; below it the dispatch takes one
+/// chunk per thread.
+///
+/// `TRACT_MMM_CHUNKS_PER_THREAD` overrides the choice. Resolved once, like the
+/// cache probe it reads, so a later change to the variable has no effect.
+#[cfg(feature = "multithread-mm")]
+fn chunks_per_thread() -> usize {
+    use std::sync::OnceLock;
+    static CPT: OnceLock<usize> = OnceLock::new();
+    *CPT.get_or_init(|| {
+        if let Some(n) =
+            std::env::var("TRACT_MMM_CHUNKS_PER_THREAD").ok().and_then(|v| v.trim().parse().ok())
+        {
+            return usize::max(n, 1);
+        }
+        let llc = crate::cache::last_level_cache()
+            .map(|(bytes, _)| bytes)
+            .unwrap_or_else(|| crate::cache::cache_info().l2);
+        if llc >= CHUNK_SLACK_LLC_BYTES { CHUNKS_PER_THREAD } else { 1 }
+    })
+}
+
 /// Chunk grid for the 2D dispatch: `(nchunks_m, nchunks_n, dr_m, dr_n)`.
 ///
-/// Aims for [`CHUNKS_PER_THREAD`]`· nth` chunks and shapes them to minimise how
+/// Aims for [`chunks_per_thread`]`· nth` chunks and shapes them to minimise how
 /// often the packed operands are re-read: a chunk covering `dr_m × dr_n` panels
 /// reads `dr_m·mr·k` of A and `dr_n·nr·k` of B, so over the whole grid A is read
 /// `nchunks_n` times and B `nchunks_m` times. Minimising
@@ -687,7 +721,7 @@ fn chunk_grid(
     nr: usize,
     nth: usize,
 ) -> (usize, usize, usize, usize) {
-    let chunks = (CHUNKS_PER_THREAD * nth).max(1);
+    let chunks = (chunks_per_thread() * nth).max(1);
     let (m, n) = (n_panels_m * mr, (n_panels_n * nr).max(1));
     let nchunks_m = (chunks.saturating_mul(m) / n).isqrt().clamp(1, n_panels_m);
     let nchunks_n = (chunks / nchunks_m).clamp(1, n_panels_n);
--- a/linalg/src/frame/mmm/mod.rs
+++ b/linalg/src/frame/mmm/mod.rs
@@ -629,18 +629,52 @@ unsafe fn run_with_scratch_space_2d<K: MatMatMulKer>(
     }
 }
 
-/// Chunks per thread the 2D dispatch aims for. Above one, rayon can steal work
-/// when threads progress unevenly — a contended core, an E-core on a big.LITTLE
-/// part, a chunk carrying more border tiles — so a straggler costs at most its
-/// share rather than the whole grid's tail. Each extra chunk also re-reads a band
-/// of the packed operands, and that cost grows as `sqrt(chunks)` against a linear
-/// gain in slack, which is what keeps this small.
+/// Chunks per thread the 2D dispatch aims for on a part whose last-level cache
+/// can absorb the re-reads (see [`chunks_per_thread`]). Above one, rayon can steal
+/// work when threads progress unevenly — a contended core, an E-core on a
+/// big.LITTLE part, a chunk carrying more border tiles — so a straggler costs at
+/// most its share rather than the whole grid's tail. Each extra chunk also
+/// re-reads a band of the packed operands, and that cost grows as `sqrt(chunks)`
+/// against a linear gain in slack, which is what keeps this small.
 #[cfg(feature = "multithread-mm")]
 const CHUNKS_PER_THREAD: usize = 4;
 
+/// Smallest last-level cache that makes the extra chunks worth their re-reads.
+/// Below this a part has only a small cluster L2 and no LLC behind it
+/// (Cortex-A7/A9/A53), so the band a chunk re-reads comes from DRAM every time.
+#[cfg(feature = "multithread-mm")]
+const CHUNK_SLACK_LLC_BYTES: usize = 2 * 1024 * 1024;
+
+/// Chunks per thread [`chunk_grid`] aims for on this machine.
+///
+/// An extra chunk buys load-balance slack and costs one more pass over a packed
+/// operand. That pass is a cache hit on a part with a last-level cache big enough
+/// to hold the band and a DRAM round trip on one without, so the slack is only
+/// worth taking above [`CHUNK_SLACK_LLC_BYTES`]; below it the dispatch takes one
+/// chunk per thread.
+///
+/// `TRACT_MMM_CHUNKS_PER_THREAD` overrides the choice. Resolved once, like the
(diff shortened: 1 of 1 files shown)


czoli1976:
@kali maybe will save A53

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali:
inception is pretty reliable signal i think. let's rerun for the rest.

kali:
ok, we have a regression on mobilenet. Any chance I can get fromage et dessert ?

czoli1976:
Not a dietitian here but the gate acts different per machine, not workload so there is the key .....

a fromage-et-dessert fix may look like 

`cpt = 4  if llc >= CHUNK_SLACK_LLC_BYTES          // unchanged, big-cache parts
      4  if (m·k + k·n) · elem_size <= llc        // new: re-reads are cache hits anyway
      1  otherwise`
      
sending a commit soon to try ...

czoli1976:
@kali fromage *et* dessert, hopefully.

The gate was reading the machine only, so on a small-cache part it took the slack away from every dispatch — including mobilenet's, which was paying nothing for it. It now also reads the packed-operand footprint: inceptionv3's convs pack to 2.5–6.9 MB, mobilenet_v2's to 0.3–1.7 MB, so a 2 MB boundary keeps the a53 at exactly the chunk counts that measured −8.2% and puts mobilenet back on main's grid on a7/a9/beaglev.

Untouched above `CHUNK_SLACK_LLC_BYTES`, so m1-max/i9/Orin/a55 still cannot move. `cargo test -p tract-linalg --features multithread-mm` green (4459), parallel-matmul bit-exact at 2/3/8 threads on the default path and under both forced gate paths.

Still a calibration guess — ~1.4× headroom either side — so `TRACT_MMM_CHUNK_SLACK_BYTES` moves the boundary if a sweep wants it elsewhere. Could you run a bench?

kali:
Let me have a look.

czoli1976:
sounds like some progress to me
image

kali:
i'm having claude doing the sweep on the a53.

czoli1976:
@kali mobilenet is fixed, but I owe you a correction on the a53.

From the run's own metrics: mobilenet_v2_1 `pass_mt` is 933 ms / 962 ms / 503 ms on a7 / a9 / beaglev against 930 / 954 / 497 on main — all three regressions gone. But inceptionv3 on a53 only came in at 1.173 s, **−4.6%**, not the −8.2% the machine-only gate got. It dropped under the flag threshold, which is why the improvements table is empty.

I dumped every threaded matmul of both models to find out why, and my "clean 2 MB separation" was wrong — it came from five hand-picked convs, not the model. At 12x8/f32 InceptionV3 spans **0.16–23.8 MB** (median 0.54) and MobileNet v2 **0.20–6.95 MB** (median 0.88). They overlap almost completely. No footprint boundary sorts one model from the other.

What the boundary really sets is the share of each model's FLOPs that keeps the slack, and both effects scale with it:

```
 boundary   incep FLOPs at cpt 1 -> a53      mobilenet at cpt 1 -> a7/a9/bv
   0.5 MB          88.5%          -7.3%           79.7%          +5.6%
   1.0 MB          85.9%          -7.0%           53.5%          +3.7%
   1.5 MB          76.9%          -6.3%           21.5%          +1.5%
   2.0 MB          62.0%          -5.1%           18.9%          +1.3%   <- this run
   4.0 MB          41.5%          -3.4%            0.3%           0.0%
```

Linear from the two measured endpoints; it reproduces this run to within half a point on both models (−5.1 vs −4.6, +1.3 vs +0.3..1.2).

MobileNet falls off a cliff between 1.5 and 1.0 MB while InceptionV3 barely moves, so **1.5 MB is the knee** — pushed as `72e3af4e87`. Predicted a53 −6.3% for +1.5% on the other three.

On beaglev inceptionv3 +7.3%: main (all cpt 4) and the previous push (all cpt 1) were both ~2.28 s there, and only the mixture is 2.44 s — that is off-model, so I would want it re-sampled before reading anything into it. A re-run would settle both that and the knee. Sorry for the wrong claim first time round.

czoli1976:
> i'm having claude doing the sweep on the a53.

Same, now new commit,  1.5 MB is the knee apparentely, give it another ride on the bull

czoli1976:
@kali  We are playing with a knob here. Could tract figure it out by itself ? Maybe Yes, but it needs one physical number it doesn't currently measure: effective streaming bandwidth. tract already probes cache geometry (cache::detect); it has no notion of cache speed, so it can count bytes but can't price them. A ~1 ms probe (stream a buffer past the LLC, time it) sits naturally next to cache_info() and turns both magic constants into arithmetic. The benefit side needs no constant at all — it's observable: the dispatch can time its own chunks and see whether stragglers actually happened.

So perhaps extending the Cache Probing even further to the understand its speed will provide a better solution.

WDYT?

kali:
i think there is code somewhere in the cli that does a bw probe, actually...

czoli1976:
> i think there is code somewhere in the cli that does a bw probe, actually...

linalg ?

czoli1976:
@kali you were right about the probe — `tract_linalg::hwbench::bandwidth` is already in linalg, not just the cli. Following it I found `LinearCostModel` next door, with fitted per-part tables (`cortex_a53_linear.rs`, cpu 0xd03) and already a `restream` coefficient for packed-A re-stream volume — the exact physics this gate keeps trying to guess. Two things from the a53 table: its 12x8_a53 fit is 2.82e-10 s/MAC, so InceptionV3's 5.7 GMAC over 4 threads is ~0.40 s of compute against 1.23 s measured — two thirds of that model on that part is memory, which is why the chunk count matters there and nowhere else. And a53's `restream` is `0e0` because, as the doc says, it is fitted single-threaded with the weight resident; threaded chunking is precisely the regime that term would be non-zero in, and `cost-model gather` never varies the chunk grid.

So the cost-model route is real and it is mostly wiring, not new machinery. But it still ends in a fitted constant, so I tried the option that ends in none — pushed as `475f3cdb9b`.

**Don't decide the count at all.** An extra chunk only buys anything if a thread actually went idle, and rayon knows that where we don't: `Splitter::try_split` is thief-driven, spending a split budget and re-arming it only on a genuine steal. Handing it `(0..total).into_par_iter()` over a pre-computed grid threw that away. Now `split_panel_rect` halves the longer output extent — same shape objective `chunk_grid` got from `sqrt(chunks·m/n)` — and rayon decides the depth. Balanced work settles at ~one rectangle per thread, the minimum-traffic choice; imbalance splits further exactly where a thread stalled, so the re-read is paid only where the slack is used.

`CHUNKS_PER_THREAD`, both `CHUNK_SLACK_*` constants, the cache probe and the footprint arithmetic are gone; net −87 lines. `TRACT_MMM_CHUNKS_PER_THREAD` survives as "force the old fixed grid at this count" so you can A/B on the fleet without a rebuild — `=4` is the pre-PR dispatch, `=1` is what the machine gate forced.

Fair warning: this is the first version that is **not** inert on a large-cache part — it changes the split everywhere, so m1-max/i9/Orin can move. Best I can say locally, on an M1 Pro at 8 threads: inceptionv3 55.0 ms adaptive vs 55.9 forced-4 vs 56.4 forced-1, mobilenet 30.2 / 29.9 / 29.8 — all inside run-to-run spread, this machine's cache making the traffic free either way. Tests green, parallel-matmul bit-exact at 2/3/8 threads on all three paths.

Worth a bench when you have a slot. If the a53 still over-splits after this, the cost-model route above is the follow-up.

czoli1976:
A third version as well with a different approach if you fell like trying it

czoli1976:
Negative result, reverted in `10dd338570`. The adaptive split made it worse — a53 inceptionv3 **+9.3%** (1.23 → 1.35 s), the opposite direction to the −8.2% this PR is chasing, plus +11.8% on i9 and +8.7% on orangepi-rv2. That is the "not inert on big-cache parts" risk landing exactly as warned.

I instrumented the leaf count to find out why, on a 64x64 panel grid at 4 threads:

```
fixed cpt=4 grid   16 chunks
adaptive           42-140 leaves, typically 70-140
```

Four to nine times the chunks of cpt=4, up to 35× cpt=1. `Splitter::try_split` spends a budget of `current_num_threads()` splits, and on a *genuine steal* re-arms it to `max(nth, splits/2)`. So contention feeds more splitting, more splitting feeds more re-reads, and the re-reads lengthen the straggler that caused the steal in the first place. The feedback loop runs the wrong way on a memory-bound GEMM. Demand-driven splitting is only the right shape of answer if the splitter knows a split costs memory traffic, and rayon's does not.

So the branch is back to `72e3af4e87` — the 1.5 MB knee — which I should flag **has never actually been benched**: I pushed it and then superseded it with the adaptive commit before a run picked it up. On the linear fit that reproduced the 2 MB run to within half a point on both models, it predicts a53 −6.3% for +1.5% on a7/a9/beaglev, against the −4.6%/+1% that 2 MB actually measured.

Could you run one more bench on it? If the knee lands where the fit says, that is the best net this PR has had: −6.3% on the a53 with nothing else above the flag threshold. If it does not, the fit is wrong and the cost-model route is the one worth the effort, not another constant.

czoli1976:
image

czoli1976:
So I think the version 2 sounds good enought, third attempt went complitely wrong

kali:
We should consider the beaglev performance cautiously, it's very volatile. I'm working on it, but it actually means printing a stand with support for a fan, nothing off the shelf really fits. Making progress, but not there yet :)

kali:
Are we happy with this ? We saw some -8% yesterday on the same configuration, but this one comes with no counterpart...

czoli1976:
as good as it gets now

PR #2581, opened 2026-08-03
gelu: fuse the erf-form chain into a single op

czoli1976:
ONNX `Gelu` without `approximate="tanh"`, and `BiasGelu`, expand to five unfused nodes — `Mul -> Erf -> Add -> Mul -> Mul`. The tensor makes five round trips through memory and four intermediates are allocated, for what is one element-wise function. This adds a `GeluExact` op and a declutter rule on `Erf` that collapses the chain onto it, mirroring `detect_gelu_approx`.

### Shape of it

The rule anchors on the `Erf` because that is the one node in the chain with no ambiguity, then walks out to the `Mul` by `1/sqrt(2)` feeding it, the `Add` of 1, the `Mul` by 0.5 sharing the same source outlet, and the final `Mul`.

The op evaluates the same operations in the same order over a bounded scratch, so it is bit-identical to the chain rather than a re-derivation — including f16, where each step still rounds to f16 exactly as the separate binary ops did. `fuses_the_chain_and_keeps_the_values` builds the chain, checks a `GeluExact` appears and the `Erf` is gone, and compares the output bytes against the undecluttered model, for f32 and f16.

### Numbers

Chain through an optimized plan, criterion against a saved baseline with the rule disabled:

| dtype | shape | before | after | |
|---|---|---|---|---|
| f32 | 128x3072 | 468 us | 321 us | -31.6% |
| f32 | 512x3072 | 1.912 ms | 1.292 ms | -32.3% |
| f16 | 128x3072 | 605 us | 544 us | -7.7% |
| f16 | 512x3072 | 2.521 ms | 2.175 ms | -13.7% |

p = 0.00 throughout. f16 gains less because the fused op still widens to f32 around the erf kernel; #2579 replaces that with a table lookup and the two should compound.

### Note

This is the payoff I was looking for after measuring that a hand-written NEON `erf` kernel is **not** worth writing: the generic `SErf4` is branch-free arithmetic with no libm call, so LLVM already auto-vectorises it to within 8% of a hand-written NEON kernel of comparable op count (594 ns vs 551 ns per 1024 elements, against 1739 ns for a genuinely scalar loop). The cost on this path was never the erf itself — it was the five memory round trips.

### Tests

tract-core 271, test-f16 2378, test-unit-core 816, tract-onnx 14 — green. `cargo fmt --all` clean; `cargo clippy` clean (the two remaining warnings are pre-existing on main, in `softmax/mod.rs` and `optim/propagate_roi.rs`).

🍍

### DIFF
--- a/core/src/ops/math/mod.rs
+++ b/core/src/ops/math/mod.rs
@@ -702,8 +702,13 @@ element_wise!(erf, Erf,
      xs.iter_mut().zip(f32s.into_iter()).for_each(|(x, f)| *x = f16::from_f32(f));
      Ok(())
 };
- cost: |dt| {tvec!((Cost::FMA(dt), 11), (Cost::Div(dt), 1))}
+ cost: |dt| {tvec!((Cost::FMA(dt), 11), (Cost::Div(dt), 1))};
+ declutter: declutter_erf
 );
+
+fn declutter_erf(model: &TypedModel, node: &TypedNode) -> TractResult<Option<TypedModelPatch>> {
+    crate::ops::nn::gelu_exact::detect_gelu_exact(model, node)
+}
 
 element_wise!(acosh, Acosh, [f16, f32, f64] => |_, xs| {
     xs.iter_mut().for_each(|x| *x = x.acosh());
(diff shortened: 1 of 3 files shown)


kali (approved):

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2582, opened 2026-08-03
gelu: saturate the fused f32 kernel's tanh to exactly -1

### LINKED ISSUE #2569: arm64: fused f32 GELU returns 0.5*x*2^-23 instead of 0 for large negative x
`arm64simd_gelu_f32_4n_fused` (and `arm64simd_gelu_f32_4n`, which aliases it) does not decay to 0 for large negative inputs. The absolute error grows linearly with |x| and is unbounded.

Measured on an M-series host, against the generic `SGelu4`:

| x | fused | generic / libm |
|---|---|---|
| -3 | -3.6375225e-3 | -3.637433e-3 |
| -9 | -5.364418e-7 | -0 |
| -10 | -5.9604645e-7 | -0 |
| -100 | -5.9604645e-6 | -0 |
| -1000 | -5.9604645e-5 | -0 |
| -65504 | -3.9043427e-3 | -0 |

Repro:

```rust
use tract_linalg::arm64::arm64simd_gelu_f32_4n_fused;
use tract_linalg::generic::SGelu4;
use tract_linalg::element_wise::ElementWiseKer;

for x in [-9.0f32, -100.0, -1000.0, -65504.0] {
    let mut fused = vec![x; 16];
    arm64simd_gelu_f32_4n_fused::ew().run(&mut fused).unwrap();
    let mut generic = vec![x; 16];
    SGelu4::ew().run(&mut generic).unwrap();
    println!("x={x:<10} fused={:e} generic={:e}", fused[0], generic[0]);
}
```

### Cause

`linalg/src/arm64/arm64simd/gelu_fused.rs` clamps the pre-tanh argument to `[-8.9, 8.9]` (`COEFFS[0..2]`, applied by the `fmax`/`fmin` block around line 104), then evaluates the Padé polynomial. At the clamp bound the polynomial yields `-(1 - 2^-23)` — the largest-magnitude f32 below 1, one ULP short of `-1`.

One ULP of error in `tanh` is fine on its own. GELU then computes `0.5 * x * (1 + tanh)`, and that sum is a catastrophic cancellation: instead of `1 + (-1) = 0` it gives `2^-23 = 1.1920929e-7`, so the result is `0.5 * x * 2^-23` rather than 0. Every value in the table above is exactly that. libm's `tanhf` returns exactly `-1` past about 9.01, which is why the generic kernel is unaffected.

### How to fix

The lanes that need special-casing are exactly those the low clamp touched, so the pre-clamp comparison is already available as an equality against the clamp constant. After the final `0.5 * x * (1 + tanh)` combine, zero those lanes — two extra instructions per vector, no change to the polynomial:

```asm
    // v16 = clamped pre-tanh arg, v5 = dup(-8.9), vOUT = 0.5*x*(1+tanh)
    fcmeq vM.4s, v16.4s, v5.4s     // lanes that saturated at the low bound
    bic   vOUT.4s, vOUT.4s, vM.4s  // vOUT AND NOT mask -> 0 in those lanes
```

applied to each of the four accumulators in `loop4` and to the single accumulator in the `loop1` tail. That makes the kernel agree with `SGelu4` and libm at saturation.

The high side needs no change: there `1 + tanh` gives `2 - 2^-23`, so `gelu(x)` comes out as `x * (1 - 2^-24)` — a relative error of one half-ULP, not a cancellation.

### How it turned up

Building an `ew_impl_f16_via_f32!` f16 GELU on top of this kernel (as `silu_f16`/`sigmoid_f16` do) and comparing it against `HGelu8` over all 65536 f16 bit patterns: 14419 differed, worst case 7167 ULP, every one of them on the negative side with inputs from -3.59 down to -65504. Context in #2568, which took a different route and so is not affected.

🍍

czoli1976:
Fixes #2569. `arm64simd_gelu_f32_4n_fused` clamps the pre-tanh argument to `[-8.9, 8.9]`, but at the low bound the Padé polynomial evaluates to `-(1 - 2^-23)` rather than `-1`. `1 + tanh` therefore never cancels, and `gelu(x)` comes out as `0.5 * x * 2^-23` instead of 0 — an error that grows linearly with |x| and is unbounded.

Measured before: `-5.36e-7` at x=-9, `-5.96e-5` at x=-1000, `-3.90e-3` at x=-65504, against `-0` from the scalar kernel and from libm.

### The fix

The lanes that need correcting are exactly the ones the low clamp pinned, so `fcmeq` against the clamp constant identifies them and `bsl` substitutes an exact `-1.0` before the final combine. `v12`-`v15` were unused, so the masks need no spill. Nine instructions in the unrolled body, three in the tail; nothing else in the polynomial changes.

`decays_to_zero_on_large_negatives` covers both the unrolled body and the tail at six magnitudes.

### Cost

This is not free, and the microbenchmark number looks worse than the real one:

| | before | after | |
|---|---|---|---|
| kernel, 1024 elems | 630 ns | 682 ns | +8.2% |
| f32 FFN block (matmul -> gelu -> matmul) | 2.244 ms | 2.263 ms | +0.85% |

Both interleaved within one session rather than compared across runs — on this box a cross-run criterion baseline drifts far enough to invert the sign of a result this size, so I re-ran the two arms alternately and both rounds agreed. Sub-1% on a block where GEMMs dominate seemed a fair price for removing an unbounded error, but if you would rather not pay even that in the hot path I am happy to gate it.

### Two things I did not do

- **Moving the clamp instead.** Zero added instructions, and `-9.30` does land on exactly `-1` — but the neighbourhood is noisy under fma rounding (`-9.28` and `-9.32` do not), so it would be luck rather than a property.
- **Matching the sign of zero.** Corrected lanes return `+0` where the scalar kernel returns `-0`, because the combine is `fma(0.5x, tanh, 0.5x)` and `a - a` is `+0`, whereas the reference multiplies by zero. Making that exact costs another nine instructions per iteration. The two compare equal and behave identically downstream, so I left it — say the word if you want it exact. Note the fused kernel was never bit-identical to the scalar one anyway: the `fmla` rounds once where the reference rounds twice.

### Tests

tract-linalg 4415 — green. `cargo fmt --all` and `cargo clippy` clean.

🍍

### DIFF
--- a/linalg/src/arm64/arm64simd/gelu_fused.rs
+++ b/linalg/src/arm64/arm64simd/gelu_fused.rs
@@ -55,6 +55,7 @@ ew_impl_wrap!(
             //   v6:    tanh clamp high (8.9, dup v0.s[1])
             //   v7:    0.5 (broadcast of v3.s[1])
             //   v8-v11: 0.5 * original x (saved after load)
+            //   v12-v15: low-clamp masks, then the corrected tanh
             //   v16-v19: working (load -> pre_tanh -> clamped -> numerator)
             //   v20-v23: x² for tanh polynomial
             //   v24-v27: polynomial intermediates (denominator at end)
@@ -110,6 +111,13 @@ ew_impl_wrap!(
                     fmin v17.4s, v17.4s, v6.4s
                     fmin v18.4s, v18.4s, v6.4s
                     fmin v19.4s, v19.4s, v6.4s
+
+                    // Lanes pinned to the low clamp: tanh is -1 there to f32
+                    // precision, and the polynomial lands one ulp short.
+                    fcmeq v12.4s, v16.4s, v5.4s
+                    fcmeq v13.4s, v17.4s, v5.4s
+                    fcmeq v14.4s, v18.4s, v5.4s
+                    fcmeq v15.4s, v19.4s, v5.4s
 
                     // Tanh Padé polynomial (cloned from arm64simd_tanh_f32_4n.S.j2)
                     fmul v20.4s, v16.4s, v16.4s
@@ -209,11 +217,17 @@ ew_impl_wrap!(
                     fdiv v18.4s, v18.4s, v26.4s
                     fdiv v19.4s, v19.4s, v27.4s
 
+                    fmov v20.4s, #-1.0
+                    bsl v12.16b, v20.16b, v16.16b
+                    bsl v13.16b, v20.16b, v17.16b
+                    bsl v14.16b, v20.16b, v18.16b
+                    bsl v15.16b, v20.16b, v19.16b
+
                     // result = 0.5*x * (1 + tanh) = (0.5*x) + (0.5*x) * tanh
-                    fmla v8.4s, v8.4s, v16.4s
-                    fmla v9.4s, v9.4s, v17.4s
-                    fmla v10.4s, v10.4s, v18.4s
-                    fmla v11.4s, v11.4s, v19.4s
+                    fmla v8.4s, v8.4s, v12.4s
+                    fmla v9.4s, v9.4s, v13.4s
+                    fmla v10.4s, v10.4s, v14.4s
+                    fmla v11.4s, v11.4s, v15.4s
 
                     st1 {{ v8.4s, v9.4s, v10.4s, v11.4s }}, [{ptr}], #64
                     sub {len}, {len}, #16
@@ -235,6 +249,7 @@ ew_impl_wrap!(
 
                     fmax v16.4s, v16.4s, v5.4s
                     fmin v16.4s, v16.4s, v6.4s
+                    fcmeq v12.4s, v16.4s, v5.4s
 
                     fmul v20.4s, v16.4s, v16.4s
 
@@ -261,7 +276,10 @@ ew_impl_wrap!(
 
                     fdiv v16.4s, v16.4s, v24.4s
 
-                    fmla v8.4s, v8.4s, v16.4s
+                    fmov v20.4s, #-1.0
+                    bsl v12.16b, v20.16b, v16.16b
+
+                    fmla v8.4s, v8.4s, v12.4s
 
                     st1 {{ v8.4s }}, [{ptr}], #16
                     subs {len}, {len}, 4
@@ -275,6 +293,7 @@ ew_impl_wrap!(
             out("v0") _, out("v1") _, out("v2") _, out("v3") _,
             out("v4") _, out("v5") _, out("v6") _, out("v7") _,
             out("v8") _, out("v9") _, out("v10") _, out("v11") _,
+            out("v12") _, out("v13") _, out("v14") _, out("v15") _,
             out("v16") _, out("v17") _, out("v18") _, out("v19") _,
             out("v20") _, out("v21") _, out("v22") _, out("v23") _,
             out("v24") _, out("v25") _, out("v26") _, out("v27") _,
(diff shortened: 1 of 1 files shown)


kali:
@czoli1976 @cverrier i think we have implemented double fixes for this one ?

cverrier:
> @czoli1976 @cverrier i think we have implemented double fixes for this one ?

Yes, this seems to address the same bug that #2633 fixed for all supported architectures. Commit 7399d23e108da1487abf32c65a3e82989ff96819 contains the changes for both GELU and Tanh on arm64.

@czoli1976 Sorry I missed the issue #2569 you opened! I initially did the same kind of fix for Sigmoid and SiLU in #2620 , then found Tanh / GELU equivalent bug right after.

czoli1976:
Any Perf Difference ?

Il giorno gio 20 ago 2026 alle ore 20:17 Clément Verrier <
***@***.***> ha scritto:

czoli1976:
Closing then!

cverrier:
Actually that might be more complex that it seems. I'll try to spend some time to dive deeper into this issue.

czoli1976:
@cverrier no apology needed — and thanks for #2620/#2633, that was the right shape of fix.

But I don't think this one is a duplicate. #2633 and #2569 are opposite-direction bugs that happen to live at the same instruction slot:

- **#2633 — overshoot.** The Padé quotient reaches ±(1 + 2^-23), so `0.5 * (1 + tanh)` leaves `(0, 1)`: sign flip below, magnitude past `x` above. `fmax -1.0` / `fmin 1.0` fixes it.
- **#2569 — undershoot.** At the *low input clamp* (-8.9) the polynomial lands one ulp **short** of -1, so `1 + tanh` is `2^-23` and never cancels: `gelu(x) = 0.5 * x * 2^-23`, an error growing linearly with `|x|`. A clamp to `[-1, 1]` is a no-op on a value that is already inside `[-1, 1]`.

Probed the registered kernel on `679c9ca6e`, which has `7399d23e` in it:

| x | main today | generic kernel / libm |
|---|---|---|
| -9 | -5.364e-7 | -0 |
| -1000 | -5.960e-5 | -0 |
| -65504 | -3.904e-3 | -0 |

Same numbers as the issue. `arm64simd_gelu_f32_4n_fused` in every case.

### The two compose, and #2633 makes this patch cheaper

Rebasing conflicts in `gelu_fused.rs` — same slot — but the resolution is just clamp first, then the `bsl`. My `fmov v20.4s, #-1.0` **disappears**, because #2633 already has -1.0 sitting in v20: 8 added instructions per 16 lanes instead of 9. Rebased, the whole negative tail decays to zero and `cargo test -p tract-linalg --lib` is 4439 pass / 0 fail, #2633's new range tests included.

### Perf, since I asked

Three-way interleaved criterion, 5 rounds, min stat, `gelu_f32/linalg-asm-fused` @1024 elems, aarch64-apple-darwin:

| variant | ns | vs unfixed |
|---|---|---|
| #2633's gelu clamp reverted | 642.5 | — |
| main (#2633) | 706.7 | +10.0% |
| main + this PR | 750.7 | +16.8% (+6.2% over main) |

Every round agreed in sign, and the gap is well outside the round-to-round spread. Worth noting the ordering: the clamp for the bounded error costs ~10%, and this patch for the unbounded one costs ~6% on top. E2E was +0.85% on an f32 FFN block when I first measured it; the instruction count barely moved, so that should still hold.

### Two things still unfixed after both

- The **high** clamp mirrors the undershoot — `gelu(1000) = 999.99994`, `gelu(8) = 7.9999995`. Bounded at 1 ulp relative, so I left it alone deliberately.
- `gelu(-5.1257324)` returns `+0` on main against `-1.527e-7` from libm. #2633 removed the sign flip, but the value is still off; that one is inside the fit region, so neither patch addresses it.

Happy either way on whether this is worth 6% — closing it is a legitimate call if you'd rather not pay that in the hot path, and I offered to gate it in the original description. I just didn't want it closed on the assumption the tail is already fixed, because it isn't.

🍍

czoli1976:
It beats me it gets slower but E2E should be invisible ...

cverrier:
@czoli1976 I'm currently working on a more general fix. I'll let you know once it's ready.

czoli1976:
Roger that!

Il giorno ven 21 ago 2026 alle ore 13:37 Clément Verrier <
***@***.***> ha scritto:

PR #2590, opened 2026-08-05
use f32 for f16 in npy

kali:
ndarray-npy cannot decode float16 arrays, so npz bundles holding f16 (and the writer stored f16 as f32). Switch the CLI npz reader and writer to npyz, which supports every numpy scalar dtype tract has a Datum for, so f16 bundles round trip natively.


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2592, opened 2026-08-06
core/cnn: SIMD-vectorize BlockedConv's inner loop for WASM

czoli1976:
`BlockedConv` reduced its 16-wide W blocks with scalar multiplies on every target; on wasm `+simd128` hosts this runs the `(kh, icg)` reduction as 4×f32x4 load/madd/store per output channel instead, with `f32x4_relaxed_madd` (FMA) when `+relaxed-simd` is active and the existing scalar path kept as the fallback when simd128 is off.

### Numbers

`df_convp.1` shape (group=2, 64→10ch, kernel [5,1], 100×96), release under wasmtime on x86-64:

| | scalar `run` | `run_simd` | speedup |
|---|---|---|---|
| `+simd128` | 4.15 ms | 3.10 ms | 1.34x |
| `+simd128,+relaxed-simd` | 4.65 ms | 3.13 ms | 1.48x |

The ratio is modest because LLVM already auto-vectorizes part of the scalar const-OCG loop; the handwritten version wins on schedule and, under relaxed-simd, on FMA.

### Numerics

On plain simd128 each output lane accumulates over `(kh, icg)` in the same order and with the same mul-then-add rounding as the scalar kernel, so results are bit-exact against `run`. The relaxed-simd FMA variant rounds once per madd; the test tolerance (1e-5 max abs) covers it.

### Coverage

The pre-existing `blocked_conv_matches_reference` cases all landed in the const-dispatched OCG set, so the non-const `run_simd_generic` path had no test; an ocg=7 case now exercises it (full block + remainder).

New `unsafe` here is the v128 load/store blocks; the file already carries `unsafe` (`get_unchecked` / `as_slice_unchecked`) in the scalar path.

Verified on native (scalar fallback), `wasm32-wasip1` `+simd128` and `+relaxed-simd` under wasmtime, and `wasm32-unknown-unknown` without simd; `cargo fmt` / `clippy` clean.

🍍

### DIFF
--- a/core/src/ops/cnn/conv/blocked.rs
+++ b/core/src/ops/cnn/conv/blocked.rs
@@ -24,6 +24,20 @@ use crate::internal::*;
 /// Width of the inner SIMD-vectorised block over the contiguous W axis.
 const WB: usize = 16;
 
+#[cfg(all(target_feature = "simd128", not(target_feature = "relaxed-simd")))]
+macro_rules! blocked_madd {
+    ($acc:expr, $a:expr, $b:expr) => {
+        f32x4_add($acc, f32x4_mul($a, $b))
+    };
+}
+
+#[cfg(target_feature = "relaxed-simd")]
+macro_rules! blocked_madd {
+    ($acc:expr, $a:expr, $b:expr) => {
+        f32x4_relaxed_madd($a, $b, $acc)
+    };
+}
+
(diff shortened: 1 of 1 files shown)


kali (approved):

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2594, opened 2026-08-06
linalg/wasm: make the relaxed-dot int8 path bit-exact on all engines

czoli1976:
Under +relaxed-simd, wasm_i32_4x4 feeds full-signed i8 B into i32x4.relaxed_dot_i8x16_i7x16_add's i7-range operand, which x86 lowerings (pmaddubsw/vpdpbusd) read as unsigned: the wasm::mmm_i32 i8i8 tests fail under wasmtime on x86-64, and would block #2565's relaxed-simd CI lane once it merges. This sign-splits B into two i7-safe planes with one relaxed dot each, folded back after the K-loop as `acc -= acc_hi << 7` — bit-exact on every engine, still dot-product speed, no packing change.

🍍

### DIFF
--- a/linalg/src/wasm/mmm_i32.rs
+++ b/linalg/src/wasm/mmm_i32.rs
@@ -2,13 +2,14 @@ use crate::Scaler;
 use crate::mmm::FusedKerSpec;
 use crate::mmm::ImplementationQuality;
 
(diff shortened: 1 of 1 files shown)


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali (approved):

PR #2596, opened 2026-08-06
linalg,core: zero the padding lanes the mmm kernels compute on

czoli1976:
In release builds the packed-operand allocators (pack_tensor/pack_tensor_view, im2col's PackedFormat buffers) and the border-tile scratch buffers leave their padding lanes uninitialized; the mmm kernels compute on those lanes before discarding them, and garbage bytes there frequently decode to denormal f32s, stalling partial tiles ~14x on x86 via microcode assists (with no FTZ escape hatch under WASM). Zeroing the partial last panel at pack time and the border scratch unconditionally fixes this — DeepFilterNet3's streaming decoders come out ~11% faster under wasmtime on x86-64 with bit-identical output — and also corrects AddRowColProducts' col-tail zeroing, which wrote to the rows buffer.

🍍

### DIFF
--- a/core/src/ops/cnn/conv/im2col.rs
+++ b/core/src/ops/cnn/conv/im2col.rs
@@ -157,8 +157,10 @@ impl EvalOp for Im2Col {
             let r = geometry.out_format.r();
             // Buffer geometry. zero_init for PackedI8K4: the K=4-inner writer skips
             // the K-padding lanes (k..k_aligned), which SMOPA accumulates — they must
-            // be 0. PackedFormat has no K padding; its mn-padding maps to discarded
-            // output rows, so uninitialized is fine (matches prior behaviour).
+            // be 0. PackedFormat has no K padding; its mn-padding lanes are computed
+            // on (then discarded) by the kernel, so the partial last panel still
+            // needs zeroing — garbage there decodes to denormals that stall the fp
+            // pipeline. Done after allocation below.
             let (single_panel_len, buf_align, zero_init) =
                 if let Some(pf) = geometry.out_format.downcast_ref::<PackedFormat>() {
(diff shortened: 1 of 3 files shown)


czoli1976:
Additional measurements on scope. The touched code is arch-independent (generic packers, shared tile driver, im2col), so every CPU backend runs through it; whether the zeroing shows up as speed depends on the hardware's denormal cost.

Native x86-64 (AVX/FMA kernels, same pulsed DeepFilterNet3 submodels, 100 frames/iter):

| model | main | this PR |
|---|---|---|
| enc | 43.2 ms | 41.8 ms (−3%) |
| erb_dec | 50.0 ms | 43.2 ms (−14%) |
| df_dec | 80.2 ms | 42.3 ms (−47%) |

Native x86 is hit harder than wasm: the wider native tiles carry more padding lanes per border tile, and neither tract nor Rust sets FTZ/DAZ (MXCSR is never touched), so every op on a garbage lane pays the microcode assist — on main, native df_dec actually ran slower than the wasm build of the same model (80 ms vs 49 ms). aarch64 should be roughly neutral on speed (modern cores handle denormals at full rate; armv7 NEON is architecturally flush-to-zero), but every platform gains determinism: previously the padding content of packed buffers — and therefore timing — depended on allocator history, which is exactly the kind of run-to-run noise CI bench thresholds hate.

Outputs are unaffected everywhere: padding lanes never feed valid results. Verified bit-identical enhanced audio through the full DeepFilterNet3 pipeline (STFT → three models → synthesis), main vs this PR.

czoli1976:
@kali really curious to see what this does to the production workloads and hey snips

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali (approved):

PR #2598, opened 2026-08-06
linalg: AVX (no FMA) kernel tier for pre-Haswell x86

czoli1976:
CPUs with AVX but without AVX2+FMA (Sandy Bridge / Ivy Bridge, AMD Bulldozer family) currently run the generic f32 and i32 kernels, since the x86_64 plug tree is gated on avx2 and all f32 mmm kernels need fma. This adds an AVX kernel tier: the full fma-tier f32 tile set as mul+add (with gather-free add_unicast and the same shape-fitted n dispatch), mul+add ports of sigmoid/tanh, a 128-bit VEX i32/i8 kernel with the full quantization epilogue, and the already-AVX-clean mul_by_scalar/max/min element-wise kernels.

Measured on an i7-3770 (Ivy Bridge) against the generic kernels each shape previously ran on:

| kernel | shape (m×k×n) | generic | this PR | speedup |
|---|---|---|---|---|
| avx_mmm_f32_16x6 | 256×256×256 | 18.4 GFLOP/s | 47.9 GFLOP/s | 2.6× |
| avx_mmm_f32_16x6 | 1024×1024×1024 | 18.3 GFLOP/s | 43.2 GFLOP/s | 2.4× |
| avx_mmm_f32_16x5 | 256×256×5 | 11.6 GFLOP/s | 40.3 GFLOP/s | 3.5× |
| avx_mmm_f32_24x4 | 256×256×4 | 18.4 GFLOP/s | 32.2 GFLOP/s | 1.7× |
| avx_mmm_f32_32x3 | 256×256×3 | 13.6 GFLOP/s | 31.6 GFLOP/s | 2.3× |
| avx_mmm_f32_40x2 | 256×256×2 | 9.0 GFLOP/s | 24.8 GFLOP/s | 2.7× |
| avx_mmm_f32_8x8 | 64×512×8 | 18.8 GFLOP/s | 36.4 GFLOP/s | 1.9× |
| avx_mmm_f32_64x1 (mmv) | 1024×1024×1 | 9.6 GFLOP/s | 17.0 GFLOP/s | 1.8× |
| avx_mmm_i32_8x4 (i32×i32) | 256×256×256 | 3.9 GMAC/s | 14.5 GMAC/s | 3.7× |
| avx_mmm_i32_8x4 (i8×i8) | 256×256×256 | 3.4 GMAC/s | 7.4 GMAC/s | 2.2× |
| avx_sigmoid_f32 | 1M elements | 0.75 Gelem/s | 1.42 Gelem/s | 1.9× |
| avx_tanh_f32 | 1M elements | 0.83 Gelem/s | 1.39 Gelem/s | 1.7× |

🍍

### DIFF
--- a/linalg/src/x86_64_fma.rs
+++ b/linalg/src/x86_64_fma.rs
@@ -61,6 +61,7 @@ pub mod panel_extract;
 pub mod rms_norm;
 pub mod softmax;
 
+const AVX: fn() -> bool = || is_x86_feature_detected!("avx");
 const AVX2: fn() -> bool = || is_x86_feature_detected!("avx2");
 const FMA: fn() -> bool = || is_x86_feature_detected!("fma");
 const AVX512F: fn() -> bool = || is_x86_feature_detected!("avx512f");
@@ -77,6 +78,17 @@ tanh_impl!(f32, avx512_tanh_f32, 16, 16, is_x86_feature_detected!("avx512f"));
 sigmoid_impl!(f32, avx512_sigmoid_f32, 16, 16, is_x86_feature_detected!("avx512f"));
 
 fn plug_avx2(_ops: &mut Ops) {}
+
+/// Element-wise kernels for AVX-capable CPUs outside the fma tier: the
+/// mul_by_scalar / max / min asm is plain AVX. sigmoid, tanh and softmax use
+/// fma asm and keep their generic fallback on this tier.
+fn plug_avx(ops: &mut Ops) {
+    ops.mul_by_scalar_f32 = Box::new(|| by_scalar::x86_64_avx_f32_mul_by_scalar_32n::ew());
+    ops.max_f32 = Box::new(|| max::x86_64_fma_max_f32_32n::red());
+    ops.min_f32 = Box::new(|| min::x86_64_fma_min_f32_32n::red());
+
+    log::info!("mul_by_scalar_f32, max_f32, min_f32: x86_64/avx activated");
+}
 
 fn plug_fma(ops: &mut Ops) {
     panel_extract::plug(ops);
@@ -146,6 +158,11 @@ fn plug_avx512f(ops: &mut Ops) {
 
 pub fn plug(ops: &mut Ops) {
     mmm::plug(ops);
+    if is_x86_feature_detected!("avx")
+        && !(is_x86_feature_detected!("avx2") && is_x86_feature_detected!("fma"))
+    {
+        plug_avx(ops);
+    }
     if is_x86_feature_detected!("avx2") {
         plug_avx2(ops);
         if is_x86_feature_detected!("fma") {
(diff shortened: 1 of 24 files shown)


czoli1976:
@kali been doing last PRs on a Linux Intel® Core™ i7-3770  and noticed this gap .... will not move your bench :-)

 16x5/24x4/32x3/40x2 and  AVX sigmoid/tanh to follow

czoli1976:
End-to-end numbers to go with the kernel microbenchmarks above, on the two models I had to hand.

Setup: `a469e802d` (merge-base) vs `857a90c05` (this branch), both built with identical flags
(`CARGO_PROFILE_RELEASE_LTO=false` on both sides — the asm kernels are unaffected by it, and full LTO
does not fit in this box's RAM). i7-3770, Ivy Bridge, AVX without FMA. Runs interleaved
base/branch/base/branch to absorb thermal drift, seeded fixed inputs rather than `-R`.

| model | baseline | this branch | |
|---|---|---|---|
| trunet, S=100 | ~94.6 ms/i | ~58.6 ms/i | **1.61×** |
| nemotron_decoder, BATCH=1 | ~2.15 ms/i | ~2.07 ms/i | **1.04×** |

Outputs are bit-exact against baseline on every output tensor of both models
(`max_abs_diff = 0.000e+00`, `np.array_equal` true): trunet's single output, and the decoder's
`outputs` / `out_states_0` / `out_states_1`.

### The decoder result is not a dispatch failure

Worth recording since the 1.04× looks like the kernels did not get picked up. They did — `perf` on the
branch build shows the decoder spending 84% in `avx_mmm_f32_40x2`, replacing the 80% it spent in
`tract_linalg::generic::mmm::kernel::<f32, 4, 1>` on baseline. The swap happens; the wall clock just
does not move.

That model is DRAM-bandwidth-bound. It carries 27.6 MB of weights, all streamed once per token:
27.6 MB / 2.07 ms ≈ 13.3 GB/s, which is most of what DDR3-1600 dual-channel delivers here. Making the
arithmetic ~1.8× faster cannot help when the arithmetic is not the constraint.

So the table above is the honest shape of the win: compute-bound models collect the full speedup
(trunet's conv stack, 1.61×), weight-streaming decoders do not. The `avx_mmm_f32_64x1` microbenchmark
at 1024×1024×1 is a cache-resident shape and overstates what a real decoder sees. Nothing wrong with
the kernel — it is just not the bottleneck in that regime.

### Method note

This box needs the first two samples after a build discarded. A cold-cache warm-up curve here spans
88 → 58 ms/i on the *same binary*, which is large enough to manufacture a convincing fake speedup if
you sample once. Everything above is from settled samples.

I also stacked #2596 on top of this branch (cherry-picks clean — the two touch disjoint files) and saw
no measurable difference on either model, under default glibc and with
`MALLOC_TRIM_THRESHOLD_`/`MALLOC_MMAP_THRESHOLD_` raised to force dirty-heap reuse. That does not
contradict the 11% measured there under wasmtime: native glibc returns freed pages to the OS and gets
zero-filled ones back, so the padding lanes read as zeros and the hazard is masked. It just means I
could not reproduce that one natively on these shapes.

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2599, opened 2026-08-07
riscv64: 1 of 4: RVV 1.0 f32 matmul kernels

czoli1976:
Adds a RISC-V backend for the ratified Vector extension. rv64gc already built,
but ran the generic Rust kernels for every matmul; this plugs an RVV 1.0 f32
mmm tier in behind runtime detection.

## What's here

- Detection: V from the `AT_HWCAP` bit, VLEN from the `vlenb` CSR. That bit is
  a sufficient RVV-1.0 gate on its own — Linux never sets it for the
  incompatible 0.7.1 draft on the Allwinner D1 and Sophgo SG2042.
- Four f32 kernels rendered from one template: 8x8 and 16x8 GEMM tiles, 32x1
  and 64x1 GEMV.
- `.travis/cross.sh` and CI matrix entries at VLEN 128 and 256.

## The part worth reviewing carefully

RVV is vector-length agnostic, so VLEN is a runtime property of the hart, while
`MR` and `NR` have to be const generics because they select the packing format.
Each kernel reconciles that by fixing `(MR, NR, LMUL)` and pinning `vl` to `MR`.
`vsetvli` clamps to `VLMAX`, so such a kernel is correct wherever
`VLMAX >= MR`, merely leaves lanes idle when `VLMAX > MR`, and computes a
**short tile** when `VLMAX < MR` — wrong, not failed.

So dispatch is gated on `vlmax_f32(LMUL) >= MR`, and each kernel re-checks the
granted `vl` on entry, turning a predicate/assembly disagreement into a clean
fallback rather than a quiet wrong answer.

That guard cannot fire on a hart with no vector unit at all, where the guard
instruction is itself illegal; there the predicate is the whole protection.

One consequence for reviewers: tract's generated kernel suites early-return as
passes on an unsupported kernel, so a green run says nothing about whether the
predicates are right — all four VLENs report identical counts either way.
`dispatch_matches_vlen` asserts the dispatch set directly instead.

## Assembly, not intrinsics

Rust exposes no stable RVV intrinsics and `-C target-feature=+v` is itself
unstable, so this follows the x86_64/arm64 pattern rather than the wasm one. An
assembler probe keeps toolchains predating RVV 1.0 on the generic fallback, as
the SME and dotprod probes already do.

## Testing

tract-linalg and tract-core pass under qemu-riscv64 at VLEN 128, 256, 512 and
1024, and with V absent. The two CI entries cover 128 and 256, which select
disjoint halves of the kernel set.

End to end, InceptionV3 (NNEF, 5.7 GFLOP) loads, optimises and executes on
riscv64 under emulation, returning a well-formed softmax over 1001 classes.
`tract kernels` lists the RVV kernels as registered and dispatchable.

No hardware numbers yet: QEMU proves correctness and nothing whatsoever about
speed. Happy to hold this until I can benchmark on a K1 board if you would
rather not take it on emulation alone — and equally happy to hear that a
RISC-V backend is not something you want to carry, before I spend more on it.

🍍

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -267,8 +267,9 @@ jobs:
           - triple: armv7-unknown-linux-gnueabihf-stretch
             rustc_triple: armv7-unknown-linux-gnueabihf
           # RISC-V boards run a fully-static musl CLI (built directly by cross.sh, no stretch
-          # container); one self-tuning build serves every rv64gc board. No RVV kernels yet, so
-          # this is the generic scalar path.
+          # container); one self-tuning build serves every rv64gc board. The RVV kernels are
+          # picked at runtime off the hart's VLEN, so this same binary runs them on an RVV 1.0
+          # board and the generic scalar path where the vector unit is absent or the 0.7.1 draft.
           - triple: riscv64gc-unknown-linux-musl
             rustc_triple: riscv64gc-unknown-linux-musl
     steps:
--- a/.github/workflows/cross-platform.yml
+++ b/.github/workflows/cross-platform.yml
@@ -91,6 +91,8 @@ jobs:
           - aarch64-unknown-linux-musl
           - cortexa53-unknown-linux-musl
           - armv7-unknown-linux-musl
+          - riscv64gc-unknown-linux-gnu
+          - rvv128-unknown-linux-gnu
           - aarch64-linux-android
           - armv7-linux-androideabi
           - i686-linux-android
--- a/.travis/cross.sh
+++ b/.travis/cross.sh
@@ -211,9 +211,12 @@ case "$PLATFORM" in
                 export ARCH=riscv64
                 export QEMU_ARCH=riscv64
                 export RUSTC_TRIPLE=riscv64gc-unknown-linux-musl
-                export CUSTOM_TC=`pwd`/riscv64-linux-musl-cross
-                [ -d "$CUSTOM_TC" ] || curl -s https://tract-test-assets.tract.rs/toolchains/riscv64-linux-musl-cross.tgz | tar zx
-                export TARGET_CC=$CUSTOM_TC/bin/riscv64-linux-musl-gcc
+                # Bootlin rather than musl.cc: the linalg RVV kernels need an assembler that
+                # knows the ratified RVV 1.0 encodings, which arrived in binutils 2.38, and
+                # musl-cross-make has cut no release since 2021 and is stuck on 2.37.
+                export CUSTOM_TC=`pwd`/riscv64-lp64d--musl--stable-2025.08-1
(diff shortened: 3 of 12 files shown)


kali:
Do you happen to know if there are RISC-V devices where tract would actually make sense ? Anything affordable that could run benchs ?

czoli1976:
> Do you happen to know if there are RISC-V devices where tract would actually make sense ? Anything affordable that could run benchs ?

Where tract actually fits on RISC-V
tract's niche is CPU-only inference of modest models with low latency — no GPU, no vendor NPU toolchain. On RISC-V that maps cleanly onto K1/X60-class boards, and specifically:

Audio and streaming — wake word, VAD, ASR. tract's pulse/streaming support has few equivalents, and it's Sonos's own domain.
Models the vendor NPU compiler won't take. A discrete NPU only runs what its AOT toolchain accepts, quantized to its op set. Anything else falls back to CPU — which is where tract lives.
Where tract makes less sense: vision on NPU-equipped parts. The K230's 6 TOPS KPU does MobileNet v2 at 670 fps; a CPU vector path won't compete.

What to buy
Board	                

- Orange Pi RV2 8GB	Ky X1 (= SpacemiT K1)	~$50 AliExpress, $30 for 2GB	Buy this. VLEN=256, has zvfh
- Banana Pi BPI-F3	SpacemiT K1	~$70-100	Same silicon, more I/O
- K3 Pico-ITX	SpacemiT K3 / X100	higher cost 	RVA23, future-proof, overkill for now

czoli1976:

The conflict was a duplicate rather than a code collision: this PR carried a portable `load_a_slice` fallback for hwbench, and fb4bf2fc4 has since landed the same fix upstream. That commit is dropped here, so this is now just the two RVV commits. Confirmed upstream's version covers riscv64 by building `tract-linalg --features hwbench` for the target, which is what the dropped commit existed to fix.

Re-verified on the new base under qemu-riscv64 11.0.3: 1959 tract-linalg tests on this branch, and the full stack at 2709 across VLEN 128, 256 and 512, on stock `rva23u64`, and with V absent. `cargo fmt --all` clean.

kali:
FYI, ordered a orange pi, but delay is not great. i also have a visionfive2 somewhere in a drawer that i could setup for the pre-simd variant, but i'm not 100% it's worth the trouble.

czoli1976:
Probably not, let you decide what to do with these:
- park them until you get a board
- take them in as EXP and use QEMU to check correctness only

Il giorno gio 13 ago 2026 alle ore 09:20 Mathieu Poumeyrol <
***@***.***> ha scritto:

czoli1976:
@kali when do you expect the orange pi so we can bring this PRs Bundle in ?

kali:
Let's do this :)

kali:
lol. old toolchain.

kali:
oO

kali (approved):

kali:
one down.

PR #2600, opened 2026-08-07
riscv64: 2 of 4: 16 RVV matmul kernels behind Zvfh

czoli1976:
> **Stacked on #2599.** GitHub cannot base a pull request on a fork branch, so this
> targets `main` and its diff carries that PR's commits too. The only new commit here
> is `80a800f5` — everything else belongs to #2599 and is reviewed there.

Stacked on the f32 tier. Adds four f16 mmm kernels — 16x8 and 32x8 GEMM, 64x1
and 128x1 GEMV — from the same template at SEW=16, where VLMAX doubles and so
does every tile height for a given LMUL and VLEN.

No template change was needed: the f32 PR already parameterised element size,
so this is a kernel table, a probe, and a predicate.

## Why this is not speculative

Zvfh is optional in RVA23, so it has to be detected rather than assumed — but
it is not hypothetical hardware. The SpacemiT X60 in the K1 reports `zvfh` and
`zvfhmin` in its `/proc/cpuinfo` isa string, which covers the Banana Pi BPI-F3,
Orange Pi RV2 and Milk-V Jupiter. llama.cpp's RISC-V build enables `zvfh`
alongside `v` for exactly this hardware.

The C920v2 in the SG2044 has scalar `zfh` only, so this is a VLEN=256-class
feature in practice, and the predicates fall back cleanly on parts without it.

Zvfhmin is *not* sufficient: it provides only f16<->f32 conversion and so cannot
carry an f16 accumulator, which is why the detection looks for Zvfh
specifically.

## Testing

No CI change needed — the `-cpu max` models the f32 PR added already enable
Zvfh, so both existing riscv64 entries exercise this tier. Locally, tract-linalg
passes at VLEN 128 and 256 with Zvfh on and off, and `dispatch_matches_vlen`
covers the f16 predicates alongside the f32 ones.

Same caveat as the f32 PR: correctness under emulation, no hardware numbers.

🍍

### DIFF
--- a/linalg/build.rs
+++ b/linalg/build.rs
@@ -83,6 +83,20 @@ fn assembler_supports_rvv() -> bool {
         .is_ok()
 }
 
+// Probe whether the target assembler can encode Zvfh (f16 vector arithmetic).
+// Zvfh reached binutils later than base RVV 1.0, so a toolchain can assemble
+// the f32 kernels and still reject these. When the probe fails we skip the f16
+// kernels and the `tract_rvv_zvfh` cfg, and f16 matmul stays generic.
+fn assembler_supports_zvfh() -> bool {
+    cc::Build::new()
+        .file("riscv64/rvv/dummy_rvv_zvfh.S")
+        .cargo_metadata(false)
+        .cargo_warnings(false)
+        .warnings(false)
+        .try_compile("tract_rvv_zvfh_probe")
+        .is_ok()
+}
+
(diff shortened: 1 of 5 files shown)


czoli1976:

This was downstream of #2599, whose conflict was a duplicate of fb4bf2fc4 (the hwbench fallback, since landed upstream). Dropping that one commit cleared this PR too — everything else cherry-picked onto current `main` with no conflicts. The only new commit here is now `80a800f5`.

kali (approved):

PR #2601, opened 2026-08-07
riscv64: 3 of 4: 8 and i32 RVV matmul kernels

czoli1976:
> **Stacked on #2599 and #2600.** GitHub cannot base a pull request on a fork branch,
> so this targets `main` and its diff carries those PRs' commits too. The only new
> commit here is `0b3bce0f` — the rest are reviewed in #2599 and #2600.

Stacked on the f32 and f16 tiers. Adds four i32-accumulator kernels — 8x8 and
16x8 GEMM, 16x1 and 32x1 GEMV — covering both packings the frame offers:
i32 x i32, and i8 x i8 accumulating into i32.

## The i8 inner loop

vtype sits at e16 for that loop, which buys two things at once. `vle8.v` takes
EEW=8 from the instruction rather than from vtype, so the packed A column loads
as bytes with no separate widening step beyond one `vsext.vf2`; and
`vwmacc.vx` widens on the way in, landing straight in the e32 accumulators.
Because VLMAX at e16/mN equals VLMAX at e32/m2N, a single `vl` of MR is valid
in both states, so the loop needs one `vsetvli` on entry and one on exit rather
than a pair per K step.

`Zvqdotq` would collapse this further, but it is unratified and no hardware
implements it yet.

## Quantised ops

`QScale`, `RoundingShiftRight` and `ShiftLeft` follow the sign-magnitude
reference in `generic/rounding.rs` rather than an arithmetic shift, so the six
rounding policies differ only in the nudge term. Two things make that cheaper
than it looks:

- The negative mask is taken before the magnitude overwrites the sign, and it
  serves the MinusInf and PlusInf nudges as well. PlusInf wants `v <= 0` rather
  than `v < 0`, but those differ only at `v == 0`, where `half + nudge` never
  survives the shift and the result is 0 under every policy.
- For the same reason `signum` needs no zero case.

`QScale` widens to e64 for the i64 product, rounds there, and truncates back,
matching the reference which computes in i64 and casts.

## Testing

`test_mmm_kernel_i32!` pulls in `mmm_q_scale_tests!`, so every policy is
proptested against the reference — that is what these were developed against.
2683 tract-linalg tests and tract-core's 270 pass at VLEN 128, 256, 512 and
1024, on stock RVA23, and with V absent, under both qemu 10.2.1 and 11.0.3.

No CI change needed; the entries the f32 PR added cover this tier.

Same caveat as the earlier two: correctness under emulation, no hardware
numbers.

🍍

### DIFF
--- a/linalg/build.rs
+++ b/linalg/build.rs
@@ -546,10 +546,20 @@ fn main() {
             }
         }
         "riscv64" if assembler_supports_rvv() => {
-            let mut files = render_rvv_kernels("f32", "4", "+v", RVV_F32_KERNELS, &suffix);
+            const F32: &str = "riscv64/rvv/rvv_mmm.S.j2";
+            let mut files = render_rvv_kernels(F32, "f32", "4", "+v", RVV_F32_KERNELS, &suffix);
+            files.extend(render_rvv_kernels(
+                "riscv64/rvv/rvv_mmm_i32.S.j2",
+                "i32",
+                "4",
+                "+v",
+                RVV_I32_KERNELS,
+                &suffix,
+            ));
             println!("cargo:rustc-cfg=tract_rvv");
             if assembler_supports_zvfh() {
                 files.extend(render_rvv_kernels(
+                    F32,
                     "f16",
                     "2",
                     "+v, +zvfh, +zfhmin",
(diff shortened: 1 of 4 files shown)


czoli1976:

This was downstream of #2599, whose conflict was a duplicate of fb4bf2fc4 (the hwbench fallback, since landed upstream). Dropping that one commit cleared this PR too — everything else cherry-picked onto current `main` with no conflicts. The only new commit here is now `0b3bce0f`.

PR #2611, opened 2026-08-10
core: give add_const a name no node is using

zavitax:
`add_const` deduplicates a constant by value, then adds the node under the caller's name without checking that name is free — two different questions, of which only the first was being asked. A constant that outlives the node it was named after then collides with a replacement arriving under the same name, and `check_compact` rejects the graph:

```
running pass codegen: after graph compaction: duplicate name /convt3/Conv.bias
```

Found loading plain (non-`ll`) DeepFilterNet 3 through `deep_filter`. In `erb_dec`, `/convt3/Conv` has an all-zero bias, so a const named `/convt3/Conv.bias` is created for it; being a zero tensor, the value-dedup hands that same node to the padding value of `/conv0_out/0/Conv`'s im2col; `/convt3/Conv` is later rewritten and its replacement bias arrives by patch under a name the shared node still holds.

Since `check_compact` is `#[cfg(debug_assertions)]`, the model loads in release and refuses to load in debug — invisible where it is harmless, fatal where it is being checked.

`add_const` now takes its name from `unique_name`. The two tests cover the name not being reused and an identical value still being shared, since the fix must not disturb that. `cargo test -p tract-core`: 272 passed, 0 failed, 3 ignored.

🍍

### DIFF
--- a/core/src/model/typed.rs
+++ b/core/src/model/typed.rs
@@ -147,7 +147,7 @@ impl SpecialOps<TypedFact, Box<dyn TypedOp>> for TypedModel {
             }
         }
         let fact = TypedFact::try_from(v.clone())?;
-        let name = name.into();
+        let name = self.unique_name(name.into()).into_owned();
         let op = if let Some(exotic) = &fact.exotic_fact {
             crate::ops::konst::Const::new_with_exotic_fact(v, exotic.clone())?
         } else {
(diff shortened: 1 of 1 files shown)


core/src/model/typed.rs
```diff
@@ -147,7 +147,19 @@ impl SpecialOps<TypedFact, Box<dyn TypedOp>> for TypedModel {
             }
         }
         let fact = TypedFact::try_from(v.clone())?;
-        let name = name.into();
+        // The loop above deduplicates by value; this deduplicates by name.
```
kali: chatty bot comment.

core/src/model/typed.rs
```diff
@@ -157,6 +169,17 @@ impl SpecialOps<TypedFact, Box<dyn TypedOp>> for TypedModel {
     }
 }
 
+/// `wanted`, or the first `wanted#1`, `wanted#2`, ... no node is using.
```
kali: this helper already exists (as self.unique_name)

kali (changes requested):
Good catch. But please reuse exitsting helpers when you can, and make sure your code assistant read CLAUDE.md everytime it works on tract.

kali (approved):

kali:
Thank you !

zavitax:
I wonder: how an initialization step contributes to speed regression?

kali:
It does not. This is clearly just noise, a false positive. I'm trying to eliminated them, but it takes time and bandwidth...

PR #2612, opened 2026-08-11
build: let the x86 avx-512 benches build on non-x86 hosts

cverrier:
`cargo check --benches -p tract-linalg` failed on aarch64 hosts, because six x86-only benches referenced x86-only kernels unconditionally: anyone developing on Apple Silicon hit six broken bench targets on a plain `--all-targets` check. They now build on every architecture and report being skipped when run off x86_64.

Nothing in CI caught this: the only job covering `--all-targets` runs on x86_64. The workspace clippy step now passes `--all-targets` too, so the scheduled macOS (aarch64) leg compiles the bench targets.

Work in #2613 will clear clippy warnings in test and bench code, guarded by CI.

### DIFF
--- a/.github/workflows/crates.yml
+++ b/.github/workflows/crates.yml
@@ -148,7 +148,7 @@ jobs:
       with:
         save-if: ${{ github.ref == 'refs/heads/main' }}
 
-    - run: rustup component add clippy && cargo clippy
+    - run: rustup component add clippy && cargo clippy --all-targets
     - name: clippy (bench-suite feature)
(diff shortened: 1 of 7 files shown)


kali (approved):

cverrier:
@kali Failing job: [Embedded targets / apple (aarch64-apple-darwin)](https://github.com/sonos/tract/actions/runs/31492495568/job/93869765484) ; see #2617 for a fix.

cverrier:
@kali I just rebased this branch to get the fix from #2617 , should be all good now.

PR #2621, opened 2026-08-12
onnx/gru: one matmul per operand instead of one per gate

czoli1976:
# onnx/gru: one matmul per operand instead of one per gate

GRU expansion issued six matmuls per cell — `Xt` against each of `Wz`/`Wr`/`Wh`
and `Ht-1` against each of `Rz`/`Rr`/`Rh` — after slicing the gate weights apart.
The gate weights are contiguous row blocks of `W` and `R`, so a single product per
operand already yields every gate's product as contiguous column blocks; slicing
the result instead of the weights gives the same values from two matmuls.

`Rh` stays separate when `linear_before_reset` is 0, because there it multiplies
`rt (.) Ht-1` rather than `Ht-1`; the fused `R` product covers `Rz`/`Rr` only in
that case, and all three otherwise.

At M=1 (streaming, one frame per call) these are GEMVs, so the win is fewer
dispatches and one contiguous weight stream rather than three — DeepFilterNet3
drops 3.7% end to end on x86_64.

Checked against onnxruntime on five GRU-bearing ONNX models (DeepFilterNet3's
three graphs, GTCRN, UL-UNAS, the last two including bidirectional cells): worst
relative difference 1.06e-06.

🍍

### DIFF
--- a/onnx/src/ops/rec/gru.rs
+++ b/onnx/src/ops/rec/gru.rs
@@ -60,20 +60,26 @@ impl WireBody for GRU {
 
         let h_size = body.outlet_fact(R)?.shape[1].clone();
 
-        wire!(Rz = array::Slice::new(0, 0.to_dim() * &h_size, 1.to_dim() * &h_size), R);
-        wire!(Rr = array::Slice::new(0, 1.to_dim() * &h_size, 2.to_dim() * &h_size), R);
-        wire!(Rh = array::Slice::new(0, 2.to_dim() * &h_size, 3.to_dim() * &h_size), R);
-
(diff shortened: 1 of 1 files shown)


czoli1976:
Closing this in favour of #2626.

The two are alternative restructurings of the same GRU cell and conflict with each other, and this one does not pay off where I measured it: merging the per-gate `[16,20]x[20,20]` matmuls into one `[16,20]x[20,60]` regressed a streaming speech-enhancement model (0.135 -> 0.138 ms/frame). #2626 fuses the whole standard cell into a single `GruEpilogue` instead, which cut matmul dispatches by 24% (46011 -> 34743 per 627 frames) on the same family and is a clear win at every size.

PR #2622, opened 2026-08-12
core/array: rewrite constant-index ScatterNd block assignments to Slice/Concat

czoli1976:
# core/array: rewrite constant-index ScatterNd block assignments to Slice/Concat

PyTorch exports a ring-buffer cache update — `cache[.., k..k + n, ..] = new` — as a
`ScatterND` whose indices are a constant enumerating one full coordinate per scalar.
`ScatterNd::eval` then walks every tuple, navigating the data view axis by axis and
assigning a 0-d subview, so a contiguous block copy runs as one dispatch per element.

When the indices are constant and are exactly the row-major enumeration of a
contiguous block along one axis, with every other axis fully covered, the whole op is
`Concat(axis, [Slice(data, 0..start), updates, Slice(data, end..dim)])`. The
comparison against that canonical enumeration is exact and elementwise, so anything
else — a reduction, duplicate or out-of-order tuples, a partial axis, symbolic shapes,
an `updates` shape that is not the block — declines and keeps the existing eval.

On GTCRN (streaming speech enhancement, 16 kHz) this fires on 6 of the model's 18
`ScatterNd` nodes, the 6 that carry the conv/attention caches, and the frame drops from
13.63 ms to 4.20 ms, single-threaded on an i7-3770. The replacement slices and concats
cost 0.165 ms against the 8.95 ms they displace. Outputs are unchanged: the rewrite
moves data without arithmetic.

The new `harness/nnef-test-cases/scatter-nd-block` case asserts both the values and
that the rewrite fired; its expected outputs come from onnxruntime, and its block sits
in the middle of the axis so that both the head and tail slices are exercised, which
the GTCRN caches alone do not do.

🍍


core/src/ops/array/scatter_nd.rs
```diff
@@ -112,12 +112,124 @@ impl ScatterNd {
...
+    /// covered, and `updates` matches that block in shape and datum type. Symbolic
+    /// shapes are declined: full coverage of an axis cannot be established against
+    /// constant indices. `TypedConcat::declutter` and `optim::slice` clean up behind
+    /// it.
+    fn declutter(
+        &self,
+        model: &TypedModel,
+        node: &TypedNode,
+    ) -> TractResult<Option<TypedModelPatch>> {
+        if self.reduction != ScatterReduction::None {
```
kali, line 178: we have rule_if!() for this

kali (approved):

czoli1976:
will try *rule_if!()* , thank you!

Il giorno lun 17 ago 2026 alle ore 17:21 Mathieu Poumeyrol <
***@***.***> ha scritto:

core/src/ops/array/scatter_nd.rs
czoli1976: Done in #2678: the three remaining `if cond { return Ok(None) }` guards in the block rewrite are now `rule_if!`, matching the `rule_if_some!` already used further down. Order preserved, so the rank check still precedes the `shape().last()` that depends on it.

PR #2626, opened 2026-08-12
core,onnx: fuse the standard ONNX GRU cell into one GruEpilogue op

czoli1976:
# core,onnx: fuse the standard ONNX GRU cell into one GruEpilogue op

The GRU importer wires a Scan body of roughly twenty dispatched nodes per iteration —
two matmuls, the gate slices, the bias adds, three activations and the output update.
For a cell scanned many times per frame that is almost entirely dispatch: on UL-UNAS,
28 GRUs of hidden 4 scanned across subbands run 524 body iterations per frame, and the
bodies are about 72% of the network while doing a few hundred flops each.

`GruEpilogue` takes `xh = Xt·Wᵀ + Wb` and `rh = Ht-1·Rᵀ + Rb`, both `[batch, 3*hidden]`
in gate order z, r, h, plus the previous state, and computes the new state in one pass.
The W-side and R-side biases fold into the two matmuls as single `3*hidden` slices of
`b`, which also leaves each matmul with one consumer so the adds are absorbed as
matmul micro-ops rather than dispatched.

The op is emitted directly by the importer, as `LstmEpilogue` is, and only where the
whole cell is expressible: `linear_before_reset != 0` — with the reset gate applied
before the recurrent product there is a matmul that depends on `rt`, and no single-pass
form exists — plus standard sigmoid/tanh activations, checked on the ops rather than
assumed, and a concrete hidden size. Everything else keeps the decomposed wiring.

Single-threaded on an i7-3770: UL-UNAS 5.71 ms to 3.24 ms, GTCRN 3.26 ms to 2.00 ms.
Checked against onnxruntime on DeepFilterNet3's three graphs, GTCRN and UL-UNAS, worst
relative difference 1.06e-06; and on a purpose-built GRU with each
`linear_before_reset` setting, confirming the fused path is taken for 1 and declined
for 0, both matching onnxruntime.

`harness/nnef-test-cases/gru-epilogue` covers the round trip and asserts the fusion
fired, since no ONNX node test in the suite sets `linear_before_reset`.

Two notes for review. The gate sums are associated differently from the decomposed
form, so results move by rounding and the harness case compares approximately. And the
op has no `axes_mapping`, so like `LstmEpilogue` it cannot be pulsified along a
non-scan axis; that is not a regression but it is a limit.

One question before I go further, since it decides whether more ops are welcome here.
Profiling says this gets part of the way but not to parity: onnxruntime runs one `GRU`
kernel per layer that loops every timestep and both directions internally — 28 kernel
calls on UL-UNAS against tract's 524 scan iterations — so a fused body still pays
per-iteration dispatch that a fused *sequence* does not. I have two follow-ups written
or sketched: folding the recurrent product into the cell (another ~25% on these models),
and beyond that a whole-sequence GRU op that would replace the Scan rather than fuse its
body, which is the only shape that reaches onnxruntime's dispatch count.

Would you rather have the incremental ops, or should the cell math here become the inner
loop of a single whole-sequence op? I would rather ask than land three GRU ops and find
you wanted one. The decomposed path stays regardless, since `linear_before_reset = 0`,
non-standard activations and symbolic hidden sizes all still need it.

🍍


kali (approved):

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2628, opened 2026-08-12
core,onnx: carry the recurrent product inside the fused GRU cell

czoli1976:
# core,onnx: carry the recurrent product inside the fused GRU cell

Stacked on #2626.

With the cell epilogue fused, what is left in the scan body is mostly the recurrent
product: a pack, a matmul and a reshape per iteration. On the subband GRUs these models
use, that product is a 4-by-12 matrix-vector multiply — 48 multiply-adds — and the
dispatched matmul costs about 1.2 us to do it, so the kernel call is two orders of
magnitude more expensive than the arithmetic.

`GruCell` takes the recurrent weights and bias directly and computes the product inline,
leaving the body with one op. It sizes its rows from the state and hands back the state's
own shape, so the scan state feeds it directly and the reshapes either side of it go too. That only pays while the product is small, so the importer
emits it under `MAX_INLINE_HIDDEN` and keeps `GruEpilogue` with a dispatched matmul above
it, where tract's packed kernels win. Same guards otherwise: `linear_before_reset != 0`,
standard sigmoid and tanh, concrete hidden size.

Single-threaded on an i7-3770, on top of #2626: UL-UNAS 3.24 ms to 2.44 ms, GTCRN 2.00 ms
to 1.60 ms. Across the whole series these two models have gone from 7.28 and 13.80 ms to
2.44 and 1.60.

Checked against onnxruntime on DeepFilterNet3's three graphs, GTCRN and UL-UNAS: worst
relative difference 1.06e-06. `harness/nnef-test-cases/gru-cell` covers the round trip
and asserts the cell fired with no matmul left in the body; the unit test checks it
against `GruEpilogue` fed the same product computed outside, so the inline arithmetic is
pinned to the path it replaces.

The question from #2626 still stands — if you would rather have one whole-sequence op
than these two, this is the cell arithmetic it would loop over, and I am happy to rework
it that way.

🍍


czoli1976:
Closing this one: measured on top of #2626 it is a sizeable regression.

FastEnhancer (streaming speech enhancement, GRU + frequency-axis attention), aarch64, min ms/frame at 16 kHz, with #2626 already applied:

| variant | #2626 | + this PR |
|---|---|---|
| base | 0.225 | 0.296 (**+31%**) |
| small | 0.406 | 0.592 (**+46%**) |

tiny, medium and large were unchanged, so it is specifically the mid sizes that lose. #2626 on its own is a clear win across the family and I would like to keep that; carrying the recurrent product inside the cell is what costs here.

Worth noting for anyone rebasing: this branch also carries #2626's commit, so cherry-picking it whole conflicts once #2626 is in — only the second commit is unique to it.

Happy to reopen if there is a workload where it wins; I only have this model family to go on.

PR #2630, opened 2026-08-12
data: let stack_tensors' block copy handle a non-trivial outer dimension

czoli1976:
# data: let stack_tensors' block copy handle a non-trivial outer dimension

`Tensor::stack_tensors` only took its `copy_nonoverlapping` path when every axis before
the concatenated one had extent 1; anything else fell through to
`assign_slice_from_resolved`. The inputs agree on the axes before the concatenated one, so
one outer loop carrying a stride covers those shapes too, and each contribution stays a
contiguous block.

DTLN's LSTM state is `[1, 2, 128, 2]` and its two stages stack it on the last axis, which
missed the old condition on the `2` and the `128`: 4.8 us to move 2 KB, about 0.4 GB/s.
`Concat` drops out of that model's top three ops, and end to end, single-threaded on an
i7-3770, median of 7 interleaved runs against the same build with only this file reverted:
DTLN stage 1 0.079 to 0.077 ms, stage 2 0.131 to 0.129, GTCRN 1.556 to 1.531, UL-UNAS
2.329 to 2.332. So a percent or two where it applies and nothing worse elsewhere — the op
is a small share of these models to begin with.

The condition also now checks the storage is plain before taking the fast path.
`plain_storage()` panics otherwise, and `dt.is_copy()` alone did not establish it.

🍍

### DIFF
--- a/data/src/tensor.rs
+++ b/data/src/tensor.rs
@@ -263,7 +263,7 @@ impl Tensor {
         self.as_plain_mut().context("Tensor storage is not plain")
     }
 
-    /// Create an uninitialized tensor (dt as type parameter).
+    /// Create an uninitialized tensor (dt as type paramater).
     #[inline]
     pub unsafe fn uninitialized<T: Datum>(shape: &[usize]) -> TractResult<Tensor> {
         unsafe { Self::uninitialized_dt(T::datum_type(), shape) }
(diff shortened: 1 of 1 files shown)


czoli1976:
Data point from another model family: this also fixes a trailing-axis `Concat` in the FastEnhancer speech-enhancement models.

The shape is two `[1,256,1,1]` tensors concatenated on axis 3 into `[1,256,1,2]` — the interleave half of a power-law magnitude compression. With `shape[..axis]` non-trivial it fell to the strided `assign_slice_from_resolved` path: **7 us to move 512 floats, about 0.29 GB/s**. Same signature as the DTLN case in the description. With this PR it drops to ~2 us.

kali (approved):

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2632, opened 2026-08-13
Fix dropping of uninitialised memory in tensor0

Ollie-Pearce:
`tensor0` builds an uninitialised buffer and then writes to that value with an assignment operator. [Assignment drops the uninitialised `tensor` before overwriting it with `x`.](https://doc.rust-lang.org/reference/expressions/operator-expr.html#r-expr.assign.drop-target)

```rust
let mut tensor = Tensor::uninitialized::<A>(&[]).unwrap();
tensor.as_slice_mut_unchecked::<A>()[0] = x;
```
### Test case:
```rust
#[test]
fn drop_of_uninit() {
    let _t: Tensor = tensor0(1.0f32);
}
```

### Verifying with Miri:
```
cargo +nightly-2025-08-20 miri test -p tract-data --test repro
```

Miri output:
```
error: Undefined Behavior: constructing invalid value at .pointer: encountered 0, but expected something greater or equal to 1
    |
512 |         self.ptr.cast().as_non_null_ptr()
    |         ^^^^^^^^ Undefined Behavior occurred here
    |
```

## Fix: 

Replace the assignment operator with the [`std::ptr::write` function which directly writes to `x`](https://doc.rust-lang.org/std/ptr/fn.write.html), fixing the drop of uninitialised memory.

### DIFF
--- a/core/src/ops/array/range.rs
+++ b/core/src/ops/array/range.rs
@@ -40,8 +40,9 @@ impl Range {
             let step = step.try_as_plain()?.to_scalar::<T>()?;
             {
                 let mut result_plain = result.try_as_plain_mut()?;
+                let slots = result_plain.as_slice_mut_unchecked::<T>().as_mut_ptr();
                 for i in 0..len {
-                    result_plain.as_slice_mut_unchecked::<T>()[i] = v.clone();
+                    std::ptr::write(slots.add(i), v.clone());
                     v = v + step;
                 }
             }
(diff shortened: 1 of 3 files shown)


kali:
Hey, nice catch. My imaginary friend mention another possible occurences in data/src/tensor.rs:1440 and core/src/ops/array/range.rs:51 . Want to have a look at these too ?

Ollie-Pearce:
Thanks, I'll take a look sometime over the next couple of days :)

Issue #2618, opened 2026-08-12
f32 Sigmoid returns negative values (-2^-24) for some large-negative inputs, e.g. sigmoid(-18.25)


hillin:
## Summary

Running an ONNX `Sigmoid` through tract returns **`-5.9604645e-8` (= −2⁻²⁴)** for certain
large-negative inputs. A sigmoid's range is (0, 1); a negative output is a range violation
that turns into NaN in graphs that feed sigmoid outputs into `Pow` with a fractional
exponent (`pow(negative, fractional) = NaN`).

Reproduces with a graph containing nothing but a single `Sigmoid` node, at every tensor
width including `[1]`, in `into_typed()`, `into_decluttered()` and `into_optimized()` alike.

Known trigger inputs (found by scanning `[-20, -16]` on a 0.0625 grid; there are likely
more): **`-18.25`** and **`-18.5625`**. Nearby values (−18.1875, −18.3125, −20, −17, …)
return correct small positives, so the undershoot is value-specific rounding in the
approximation, not a contiguous saturation region.

For comparison, onnxruntime 1.x CPU returns `0.0` for both trigger inputs — clamped, never
negative.

## Repro

Minimal model (Python, any onnx version):

```python
import onnx
from onnx import helper, TensorProto
inp = helper.make_tensor_value_info("x", TensorProto.FLOAT, [1])
out = helper.make_tensor_value_info("y", TensorProto.FLOAT, [1])
g = helper.make_graph([helper.make_node("Sigmoid", ["x"], ["y"])], "sig", [inp], [out])
onnx.save(helper.make_model(g, opset_imports=[helper.make_opsetid("", 17)]), "sigmoid.onnx")
```

Runner (`tract-onnx = "0.23.4"`):

```rust
use tract_onnx::prelude::*;

fn main() -> TractResult<()> {
    let plan = tract_onnx::onnx()
        .model_for_path("sigmoid.onnx")?
        .into_optimized()?
        .into_runnable()?;
    for x in [-18.25f32, -18.5625, -18.1875, -20.0, -17.0] {
        let out = plan.run(tvec![Tensor::from_shape(&[1], &[x])?.into()])?;
        let view = out[0].to_plain_array_view::<f32>()?;
        let y = view.as_slice().unwrap()[0];
        println!("sigmoid({x}) = {y:e}{}", if y < 0.0 { "   <-- NEGATIVE" } else { "" });
    }
    Ok(())
}
```

[...]

PR #2633, opened 2026-08-13
fix(linalg): fix tanh and gelu output range

cverrier:
## Summary

tract's f32 Tanh kernels could return values outside tanh's `(-1, 1)` range — up to `±(1 + 2^-22)` for `|x| ≈ 8.1–9.0` — and the fused arm64 GELU kernel, which clones the tanh polynomial, inherited it at both ends. This clamps every f32 Tanh kernel's output to `[-1, 1]`, and the internal tanh of the fused GELU kernel for the same reason. Follow-up to #2620, which fixed the same class of bug in Sigmoid/SiLU.

## Root cause

Near the input clamp (`±8.9`, `±9.0` on AVX-512) the correct answer is already within one f32 ulp of `1.0` — `1 - tanh(8.9) = 3.7e-8`, less than the `2^-24` step below 1 — and the rational fit's own few-ulp error is enough to round over the boundary. So the violations land strictly *inside* the input clamp, at ordinary activation magnitudes, and which values overshoot is rounding-specific rather than a contiguous region. Downstream this mirrors #2618: `sqrt(1 - tanh(x)²)` and `atanh(tanh(x))` become NaN.

## Fused GELU

`gelu(x) = x·Φ(x)` with `Φ = 0.5·(1 + tanh θ) ∈ (0, 1)`, so `gelu(x)` lies strictly between `0` and `x`. An out-of-range tanh breaks both bounds: `arm64simd_gelu_f32_4n_fused` returned `gelu(-5.1257324) = +3.05e-7` against a true `-9.6e-8` (sign flipped), and `gelu(5.0625) = 5.0625005`, further from zero than its own input. Clamping its internal tanh restores both.

## Why clamp the output rather than pull the input clamp in

A tighter input clamp would close the band, but it is a magic number per coefficient set *and* per rounding mode, undecidable for wasm (it depends on whether the host fuses `f32x4_relaxed_madd`), and it relocates the band instead of establishing the invariant. `[-1, 1]` is correct by construction for any coefficients, any host, and any kernel added later — and it can be tested as a property. `±1` is already an accepted output today, so no consumer sees anything new; accuracy inside the fit region is unchanged (bit-identical over `[0, 8]` on x86), at a cost of ~2 SIMD instructions per vector.

## Changes

Clamps in generic `stanh`, arm64 (Tanh + fused GELU), armv7neon, FMA, AVX-512 and the wasm relaxed kernel — one commit per architecture, shared tests last. f16 Tanh needs no clamp (real margin at its `±3.84` clamp; the new test sweeps every finite f16 through it), and the GELU kernels that *call* the Tanh kernel rather than inlining its polynomial are fixed transitively.

New frame tests: `tanh_range_on_tails`, plus GELU `sign_on_tails` and `magnitude_on_tails`. The shared invariant sweep grid goes from 1/256 to 1/4096, which is load-bearing: on the coarser grid the unfixed arm64 kernel passes outright. The existing accuracy tests cannot stand in — their `1e-4` tolerance is four orders of magnitude looser than the ~`2e-7` overshoot.

## Verification

Every kernel here was executed on its own hardware except wasm: arm64 (macOS), armv7 NEON (Cortex-A9), AVX-512 + FMA (i9-11900KB). Each new test was confirmed failing on the unfixed kernel at the predicted inputs and passing after, both loop bodies exercised, register liveness at each clamp site checked against the disassembly. An exhaustive x86 sweep of every f32 bit pattern with `|x| ≤ 12` gives `max(|y| - 1) = 0`. `cargo fmt --all` and `cargo clippy -p tract-linalg` clean.

`WasmTanh4Relaxed` was compiled and its emitted `f32x4.min`/`max` pair inspected against the already-fixed `WasmSigmoid4Relaxed`, but not executed — no relaxed-SIMD host available.

### DIFF
--- a/linalg/arm32/armv7neon/armv7neon_tanh_f32_4n.S.j2
+++ b/linalg/arm32/armv7neon/armv7neon_tanh_f32_4n.S.j2
@@ -128,6 +128,15 @@ armv7neon_tanh_f32_4n_{{suffix}}:
     vmul.f32    q11, q5, q8
     vmul.f32    q12, q6, q9
 
+    vdup.32     q13, d6[1]
+    vdup.32     q14, d6[0]
+    vmax.f32    q10, q13
+    vmax.f32    q11, q13
+    vmax.f32    q12, q13
+    vmin.f32    q10, q14
+    vmin.f32    q11, q14
+    vmin.f32    q12, q14
+
     vstmia      r0!, { q10, q11, q12 }
 
     subs        r1, #12
@@ -178,6 +187,11 @@ armv7neon_tanh_f32_4n_{{suffix}}:
 
     vmul.f32    q10, q4, q7
 
+    vdup.32     q13, d6[1]
+    vdup.32     q4, d6[0]
+    vmax.f32    q10, q13
+    vmin.f32    q10, q4
+
     vstmia      r0!, { q10 }
 
     subs        r1, #4;
@@ -204,6 +218,6 @@ armv7neon_tanh_f32_4n_{{suffix}}:
     .float 0.4641733162
 
     .float 1.0
-    .float 0                        // padding
+    .float -1.0                     // output clamp low
     .float 0                        // padding
     .float 0                        // padding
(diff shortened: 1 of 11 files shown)


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali:
we can ignore the a7, it's an artefact of my current work on stabilizing performance, i think.

kali (approved):

Issue #1353, opened 2024-03-11
`gather_nd` fails when using varaiable batch dims


alexander-camuto:
Can generate `gather_nd` ops from tensorflow as such: 

```python


import tf2onnx
import tensorflow as tf
from tensorflow.keras.layers import *
from tensorflow.keras.models import Model

# gather_nd in tf then export to onnx
x = in1 = Input((15, 18,))
w = in2 = Input((15, 1), dtype=tf.int32)
x = tf.gather_nd(x, w, batch_dims=1)
tm = Model((in1, in2), x )
tm.summary()
tm.compile(optimizer='adam', loss='mse')

spec = tf.TensorSpec([None, 15, 18], tf.float32, name='input')
index_spec = tf.TensorSpec( [None, 15, 1], tf.int32, name='index')

model_path = "network.onnx"

tf2onnx.convert.from_keras(tm, input_signature=[spec, index_spec], inputs_as_nchw=['input', 'index'], opset=13, output_path=model_path)

```

The resulting graph has variable batch sizes for the inputs:

<img width="505" alt="image" src="https://github.com/sonos/tract/assets/45801863/a570aece-1180-49bd-b8ad-e0fefe05d2b0">

Tract fails to analyze with: 

```
┏ 0 Source input
┃   ━━━ unk__58,15,18,F32
┣┓ 
┃┣ 9 Shape model/tf.compat.v1.gather_nd/BatchGatherND/Shape
┃┃   ━━━ 3,TDim unk__58, 15, 18
┃┣ 10 onnx.Cast model/tf.compat.v1.gather_nd/BatchGatherND/Shape__7
┃┃   ━━━ 3,I32
┃┣┻┻┻ 13 StridedSlice model/tf.compat.v1.gather_nd/BatchGatherND/strided_slice
┃┃   ━━━ 1,I32
┃┣┓ 
┃┃┣ 14 Split model/tf.compat.v1.gather_nd/BatchGatherND/unstack
┃┃┣┻ 15 Squeeze13 model/tf.compat.v1.gather_nd/BatchGatherND/unstack__11
┃┃┃   ━━━ I32
┃┃┣┻┻ 16 Range model/tf.compat.v1.gather_nd/BatchGatherND/Range
┃┃┃   ━━━ ?,I32
┃┃┣ 17 onnx.Cast model/tf.compat.v1.gather_nd/BatchGatherND/Cast
┃┃┣┓  
[...]

PR #2635, opened 2026-08-14
onnx: fix GatherNd output shape inference

czoli1976:
The `GatherNd` inference rule derived the trailing output dimensions from the indices rank and shape rather than the data tensor, so any ONNX model whose data tensor outranks its indices failed at the analyse stage. Core's `compute_shape` was already spec-correct — only HIR inference was wrong.

Found while loading an NVIDIA Nemotron 3.5 Lightning (`nemotron_h`) export, where MoE expert routing emits `GatherND(data: [8,128,256], indices: [32,1])` and tract rejected it with `Impossible to unify Val(128) with Val(32)`. This also covers #1353, whose `batch_dims=1` TF export fails through the same rule.

Indices addressing more axes than the data has previously spun forever (`data_rank - n` underflowed into a ~1.8e19 loop bound); it now reports an error.

Verified: new inference tests in `hir`, the ONNX backend suite (4528 pass), and NNEF round-trips with numeric assertions for `batch_dims` 0 and 1.

Note: I could not run `cargo clippy --workspace` on Linux — it fails compiling the Apple-only `core-graphics-types`. Clippy is clean on `tract-hir`, `tract-onnx`, `tract-core` and `tract-cli`.

🍍

### DIFF
--- a/hir/src/ops/array/gather_nd.rs
+++ b/hir/src/ops/array/gather_nd.rs
@@ -18,11 +18,23 @@ impl InferenceRulesOp for GatherNd {
             }
             s.given_2(
                 &inputs[1].shape[indices_rank - 1],
-                &inputs[1].rank,
-                move |s, n, input_rank| {
+                &inputs[0].rank,
+                move |s, n, data_rank| {
                     if let Ok(n) = n.to_i64() {
-                        for i in 0..(input_rank - n) as usize {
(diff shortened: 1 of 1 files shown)


kali (approved):

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2636, opened 2026-08-14
fix(core): cast the variable input when dropping a quantized neutral

cverrier:
## Summary

`declutter_neutral` removes a binary op whose uniform input is the op's neutral element. In the quantized case the op can't just vanish — it still re-encodes its input into the output's quantization parameters — so it degrades to a `Cast`. That `Cast` was wired to inlet 0 instead of to the variable input, which is the same thing except when the op is commutative and the uniform sits on the left: `Add(zero, x)` then requantized the constant and produced a tensor of zeros (or, when the two inputs had different shapes, aborted the optimization pass).

## Miscellaneous

The rest of the diff is cosmetic: `pos_checked` becomes `uniform_on_neutral_side`, and the two comments that were hardest to read are rewritten (the neutral-side rule as `x - 0 == x, but 0 - x == -x`, and the quantization note as a doc comment on the function).

## Other Issues — Not Addressed Yet

`SubF`'s neutral element is on its *left* (`SubF(a, b) = b - a`), so the side check is backwards for it — harmless today since `SubF` only exists post-codegen inside `OptBin*`; and keying the quantized branch on `var_dt != out_dt` rather than `is_quantized()` would also cover mixed-float neutrals like `Add(x: f16, 0f32)`, but that changes behavior and wants its own commit.

@kali What do you think about these issues?

### DIFF
--- a/core/src/ops/binary.rs
+++ b/core/src/ops/binary.rs
@@ -473,46 +473,42 @@ fn declutter_broadcasting_operand_1(
     Ok(None)
 }
 
+/// Shunt a binary op whose uniform input holds the op's neutral element
+/// (`x + 0`, `x * 1`, `x - 0`).
+///
+/// The neutral element is compared on dequantized values, so a quantized node
+/// can be arithmetically neutral while still re-encoding its input: `out_dt`
+/// may carry other quantization parameters than the variable input, and the
+/// same real value is then a different integer. Such a node degrades to a
+/// `Cast` instead of vanishing.
 fn declutter_neutral(
     model: &TypedModel,
     node: &TypedNode,
     mini_op: &dyn BinMiniOp,
     out_dt: DatumType,
 ) -> TractResult<Option<TypedModelPatch>> {
-    if let Some(uniform) = crate::ops::binary::one_input_is_uniform(model, node)? {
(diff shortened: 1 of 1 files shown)


core/src/ops/binary.rs
```diff
@@ -1287,4 +1283,34 @@ mod tests {
...
+        input.try_as_plain_mut()?.as_slice_mut::<u8>()?.copy_from_slice(&[10, 35, 60, 200]);
+        let input = tvec!(input.into_tvalue());
+
+        let before = model.clone().into_runnable()?.run(input.clone())?;
+        let decluttered = model.into_decluttered()?;
+        assert!(decluttered.nodes().iter().all(|n| n.op_as::<TypedBinOp>().is_none()));
+        let after = decluttered.into_runnable()?.run(input)?;
+        assert_eq!(&*before[0], &*after[0]);
+        Ok(())
+    }
```
cverrier, line 1315: The new test builds `Add(quantized_zero, x)` with an output type carrying its own zero point and scale, and compares the model's output before and after `into_decluttered()`. It fails with the old inlet, giving `[0, 0, 0, 0]` instead of the requantized input.

PR #2639, opened 2026-08-16
wasm: SIMD reduction kernels for f32 and f16, with no unsafe

czoli1976:
`max_f32`, `min_f32`, `sum_f32`, `rms_norm_f32` and `softmax2_fastcompact_f32` had no wasm32 implementation and fell through to the generic scalar reductions, where LLVM cannot help: it will not reassociate a floating-point accumulator, so each reduction keeps a serial dependency chain one element long. This adds simd128 kernels — independent accumulators plus a tree-shaped horizontal reduction — bringing wasm to parity with the arm64 and x86_64 tiers, which already have all four.

The max and min lanes are combined with `total_cmp`, so results match the generic kernels exactly, NaN ordering included. The two kernels with a multiply-accumulate, softmax-L2 and rms_norm, go through the existing `madd_f32x4!` macro, so they are a plain mul+add by default and `f32x4.relaxed_madd` on `+relaxed-simd` builds — one source, both variants, no duplicated kernel. On this machine the relaxed build is worth about 1.16x on softmax-L2. max, min and sum have no multiply to fuse.

The remaining relaxed opcodes are deliberately not used. `f32x4.relaxed_max` and `relaxed_min` leave NaN behaviour to the engine, which would lose the `total_cmp` agreement with the generic kernels, and `u32x4.relaxed_trunc_f32x4` leaves out-of-range conversion to the engine, where the fast-compact exp depends on saturation to reach zero.

Kernel throughput under wasmtime 47.0.3, f32 and f16, invoked per 2048-element row (the frame shape
tract uses), generic and simd128 built into one binary and interleaved, best of 30 on an idle
machine, measured through tract's own dispatch. Two machines, so the win is not tied to one
microarchitecture:

| kernel | M1 Pro generic | M1 Pro simd128 | | M4 generic | M4 simd128 | |
| --- | --- | --- | --- | --- | --- | --- |
| max_f32 | 0.60 | 16.0 Gelem/s | 27x | 0.81 | 23.1 Gelem/s | 29x |
| min_f32 | 0.61 | 16.0 | 26x | 0.79 | 23.1 | 29x |
| sum_f32 | 1.12 | 20.0 | 18x | 2.14 | 29.2 | 14x |
| rms_norm_f32 | 0.99 | 7.9 | 8x | 1.63 | 10.4 | 6x |
| softmax-L2 f32 | 1.04 | 4.7 | 5x | 1.66 | 6.7 | 4x |
| max_f16 | 0.44 | 22.2 | 50x | 0.61 | 31.5 | 52x |
| sum_f16 | 0.80 | 3.0 | 4x | 1.01 | 4.0 | 4x |
| softmax-L2 f16 | 0.08 | 1.3 | 17x | 0.12 | 1.7 | 14x |

f16 needs a different approach than f32, because the f16x8 proposal is not exposed by stable Rust
and there is no native f16 SIMD arithmetic. `max_f16` converts nothing at all: f16 orders correctly
as sign-magnitude, so mapping each lane with `x ^ ((x >> 15) & 0x7fff)` gives a monotone i16
ordering that `i16x8_max` reduces directly, and the mapping is inverted once at the end. `sum_f16`
and the f16 softmax widen to f32 by bit surgery and accumulate there, matching what the AVX-512 f16
kernels already do; the softmax also narrows back to f16 in SIMD, since a per-lane `f16::from_f32`
costs more than the vectorisation saves.

There is no `unsafe` anywhere in the file. The wasm32 simd128 intrinsics are safe functions and only
raw-pointer loads need `unsafe`, so the f32 kernels move off `v128_load`/`v128_store` onto
`chunks_exact` and the lane constructors. That measured no slower on either machine (max and min
came out ~4% ahead, the rest inside noise) and makes the diff smaller.

`softmax2_fastcompact_f16` accumulates the pre-rounding f32 values rather than the f16-rounded ones
`HSoftMaxL2` sums. It stays inside the f16 Approximate contract on every test, and is if anything
more accurate, but it is a deviation from the generic's exact semantics.

End to end on an 8-head, seq-512, d-64 attention block — two einsums either side of a softmax — also under wasmtime. Lower is better; the last column isolates the softmax by subtracting the matmul-only row:

| | before | after | softmax portion |
| --- | --- | --- | --- |
| matmuls only, no softmax | 17.94 ms | 18.04 ms | control |
| with softmax, `Libc` (default) | 28.61 ms | 24.80 ms | 10.67 → 6.76 ms |
| with softmax, `FastCompact` | 23.32 ms | 18.59 ms | 5.38 → 0.55 ms |

The matmul-only row is the control: it exercises no reduction, and its 0.1 ms difference is measurement noise.

A softmax runs two reductions, a `max_f32` pass and then the exp-and-sum pass, so it picks up both kernels and improves by more than either alone — 5.77 ms to 0.61 ms predicted from the kernel figures above, against 5.38 to 0.55 ms measured. `max_f32` is called whichever exp implementation is selected, which is why the default path improves without opting into `FastCompact`.

Models with no softmax — RNN and CNN denoisers, for instance — see no change from this.

`cargo test -p tract-linalg --target wasm32-wasip1` in debug: 2230 passed, 0 failed on both an M1 Pro and an M4, including the frame test suites for all eight kernels. Native `cargo test -p tract-linalg`: 4424 passed, 0 failed. 🍍

### DIFF
--- a/linalg/src/wasm.rs
+++ b/linalg/src/wasm.rs
@@ -11,6 +11,7 @@ use crate::Ops;
 
 #[cfg(target_feature = "relaxed-simd")]
 use crate::frame::element_wise::ElementWiseKer;
+use crate::frame::reduce::{MapReduceKer, ReduceKer};
 
 #[macro_use]
 mod madd;
@@ -22,6 +23,7 @@ mod dispatch_tests;
 mod mmm_f32_gemm;
 mod mmm_f32_gemv;
 mod mmm_i32;
+mod reduce;
 
 #[cfg(target_feature = "relaxed-simd")]
 pub use act::*;
@@ -75,4 +77,8 @@ pub fn plug(ops: &mut Ops) {
         ops.sigmoid_f32 = Box::new(|| WasmSigmoid4Relaxed::ew());
         ops.tanh_f32 = Box::new(|| WasmTanh4Relaxed::ew());
     }
+    ops.max_f32 = Box::new(|| reduce::wasm_max_f32_32n::red());
+    ops.min_f32 = Box::new(|| reduce::wasm_min_f32_32n::red());
+    ops.sum_f32 = Box::new(|| reduce::wasm_sum_f32_32n::red());
+    ops.softmax2_fastcompact_f32 = Box::new(|| reduce::wasm_softmax2_fastcompact_f32_32n::red());
 }
--- a/linalg/src/wasm.rs
+++ b/linalg/src/wasm.rs
@@ -81,4 +81,5 @@ pub fn plug(ops: &mut Ops) {
     ops.min_f32 = Box::new(|| reduce::wasm_min_f32_32n::red());
     ops.sum_f32 = Box::new(|| reduce::wasm_sum_f32_32n::red());
     ops.softmax2_fastcompact_f32 = Box::new(|| reduce::wasm_softmax2_fastcompact_f32_32n::red());
+    ops.rms_norm_f32 = Box::new(reduce::rms_norm_f32);
 }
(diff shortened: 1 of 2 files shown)


czoli1976:
image

Another OpenEvolve Powered Optimisation

czoli1976:
Evolved Kernels
image

kali (approved):

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2640, opened 2026-08-16
Feat/bench dinghy riscv

kali:


kali:
@czoli1976 this is a beagle v ahead board, the visionfive is recalcitrant. it has some pre-standardization rvv, but i think this board will be our "scalar" representative, unless there is interest in pre-standard simd somewhere. It will be merged in a couple of days, when I get a dedicated PSU for the board.

The orange pi rv is on the move, so we may have an actual bench in a couple of weeks.

czoli1976:
Sweet

PR #2641, opened 2026-08-16
metal: size reduce_nd3 threadgroups from the reduced axis

czoli1976:
`reduce_nd3` dispatched one 32-thread simdgroup per output element, so a reduction with few output
elements used a fraction of the GPU however long the reduced axis was: `[64, 262144]` ran 2048
threads and reached 21 GB/s where the same 67 MB reduced as `[16384, 1024]` reached 159 GB/s.
Contiguous reductions over long axes now size the threadgroup from the axis and combine their
per-simdgroup partials through threadgroup memory; strided reductions keep the single simdgroup,
where the extra lanes would only land further apart. Indexing moves to 32-bit arithmetic, with a
64-bit `_large` variant selected above `u32::MAX` elements, since Apple GPUs emulate 64-bit integer
ops.

## Performance

`cargo bench -p tract-metal --bench metal_reduce`, sum f32, 20 dispatches per iteration, criterion
median. `tg` is the number of threadgroups the shape gives the GPU, `width` the threads per
threadgroup this PR picks (32 = unchanged from main).

| shape | tg | tg/core M1P : M4 | width | M1 Pro main | M1 Pro PR | | M4 main | M4 PR | |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| `64x262144` | 64 | 4 : 8 | 1024 | 21 GB/s | 167 GB/s | **7.87x** | 30 GB/s | 103 GB/s | **3.41x** |
| `256x65536` | 256 | 16 : 32 | 1024 | 81 GB/s | 167 GB/s | **2.06x** | 92 GB/s | 105 GB/s | **1.14x** |
| `1024x16384` | 1024 | 64 : 128 | 1024 | 81 GB/s | 174 GB/s | **2.15x** | 98 GB/s | 106 GB/s | 1.08x |
| `16384x1024` | 16384 | 1024 : 2048 | 256 | 159 GB/s | 183 GB/s | **1.15x** | 104 GB/s | 102 GB/s | 0.99x |
| `32768x333` | 32768 | 2048 : 4096 | 32 | 140 GB/s | 179 GB/s | **1.28x** | 101 GB/s | 103 GB/s | 1.02x |
| `65536x64` | 65536 | 4096 : 8192 | 32 | 77 GB/s | 159 GB/s | **2.07x** | 93 GB/s | 97 GB/s | 1.05x |
| `64x1024x512` | 32768 | 2048 : 4096 | 32 | 43 GB/s | 45 GB/s | 1.03x | 44 GB/s | 43 GB/s | 0.99x |

Same 67 MB working set in the first four rows; only the split between output elements and reduced
length changes. The gain tracks threadgroups per GPU core, not chip generation: expressed that way
the two machines' curves overlay, and the PR sits near each machine's ceiling (~183 GB/s M1 Pro,
~106 GB/s M4) at every shape, where main swings from 21 to 159.

```
GB/s, shared 0-200 scale.  main: #####    this PR: =====

M1 Pro  (16 GPU cores)
  64x262144    ####                                        21
               =================================          167   7.87x
  256x65536    ################                            81
               =================================          167   2.06x
  1024x16384   ################                            81
               ===================================        174   2.15x
  16384x1024   ################################           159
               =====================================      183   1.15x
  32768x333    ############################               140
               ====================================       179   1.28x
  65536x64     ###############                             77
               ================================           159   2.07x
  strided      #########                                   43
               =========                                   45   1.03x

M4  (8 GPU cores)
  64x262144    ######                                      30
               =====================                      103   3.41x
  256x65536    ##################                          92
               =====================                      105   1.14x
  1024x16384   ####################                        98
               =====================                      106   1.08x
  16384x1024   #####################                      104
               ====================                       102   0.99x
  32768x333    ####################                       101
               =====================                      103   1.02x
  65536x64     ###################                         93
               ===================                         97   1.05x
  strided      #########                                   44
               =========                                   43   0.99x

```

## Scope and costs

- At full occupancy on the M4 this is a measurable ~1.5% regression (`16384x1024`, non-overlapping
  criterion intervals) from the threadgroup-memory barrier. Recovering it means gating the wider
  threadgroup on occupancy, which needs the GPU core count; Metal does not expose it.
- The thresholds are a broad plateau: dividing the axis by 2 to 8, capping at 256 to 1024, and
  switching at 128 to 2048 all land within 5% of each other, so nothing is tuned to one machine.
  They match what MLX uses for its Metal row reductions.
- Reductions with very few output elements stay starved whatever the threadgroup width, since the
  grid itself is too small; that regime needs a split pass and is not addressed here.
- Reads are 4 scalars per thread rather than a `float4` cast: arena offsets are not guaranteed
  16-byte aligned, and the two perform the same here.
- Accuracy is unchanged or better. `max`/`min` stay bit-exact; the two-level tree shortens the f32
  accumulation chain, so `sum` error is never worse and is up to 4x smaller on long axes.

Reduce tests previously topped out at a reduced axis of 82, so nothing exercised a multi-simdgroup
threadgroup. The added long-axis cases cover it, including `MeanOfSquares`, whose normalization must
apply exactly once across the two levels; those cases fail if it is applied per simdgroup.

Kernel written with OpenEvolve. Related to #2365, which packs element-wise kernels into full-width
threadgroups. 🍍

### DIFF
--- a/metal/Cargo.toml
+++ b/metal/Cargo.toml
@@ -43,3 +43,7 @@ rand.workspace = true
 name = "metal_gemm"
 harness = false
 
+
+[[bench]]
+name = "metal_reduce"
+harness = false
--- a/metal/benches/metal_reduce.rs
+++ b/metal/benches/metal_reduce.rs
@@ -1,0 +1,55 @@
+use criterion::*;
+use tract_core::internal::*;
+use tract_gpu::tensor::{DeviceTensor, IntoDevice};
+use tract_metal::kernels::nn::reduce::{Reducer, metal_reduce_launch};
+use tract_metal::with_metal_stream;
+
+const DISPATCHES: usize = 20;
+
+fn bench_shape(
+    g: &mut BenchmarkGroup<measurement::WallTime>,
+    shape: &[usize],
+    axis: usize,
+    name: &str,
+) {
+    with_metal_stream(|stream| {
+        let len = shape.iter().product::<usize>();
+        let a = Tensor::from_shape(shape, &vec![1.0f32; len])?.into_device()?;
+        let mut o_shape = shape.to_vec();
+        o_shape[axis] = 1;
+        let output = unsafe { DeviceTensor::uninitialized_dt(a.datum_type(), &o_shape)? };
+        metal_reduce_launch(&Reducer::Sum, &a, axis, &output)?;
+        stream.wait_until_completed()?;
+
+        g.bench_function(name, |b| {
+            b.iter(|| {
+                for _ in 0..DISPATCHES {
+                    metal_reduce_launch(&Reducer::Sum, &a, axis, &output).unwrap();
+                }
+                stream.wait_until_completed().unwrap();
+            })
+        });
+        Ok(())
+    })
+    .unwrap()
+}
+
+fn reduce(c: &mut Criterion) {
+    let mut g = c.benchmark_group("reduce_sum_f32");
+    g.sample_size(20);
+    for (shape, axis, name) in [
+        (tvec![64usize, 262144usize], 1, "64x262144"),
+        (tvec![256, 65536], 1, "256x65536"),
+        (tvec![1024, 16384], 1, "1024x16384"),
+        (tvec![16384, 1024], 1, "16384x1024"),
+        (tvec![32768, 333], 1, "32768x333"),
+        (tvec![65536, 64], 1, "65536x64"),
+        (tvec![64, 1024, 512], 1, "64x1024x512_strided"),
+    ] {
+        bench_shape(&mut g, &shape, axis, name);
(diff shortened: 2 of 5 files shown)


czoli1976:
image

czoli1976:
@kali this one looks pretty good, it is the result of experimenting with Genetic Algo Approach to find the best candidate with OpenEvolve (Claude Caude driving OpenEvolve using Sonnet to generate Candidates).

It could scale across different machines even better if the number of GPU cores was known but that would require to add a dependcy: IOKit — specifically the AGXAccelerator service's gpu-core-count property. 

Which I did not do unless you agree to it, otherwise it is fine as it is in this PR without the IOKIT Dep

PR #2644, opened 2026-08-17
Proto/activation registry

kali:


kali:
image

@czoli1976 @cverrier WDYT ?

czoli1976:
image

czoli1976:
Deliberate non-gaps: the wasm Sigmoid generic cells aren't an oversight — [#2591](https://github.com/sonos/tract/pull/2591) > wrote simd128 sigmoid and tanh, measured them, and deleted them because the generic polynomial auto-vectorizes faster

czoli1976:
@kali  BTW, did you ever consider Google Highway as a stopgap at least?

kali:
> Deliberate non-gaps: the wasm Sigmoid generic cells aren't an oversight — [#2591](https://github.com/sonos/tract/pull/2591) > wrote simd128 sigmoid and tanh, measured them, and deleted them because the generic polynomial auto-vectorizes faster

Yeah, good point, I will allow annotations for "this is generic, but we can't do best". or this is "via-f32" but we can't do best.

kali:
> @kali BTW, did you ever consider Google Highway as a stopgap at least?

Nope. When I incepted tract, I had a look at possible candidates for low-level op delegation, but nothing was covering the targets that mattered to us, and never really revisited. I'm a bit reluctant to include a big dependency tbh.

kali:
Not sure if I want to merge this or not.

Issue #2345, opened 2026-06-04
ONNX import: GroupQueryAttention do_rotary + decode KV cache (ORT-GenAI fused exports, e.g. Gemma-3)


czoli1976:
Reporting concrete importer gaps found while trying to run a real Gemma-3 ONNX export through
tract. The model architecture is fully supported op-wise; the blockers are all in how the
ORT-GenAI **fused decode export** packs things into `com.microsoft.GroupQueryAttention`.

## Repro
```
onnx-community/gemma-3-1b-it-ONNX → onnx/model_q4.onnx     (int4 MatMulNBits, block_size 32)
tract model_q4.onnx -f onnx dump
→ GroupQueryAttention: internal rotary (do_rotary) is unsupported; apply RotaryEmbedding separately
```

## What the export uses (the GQA node)
```
attrs:  num_heads=4, kv_num_heads=1, scale=0.0625, local_window_size=512,
        softcap=0.0, do_rotary=1, rotary_interleaved=0
inputs: query, key, value, past_key, past_value, seqlens_k, total_seq,
        cos_cache, sin_cache, "", attention_bias
```

So beyond tract's current prefill-only / no-rotary `GroupQueryAttention`, this needs:

1. **`do_rotary=1`** — RoPE applied *inside* GQA to Q and K from `cos_cache`/`sin_cache`
   (half-rotation, `rotary_interleaved=0`). The rotation math already exists in tract's
   `RotaryEmbedding` handler; it just isn't wired from GQA.
2. **decode / `have_past`** — `past_key`/`past_value` inputs → concat with current → attend →
   `present_key`/`present_value` outputs. The current handler rejects any past KV (prefill only).
3. **RoPE positions from `seqlens_k`** — the op indexes the cos/sin tables at positions derived
   from the per-batch length, not from tensor shapes.
4. **`GatherBlockQuantized`** (1 node) — quantized embedding lookup contrib op, also unhandled.

## Note
None of this is an architecture gap — tract has every op the Gemma family needs
(`RotaryEmbedding`, GQA prefill, `SimplifiedLayerNormalization`, `Gelu`, and `local_window_size`
sliding-window from #2323). The blocker is purely the importer not unpacking the **fused
ORT-GenAI decode form**. The same gaps block the broader ORT-GenAI fused-export family
(Gemma, Phi-3, Llama "genai" exports), so handling them would unlock several models at once.

(For contrast: an optimum/transformers.js *decomposed* export — separate RoPE, no GQA op — does
[...]

PR #2645, opened 2026-08-17
onnx: import GroupQueryAttention decode steps and its internal rotary

czoli1976:
Extends the `com.microsoft.GroupQueryAttention` importer from prefill-only to the decode step and the node's own rotary, which is what ORT-GenAI fused exports emit — closing points 1–3 of #2345.

### What changed

- **KV cache.** `past_key`/`past_value` are concatenated onto the step and returned as `present_key`/`present_value`. No new mask math was needed: `wire_attention_mask` already builds a bottom-right-aligned band from `Range(kv_len - q_len, kv_len)`, so a single decode token attends the whole cache correctly.
- **`do_rotary=1`.** RoPE runs inside the node through the existing `ApplyRope` op, with the `cos_cache`/`sin_cache` halves widened to head size. Positions follow the ORT CPU kernel: `0..S` for a first prompt (a `Slice`, so symbolic sequence lengths still work) and `seqlens_k` for a generated token (a `Gather`, so it works without static shapes either). K is rotated before it enters the cache.
- **`attention_bias`** is folded into the causal band, since `Sdpa` replaces any supplied mask when `is_causal` is set.
- `causal=0` is honoured; `smooth_softmax`, `qk_output`, quantized KV, `sliding_window_cache`, packed QKV, `position_ids`, `head_sink` and fused Q/K norm are rejected with specific messages.

### Deliberately rejected: the subsequent-prompt case

A multi-token step against a non-empty cache is not a concatenation. Probing onnxruntime shows it writes the new K at cache **offset 0**, treating past/present as one aliased buffer — appending silently produces different numbers. The importer rejects that shape with an explanatory message rather than approximating it. Only a first prompt (empty cache) and single-token generation are accepted, which matches the envelope the CPU kernel documents.

### On the mask sentinel

Summing the causal band with `attention_bias` relies on the band filling with `dt.min_value()`. Filling with `-inf` instead is not an option: the band is built as `indicator * fill`, and `0 * -inf` is `NaN` on the kept entries. The finite sentinel is exact for any row with at least one unmasked key; the two sentinels only become indistinguishable on a row where every visible key is masked by the bias, which carries no meaningful output in any implementation. Noted in the code so the next reader does not repeat the experiment.

### Testing

A numpy reference was validated against onnxruntime first (prefill and decode agree to ~1e-7), then tract checked against it across prefill/decode × rotary/no-rotary × bias/no-bias × sliding-window: 10/10, max error 1.2e-7. Also covered the `[B, 1]` shaped `seqlens_k` that real exports emit, where the spec says `[B]`.

Two cases are added under `onnx/test_cases/`, generated from onnxruntime, passing all four passes including the NNEF round-trip (their `vars.sh` sets `--nnef-tract-transformers`, needed for `ApplyRope` to serialize).

`onnx/test_cases/run_all.sh` shows one unrelated failure, `qtdnn_10x5_101_i32_biases`, which fails identically on an unmodified build of the same commit.

🍍


kali (approved):

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2647, opened 2026-08-17
onnx: import GatherBlockQuantized without widening the table

czoli1976:
Adds the `com.microsoft.GatherBlockQuantized` importer. ORT-GenAI exports look up their embedding through it, so without a handler those models cannot be imported at all.

### Approach

The rows are gathered *before* being dequantized, so the table stays 4-bit in memory and only the selected rows are widened. Dequantizing the table up front — the straightforward lowering — would cost eight times its size for a vocabulary-sized embedding: on a 1.7B Qwen3 export that is 155 MB of int4 turning into 1.24 GB of f32, for a step that only ever reads a handful of rows.

Everything is expressed with existing ops, so there is no new eval kernel: `Gather` on the packed table, a nibble split (the values are integral and below 256, so `/16` and the remainder are exact in f32), a per-block broadcast of the scales, and `(value - zero) * scale`.

### Scope

Rank-2 table, `bits=4`, uint8 storage, `gather_axis` 0 (which the operator requires for uint8 anyway) and the last axis as `quantize_axis` — what the exports emit. Anything else is rejected with a specific message rather than silently mis-lowered. Absent `zero_points`, the zero defaults to 8 as the spec states.

### Testing

Checked against onnxruntime 1.28 across block sizes 16 / 32 / 128, with and without zero points, and for 1-D and 2-D index shapes: bit-exact in every case (the arithmetic is exact, so this is equality, not a tolerance). A case is added under `onnx/test_cases/` with onnxruntime-generated expectations, passing all four passes including the NNEF round-trip.

Note that onnxruntime only registers this op in reasonably recent builds — 1.19 is too old to run `generate_io.py`.

🍍


kali:
would it make sense to factorize this with eval_bq in tract_core Gather ? would the same trick make sense in the linalg supporting function ?

czoli1976:
Coming

czoli1976:
@kali  Done — pushed as a second commit, rebased on current main

Done = factorized this with eval_bq in tract_core, no need for  anything in linalg

czoli1976:
@kali  rerun me

kali (approved):

PR #2648, opened 2026-08-17
onnx: keep asymmetric MatMulNBits weights block-quantized

czoli1976:
`MatMulNBits` only kept its weight block-quantized when the export was symmetric. Every ORT-GenAI int4 export is asymmetric — all 197 nodes in a Bonsai/Qwen3 1.7B export carry zero points — so they all fell back to a dense f32 weight: eight times the memory, and enough to put a 1.7B model out of reach on a laptop.

### Approach

No new block-quant format is needed. Q4_0 fixes the zero point at 8, so the weight splits as

```
(q - z) * s  ==  (q - 8) * s  +  (8 - z) * s
```

The first term is *exactly* Q4_0, packed as today. The second is constant within a block, so its whole contribution is the block sums of the activations against an `[N, K/32]` constant — one thirty-second of the weight, and a matmul of one thirty-second the size. Both terms use the f16-rounded scale the packer already stores, so the split is exact against the rounded weight rather than introducing a second approximation.

The dense f32 fallback stays for the shapes that need it (other block sizes, a partial last block), and is now built only when that branch is actually taken instead of always.

### Effect

On `onnx-community/Bonsai-1.7B-ONNX/model_q4.onnx` (Qwen3, 28 layers, asymmetric int4) the weights stay `Q4_0` in memory: peak RSS 1.11 GB for load, optimize and an 8-token prefill, against roughly 7 GB of expanded f32 weights before. Logits match onnxruntime to 1e-4 absolute on values up to ±12.5 (4.4e-6 relative), with identical argmax on every position.

### Testing

Checked against onnxruntime across M/K/N shapes with and without zero points, and for block sizes on and off the fast path: relative error ~3e-4, the same order as the pre-existing symmetric path, which is the f16 scale rounding rather than the split. `onnx/test_cases/run_all.sh` shows no new failures.

🍍

### DIFF
--- a/onnx/src/ops/nn/mat_mul_nbits.rs
+++ b/onnx/src/ops/nn/mat_mul_nbits.rs
@@ -4,6 +4,7 @@ use tract_core::ops::cast::cast;
 use tract_core::ops::einsum::EinSum;
 use tract_core::ops::konst::Const;
 use tract_core::ops::math::add;
+use tract_core::ops::nn::{Reduce, Reducer};
 use tract_hir::internal::*;
 use tract_hir::ops::logic::wire_with_rank_broadcast;
 use tract_linalg::block_quant::{BlockQuantFact, BlockQuantStorage, Q4_0};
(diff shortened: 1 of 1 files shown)


kali (approved):

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

czoli1976:
🎉🎉🎉🎉🎉🎉

PR #2656, opened 2026-08-17
onnx: accept a MatMulNBits weight that is a constant expression

### LINKED ISSUE #2649: ONNX import: MatMulNBits rejects a weight behind a Reshape (tied lm_head in ORT-GenAI exports)
`MatMulNBits` reads its quantized weight straight off the input's `konst`:

https://github.com/sonos/tract/blob/main/onnx/src/ops/nn/mat_mul_nbits.rs

That holds when the weight is an initializer, but not when it is a constant *expression*. ORT-GenAI exports with tied embeddings feed the `lm_head` weight through a `Reshape` of the embedding table, so the import fails:

```
Error at stage "type"
  0: Translating node #1143 "/lm_head/MatMul_Quant" MatMulNBits ToTypedTranslator
  1: MatMulNBits: quantized weight B must be a constant
```

Seen on `onnx-community/Bonsai-1.7B-ONNX/model_q4.onnx`, where the node is

```
/lm_head/MatMul_Quant(
    /lm_head/num_logits_to_keep/Slice/output_0,
    model_embed_tokens_weight_quant_matmul,   <- Reshape(model_embed_tokens_weight_quant [151669,1024] -> [151669,64,16])
    model_embed_tokens_weight_scales,
    model_embed_tokens_weight_zp)
```

The reshape is a pure reinterpretation of the same bytes, so nothing but the fold is missing. This affects any tied-embedding export of this family, which is most of the small Qwen/Llama ORT-GenAI models.

`GatherBlockQuantized`, `MultiHeadAttention`, `fft` and `quant` take the same `.konst` shortcut and would trip on the same shape.

## How to fix

`TypedModel::compute_const_facts` already does exactly this job — it walks the eval order and fills in `konst` for any stateless node whose inputs are all plain constants:

https://github.com/sonos/tract/blob/main/core/src/model/typed.rs

Running it once before the expansions are wired makes `Reshape(initializer)` present itself as a constant, and the existing `.konst` reads then succeed unchanged. Narrowly, the importer could do it lazily at the point of failure:

```diff
-        let b_k = model
-            .outlet_fact(inputs[1])?
-            .konst
-            .clone()
-            .context("MatMulNBits: quantized weight B must be a constant")?;
+        if model.outlet_fact(inputs[1])?.konst.is_none() {
+            model.compute_const_facts()?;
+        }
+        let b_k = model
+            .outlet_fact(inputs[1])?
+            .konst
+            .clone()
+            .context("MatMulNBits: quantized weight B must be a constant")?;
```

I have not benchmarked what folding the whole graph costs at that point, so doing it once up front in the ONNX translation is probably the better placement than per-node.

Workaround meanwhile: rewrite the model so the reshaped weight is its own initializer. Because the reshape is a reinterpretation, the new `TensorProto` can copy the original's external-data reference and only change `dims`, so no weight bytes are touched.

czoli1976:
Fixes #2649.

`MatMulNBits` reads its quantized weight straight off the input's `konst`. That holds when the
weight is an initializer, but the ORT-GenAI exports do not always give it one: tied-embedding
models reach the `lm_head` weight through a `Reshape` of the embedding table, and the sub-4-bit
ones unpack every layer's weight through a `[256, 2]` lookup table. Both are constant, neither
is an initializer, and the import fails with `MatMulNBits: quantized weight B must be a
constant`.

### Why it only bites on real models

Small constant cones are already folded by the time the expansion runs. The gate is
`CONST_FOLD_MEM_BUDGET` in `hir/src/infer/ops.rs`: the analyser evaluates a stateless node only
while every input is plain and under the budget, and keeps the result only while
`output_mem <= input_mem.max(CONST_FOLD_MEM_BUDGET)`. At 4 MiB that is exactly the boundary I
measured -- a `[2048, 2048]` gathered `u8` result folds, `[3072, 2048]` does not.

So this is not a workaround for a gap in the analyser. The budget is deliberate, and an op that
structurally requires its weight as a constant needs a way past it.

### Approach

Walk the weight input's own constant cone and evaluate it, and only when its fact is actually
missing. `compute_const_facts` is not usable here: it drives off `eval_order`, which walks back
from the model outputs, and the model is still being translated when an expansion is wired, so
it has none yet. `eval_order_for_nodes` gives the same traversal scoped to one outlet.

The walk reads the graph and never edits it. Anything not stateless, or whose inputs are not
themselves constant, stops it, so this cannot turn a runtime input into a constant -- it only
resolves values the analyser skipped for size. The cone is left in place, dead once the weight
has been read, and the compaction after translation drops it.

A `ModelPatch` would be the idiomatic edit, and I tried it, but it cannot be applied from here.
`ModelPatch::apply` finishes by collecting every node its shunt left without a successor, and
mid-translation the model has no outputs to anchor anything -- folding the first layer's weight
retires the lookup table the remaining layers are still to be built from. That case is covered
by `shared_cone_source_survives_the_fold`.

### Testing

Two unit tests. `weight_behind_a_constant_expression` builds the same op twice, once with the
weight as an initializer and once behind a `Reshape` cone, and asserts the outputs match;
without the fix it fails with the reported error. `shared_cone_source_survives_the_fold` covers
the shared lookup table topology, and is what fails against the patch version.

No large fixture is needed for either: the budget only binds in the hir analyser, so a
hand-built typed model reaches `wire` with an unresolved fact for any cone over sixteen
elements. (The original description said a test case was impractical for this reason -- that
was wrong, and these two are the correction.)

`onnx/test_cases` does not cover this: none of its seventeen models contains a `MatMulNBits`
node, so the suite cannot reach the changed path. Worth noting separately that it is not
reliable to run locally either -- `run_all.sh` does not export `CACHEDIR`, so `cache_file.sh`
caches under `$HOME/.cache/tract-test-assets` while the suite looks in `.cached`; its failure
count on my box moved between 9 and 46 purely with how warm that cache was.

End to end, `onnx-community/Bonsai-1.7B-ONNX/model_q1.onnx` -- the model that motivated this,
whose 197 layer weights all arrive through the lookup table -- imports and runs, matching
onnxruntime 1.29 at 1.7e-6 relative with identical argmax at every position. That is measured
on a branch combining this with #2645, #2647 and #2655, since q1 needs all of them to run at
all.

Not writing `konst` facts also drops peak RSS importing that model from ~3.1 GB to ~1.9 GB
(interleaved A/B, four runs each); wall clock is unchanged, as expected, since both versions do
the identical cone walk.

Note the 2-bit exports also need #2655; this change is independent of it.

🍍

### DIFF
--- a/onnx/src/ops/nn/mat_mul_nbits.rs
+++ b/onnx/src/ops/nn/mat_mul_nbits.rs
@@ -34,6 +34,50 @@ pub fn mat_mul_nbits(
     Ok((expand(MatMulNBits { k, n, block_size, zp_input, bias_input }), vec![]))
 }
 
+/// Evaluates the constant cone feeding `outlet`, so a weight reachable only through constant
+/// ops reads back as a constant. `None` when the cone is not entirely constant: a node that is
+/// not stateless, or one whose inputs are not themselves constant, stops the walk.
+///
+/// This reads the graph and never edits it. A `ModelPatch` would be the idiomatic edit, but it
+/// cannot be applied from here: `wire` runs mid-translation, and `ModelPatch::apply` finishes
+/// by collecting every node the shunt left without a successor. Against a model that has no
+/// outputs yet, nothing anchors a constant whose remaining consumers are still to be
+/// translated -- the shared lookup table of a sub-4-bit export is collected while folding the
+/// first layer's weight, and the next layer then wires from a retired outlet. Leaving the cone
+/// in place costs nothing: it is dead once the weight is read, and the compaction after
+/// translation drops it.
+fn fold_const_cone(model: &TypedModel, outlet: OutletId) -> TractResult<Option<Arc<Tensor>>> {
+    let order =
+        tract_core::model::order::eval_order_for_nodes(&model.nodes, &[], &[outlet.node], &[])?;
+    let mut values: HashMap<OutletId, Arc<Tensor>> = HashMap::default();
+    for n in order {
+        let node = model.node(n);
+        for (slot, o) in node.outputs.iter().enumerate() {
+            if let Some(k) = o.fact.konst.as_ref().filter(|k| k.is_plain()) {
+                values.insert(OutletId::new(n, slot), k.clone());
+            }
+        }
+        if !node.op.is_stateless() || node.inputs.is_empty() {
+            continue;
+        }
+        let Some(inputs) = node
+            .inputs
+            .iter()
+            .map(|i| values.get(i).cloned().map(|t| t.into_tvalue()))
+            .collect::<Option<TVec<_>>>()
+        else {
+            continue;
+        };
+        let Ok(res) = node.op.eval_with_session(n, &TurnState::default(), inputs) else {
+            continue;
+        };
+        for (slot, output) in res.into_iter().enumerate() {
+            values.insert(OutletId::new(n, slot), output.into_arc_tensor());
+        }
+    }
+    Ok(values.remove(&outlet))
+}
+
 #[derive(Debug, Clone)]
 struct MatMulNBits {
     k: usize,
@@ -79,12 +123,16 @@ impl Expansion for MatMulNBits {
         let blob = block_size.div_ceil(2);
         let zp_blob = n_blocks.div_ceil(2);
 
-        // Read the constant quantized weight, scales and (optional) zero points.
-        let b_k = model
-            .outlet_fact(inputs[1])?
-            .konst
-            .clone()
-            .context("MatMulNBits: quantized weight B must be a constant")?;
+        // Read the constant quantized weight, scales and (optional) zero points. The weight
+        // need not be an initializer: tied-embedding exports reach it through a reshape, and
+        // 2-bit ones through a lookup table, both constant. The model is still being
+        // translated, so it has no outputs to drive a whole-graph fold; patch this input's
+        // own cone instead.
+        let b_k = match model.outlet_fact(inputs[1])?.konst.clone() {
+            Some(k) => k,
+            None => fold_const_cone(model, inputs[1])?
+                .context("MatMulNBits: quantized weight B must be a constant")?,
+        };
         let b_plain = b_k.try_as_plain()?;
         let b: &[u8] = b_plain.as_slice()?;
         let scales_k = model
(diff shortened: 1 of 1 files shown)


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali:
Can we rewrite the cone-folding function as a Patch generator ?

czoli1976:
ok, on it!

Il giorno ven 21 ago 2026 alle ore 12:25 Mathieu Poumeyrol <
***@***.***> ha scritto:

czoli1976:
Tried it, and it turns out this is the one place a patch cannot go.

`ModelPatch::apply` finishes by collecting every node the shunt left without a successor,
walking up from the shunted outlet. `Expansion::wire` runs mid-translation, so the target
model has no outputs yet and the `target.outputs` guard in that loop anchors nothing. On a
sub-4-bit export every layer unpacks its weight through the same `[256, 2]` lookup table, so
folding layer 0's weight shunts its `Gather`, the collector walks up to the shared table,
finds it momentarily consumerless because layers 1.. are still untranslated, and retires it.
The next layer then wires from a dead outlet:

    Error: Invalid outlet reference: 1/0

There is a test for that now (`shared_cone_source_survives_the_fold`), which fails exactly
that way against the patch version.

What I have done instead is drop the part you were really objecting to: the fold no longer
writes `konst` facts, or touches the graph at all. It takes `&TypedModel`, accumulates into a
local map, and returns the tensor. The cone is left in place, dead, and the compaction after
translation drops it.

That turned out to be worth doing on its own. Writing the facts kept every folded tensor alive
in the graph until compaction -- all 197 expanded tables at once on Bonsai-1.7B q1. Not writing
them takes peak RSS importing that model from ~3.1 GB to ~1.9 GB (interleaved A/B, 4 runs each,
same box). Wall clock is unchanged, as expected: both versions do the identical cone walk.

If you would rather have a real patch generator, the way to get one is to stop resolving the
weight in `wire`: make `MatMulNBits` a `TypedOp` and do this in `declutter`, where the model
has outputs, the collector is anchored, and returning a `TypedModelPatch` is the house style.
`PropConst` runs first in that pass list but stops at `CONST_FOLD_MEM_BUDGET`, so the op would
still need its own cone walk -- it could just return a patch safely. Happy to do that if you
want it, it is a bigger change than this PR.

czoli1976:
but if you want to do it @kali , I will not object :P

Issue #2619, opened 2026-08-12
Binary elementwise fast paths gated at 32 elements: ~20% left on small-tensor graphs


czoli1976:
`find_most_efficient_config` (core/src/ops/binary.rs) only picks the `OptBinByScalar` /
`OptBinUnicast` fast paths when the operand has more than `min_num_elements = 32`
elements; below that the node keeps the generic broadcast eval. On graphs made of many
small tensors that cutoff is well above the real crossover, and the generic path it
falls back to is much slower per node than the path being avoided.

Measured on UL-UNAS (ultra-light speech enhancement, 1263 nodes after optimise, most
elementwise tensors 1-24 elements) and GTCRN, single frame, single-threaded, median of
3, tract at fc55e6dc6:

| cutoff | UL-UNAS | GTCRN | DFN3 enc | DFN3 erb_dec | DFN3 df_dec |
|---|---|---|---|---|---|
| 32 (current) | 6.906 ms | 13.118 ms | 0.282 ms | 0.275 ms | 0.202 ms |
| 16 | 6.666 | 13.219 | 0.296 | 0.271 | 0.203 |
| 8 | 6.699 | 13.602 | 0.278 | 0.268 | 0.210 |
| 4 | 5.734 | 13.044 | 0.284 | 0.281 | 0.211 |
| 2 | 5.487 | 12.771 | 0.293 | 0.286 | 0.212 |

UL-UNAS gains 20.5% at a cutoff of 2, GTCRN 2.6%. Per-node, a generic `Mul` on a
24-element tensor measures ~9.2 us against ~0.6 us for the by-scalar path.

DeepFilterNet3 moves the other way, 4-5% slower across its three graphs, so simply
lowering the constant trades one model shape against another and is not a patch I can
offer as-is. Reporting the measurement in case the intended fix is to make the choice
depend on the shape rather than on a single global cutoff, or to reduce the fixed cost
of the fast paths so the guard is not needed at small sizes.

Environment: i7-3770 (AVX, no AVX2/FMA), Linux, single-threaded, release build.

Secondary, not reachable today: `by_scalar_is_possible` and `by_scalar_should_be_efficient`
are coupled only through the `TDim::Val(0)` sentinel, and the call site at binary.rs
selects `OptBinByScalar` from the latter alone. That holds only while
`min_num_elements >= 1` — `gt_tdim(TDim::Val(0), 0)` is true, so a cutoff of 0 would
[...]

PR #2661, opened 2026-08-18
core/binary: read the scalar in place when the by-scalar group is short

czoli1976:
# core/binary: read the scalar in place when the by-scalar group is short

Partially answers #2619.

A trailing broadcast such as `[1,256,1,2] * [1,256,1,1]` has a by-scalar group
of 2, so it fails the flat cutoff of 32 and runs the generic binary path at
6.6 ns/element — 3.4 us to move 512 elements. It is not the arithmetic: a plain
loop over the same shape takes **0.072 us**, so ~40x of that is dispatch and
broadcast iteration.

Lowering the cutoff alone does not fix it, and measurably hurts: at 2 the
FastEnhancer frame goes 0.135 -> 0.137 (tiny) and 0.274 -> 0.281 (base),
because `OptBinByScalar` then dispatches its kernel once per block — 256 calls
carrying two elements each.

So this changes both halves:

- the eval reads `b`'s scalar straight out of place for a short period, one
  pass, no broadcast buffer (an earlier attempt that materialised the broadcast
  with `broadcast_to_shape` was a wash — the allocation costs what the
  dispatches saved);
- the choice weighs the total element count rather than the group alone, which
  is what decides whether the op pays off. The added `>= 2` term also confirms
  `b` really has a trailing unary axis: identical shapes satisfy
  `check_input_shapes` vacuously, and the by-scalar eval has no scalar to
  broadcast in that case.

Output is bit-identical on ten models over a full streaming pass;
`test-unit-core` 816 and `test-onnx-core` 4528 pass; fmt and clippy clean.

🍍

### DIFF
--- a/core/src/ops/binary.rs
+++ b/core/src/ops/binary.rs
@@ -409,7 +409,11 @@ impl TypedOp for TypedBinOp {
                         model,
                         node,
                         &inputs,
-                        OptBinByScalar { binop: actual_core_op, eval_fn },
+                        OptBinByScalar {
+                            binop: actual_core_op,
+                            eval_fn,
+                            linalg_op: actual_linalg_op,
+                        },
                     )?
                     .with_context("ByScalar"),
                 ));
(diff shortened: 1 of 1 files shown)


kali (approved):

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2667, opened 2026-08-19
py: keep setuptools installable for cp39 wheel builds

kali:
setuptools 84 dropped Python <3.10 support, breaking the isolated build env for the cp39 wheel target. Mark the requirement by python_version so 3.9 keeps the older pin and 3.10+ gets the bump.


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

Issue #2670, opened 2026-08-19
Sign(-0.0) returns -0.0 instead of 0 for float tensors


keevin-xu:
`Sign` returns `-0.0` when given `-0.0`. ONNX specifies `0`.

The [ONNX `Sign` reference](https://onnx.ai/onnx/operators/onnx__Sign.html) says, identically at
versions 9 and 13:

> Calculate the sign of the given input tensor element-wise. If input > 0, output 1. if input < 0,
> output -1. **if input == 0, output 0.**

`-0.0 == 0` is true in IEEE-754, so the third branch applies and the result should be `0`.

## Reproduction

A single-node `Sign` model over one `float32` input. Showing bit patterns, since `-0.0` and `0.0`
print identically:

```
input      -0.0        0.0         -2.0        2.0
           0x80000000  0x00000000  0xc0000000  0x40000000

tract      0x80000000  0x00000000  0xbf800000  0x3f800000     <- -0.0
```

For comparison, on the same model:

```
onnx.reference (onnx 1.22.0)   0x00000000  ...   <- 0.0
onnxruntime 1.28.0             0x00000000  ...   <- 0.0
candle-onnx 0.11.0             0x00000000  ...   <- 0.0
```

`float64` behaves the same way, returning `0x8000000000000000`.

Smallest failing case found by an automatic minimiser: `Sign` of a `float32` tensor of shape
`[1, 1, 1]` containing `-0.0`.

### Steps to reproduce

Builds the model, checks it, saves it, and prints what the specification's own implementation
returns. Needs only `pip install onnx` (tested with onnx 1.22.0). Verified to run as shown.

```python
import numpy as np, onnx
from onnx import TensorProto as T, helper as h
from onnx.reference import ReferenceEvaluator

g = h.make_graph([h.make_node("Sign", ["a"], ["y"])], "sign",
                 [h.make_tensor_value_info("a", T.FLOAT, [1])],
                 [h.make_tensor_value_info("y", T.FLOAT, [1])])
m = h.make_model(g, opset_imports=[h.make_opsetid("", 22)]); m.ir_version = 10
onnx.checker.check_model(m); onnx.save(m, "repro.onnx")

r = ReferenceEvaluator(m).run(None, {"a": np.array([-0.0], np.float32)})[0]
print(hex(r.view(np.uint32)[0]))   # 0x0  -- tract gives 0x80000000
```

`repro.onnx` is a complete single-node model; loading it with tract is the whole reproduction.

## Where it comes from

[...]

PR #2671, opened 2026-08-19
fix(core): Sign returns +0 for -0.0

keevin-xu:
Addresses issue #2670

Sign's zero branch returned the input unchanged, so -0.0 came back as -0.0. ONNX specifies 0 when input == 0, and -0.0 == 0 is true in IEEE-754.

One line in the signed/float arm; the unsigned arm is already correct, since those types have no signed zero. The test compares bit patterns, because -0.0 == 0.0 is true and a value assertion passes on the bug — I checked it fails without the fix.

cargo test -p tract-core: 277 passed. Also verified with an independent differential harness against onnx.reference and ONNX Runtime.

### DIFF
--- a/core/src/ops/math/mod.rs
+++ b/core/src/ops/math/mod.rs
@@ -834,7 +834,7 @@ mod tests {
     fn sign_of_negative_zero_is_positive_zero() -> TractResult<()> {
         let mut t = tensor1(&[-0.0f32, 0.0, -2.0, 2.0]);
         Sign {}.eval_in_place(&mut t, None)?;
-        let got = unsafe { t.as_slice_unchecked::<f32>() };
+        let got = t.try_as_plain()?.as_slice::<f32>()?;
         // Compared as bit patterns: -0.0 == 0.0 is true, so an equality check on the
         // values would pass even when -0.0 is returned.
         assert_eq!(got[0].to_bits(), 0f32.to_bits());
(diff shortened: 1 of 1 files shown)


kali (dismissed):

core/src/ops/math/mod.rs
```diff
@@ -829,4 +829,18 @@ mod tests {
         assert!(op.0.downcast_ref::<ShiftRight>().is_some());
         Ok(())
     }
+
+    #[test]
+    fn sign_of_negative_zero_is_positive_zero() -> TractResult<()> {
+        let mut t = tensor1(&[-0.0f32, 0.0, -2.0, 2.0]);
+        Sign {}.eval_in_place(&mut t, None)?;
+        let got = unsafe { t.as_slice_unchecked::<f32>() };
```
cverrier: The following form is both safe and already extensively used in `core/src`, *e.g.* in https://github.com/sonos/tract/blob/18363fc4f37b1278668df8eacf21395bd7016f90/core/src/ops/binary.rs#L1306

```suggestion
        let got = t.try_as_plain()?.as_slice::<f32>()?;
```

cverrier:
@keevin-xu FYI in #2673 I've already addressed the "quantized arm" issue you raised in #2670. (thanks!)

kali (approved):

PR #2672, opened 2026-08-19
fix(onnx): DynamicQuantizeLinear rounds ties to even

keevin-xu:
ONNX specifies y = saturate(round(x / y_scale) + y_zero_point) for DynamicQuantizeLinear, with rounding to nearest, ties to even. The QuantizeLinear page states it verbatim: "For (x / y_scale), it rounds to the nearest even." Rust's f32::round rounds ties away from zero, so the two differ on every tie whose integer part is even.

Two sites are affected:
1. The quantized value. DynamicQuantizeLinear on [-127, 128, 0.5] returns [0, 255, 128], where onnx.reference and ONNX Runtime both return [0, 255, 127]. This is the one I measured.
2. The derived zero point in scale_and_zero_point, which applies the same rounding to -min/scale. I have not caught this one in the wild, but [-253, 257] puts it exactly on 126.5, and it comes out 127 instead of 126, shifting every output by one.

The test covers both, and fails if either site alone is reverted. The existing test_dynamic_quantize_linear_u8 is unaffected, since none of its data lands on a tie.

I also verified against a differential harness comparing onnx.reference, ONNX Runtime and tract.

### DIFF
--- a/onnx/src/ops/quant.rs
+++ b/onnx/src/ops/quant.rs
@@ -213,7 +213,8 @@ impl Expansion for DynamicQuantizeLinear {
 }
 
 fn dynamic_quantize_linear_f32_u8(x: f32, scale: f32, zero_point: u8) -> u8 {
-    (((x / scale).round() as i32) + zero_point as i32).clamp(u8::MIN as i32, u8::MAX as i32) as u8
+    (((x / scale).round_ties_even() as i32) + zero_point as i32)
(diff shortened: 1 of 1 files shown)


kali:
please ignore the bench failures, this is on me.

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2677, opened 2026-08-19
softmax: back the accurate f32 exp with a vectorized kernel

czoli1976:
# softmax: back the accurate f32 exp with a vectorized kernel

`SoftmaxExp` offers two tiers and the accurate one was the slow one:
`Libc` ran a scalar libm `exp` per element with a serial sum, while
`FastCompact` (Schraudolph) is vectorized but measures **5.9e-2** max relative
error. There was nothing in between, so an accurate softmax cost a libm call per
element.

This adds `SSoftMaxL2Accurate`: a Cody-Waite argument reduction and a degree-6
minimax fit of `e^r` on `|r| <= ln2/2`, every step an FMA so the row loop
vectorizes, with four accumulators to keep the sum off one dependency chain.
`SoftmaxExp::Libc` routes to it.

It is faster **and** slightly more accurate than what it replaces, so there is no
trade to weigh:

| | max rel err | note |
|---|---|---|
| `Libc` (scalar libm) | 1.42e-6 | f32 sum accumulation |
| **this kernel** | **1.22e-6** | four accumulators sum pairwise |
| onnxruntime | 1.19e-6 | for reference |
| `FastCompact` | 5.9e-2 | unchanged |

## Effect

Microbench over the row lengths a frequency-axis attention produces (16..64),
interleaved A/B: **2.1x** (0.632 vs 0.301 Gelem/s).

End to end on the FastEnhancer speech-enhancement family (Apple M1 Pro, single
thread, min per frame over a 627-frame streaming pass):

| variant | before | after | |
|---|---|---|---|
| tiny | 0.127 ms | 0.116 | 9% |
| base | 0.271 | 0.236 | 13% |
| small | 0.483 | 0.433 | 10% |
| medium | 0.814 | 0.726 | 11% |
| large | 1.790 | 1.596 | 11% |

Agreement with onnxruntime over the full streaming pass is unchanged at
2.1-2.3e-05 max absolute, with no drift across frames.

## The map neutral

`map_neutral` is `f32::NEG_INFINITY`, not `f32::MIN` as the sibling kernel uses.
`max_f32` pads with `f32::MIN`, so a fully masked row maxes to `f32::MIN`; a
padding lane then reduces to `exp(f32::MIN - f32::MIN) = exp(0)` and adds one to
the sum, which the scalar form never did because it iterated the row directly.
With `-inf` the padding underflows to zero for any max.

**`SSoftMaxL2` still carries that neutral**, so `FastCompact` has the same
latent behaviour on fully masked rows. Left alone here rather than changed
untested, but worth a look.

The `x < -103.0 || x > 0.0` guard is deliberately not a range check: `contains`
is true-by-negation for NaN, which would return zero where the scalar form
propagates NaN — a fully masked row must still reduce to NaN.

## Checks

`test-unit-core` 816 pass, `test-onnx-core` 4528 pass, full workspace suite,
`cargo fmt` and `cargo clippy` clean.

🍍

### DIFF
--- a/core/src/ops/nn/softmax/mod.rs
+++ b/core/src/ops/nn/softmax/mod.rs
@@ -309,12 +309,7 @@ impl Softmax {
             SoftmaxKind::Softmax(exp_impl) => {
                 let sum = match exp_impl {
                     SoftmaxExp::Libc => {
-                        let mut s = f32::zero();
-                        slice.iter_mut().for_each(|x| {
-                            *x = (*x - max).exp();
-                            s += *x;
-                        });
-                        s
+                        (tract_linalg::ops().softmax2_f32)().run_with_params(slice, max)?
                     }
                     SoftmaxExp::FastCompact => (tract_linalg::ops().softmax2_fastcompact_f32)()
                         .run_with_params(slice, max)?,
--- a/linalg/src/generic/reduce.rs
+++ b/linalg/src/generic/reduce.rs
@@ -186,6 +186,73 @@ pub mod softmax_l2 {
         }
     );
 
+    /// exp(x - max) with a Cody-Waite reduction and a degree-6 fit: accurate to
+    /// about 1e-7 relative, against [`fast_compact_exp_f32`]'s 6e-2, while still
+    /// being all FMAs so the row loop vectorizes.
+    #[inline(always)]
+    pub fn accurate_exp_f32(x: f32) -> f32 {
+        const LOG2E: f32 = 1.442_695_04;
+        const LN2_HI: f32 = 0.693_145_75;
+        const LN2_LO: f32 = 1.428_606_8e-6;
+        const MAGIC: f32 = 12_582_912.0;
+        // The argument is `v - max` over a row, so it is never positive; a
+        // positive value only arises from the f32::MIN lanes a short row is
+        // padded with, and below -103 exp underflows to zero anyway. Both
+        // comparisons are false for NaN, so a fully masked row still reduces to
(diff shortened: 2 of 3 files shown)


czoli1976:
back to draft as seems to need a little fix

czoli1976:
Pushed a hand-written kernel for the accurate exp, because "vectorized" in the original description overstated what this branch shipped: it relied on the compiler vectorizing the scalar loop, and on wasm that never happens.

wasm32 has no scalar fused multiply-add, so every `mul_add` in the polynomial lowers to a libm `fmaf` **call**, and LLVM does not vectorize f32 reductions on that target at all. Measured on the exp+sum loop under wasmtime:

| form | Gelem/s |
|---|---|
| scalar, `mul_add` | 0.068 |
| scalar, plain `mul` + `add` | 0.201 |
| simd128 | 1.47 |

End to end that made this branch a **~13% regression on wasm** for a GRU/attention speech-enhancement model (softmax went from 7.6% to 22.1% of the frame). With the NEON and simd128 bodies it is back to a win on both targets.

The kernel also needs the bound the scalar form already applies when it rebuilds the scale — `((k + 127).clamp(1, 254) as u32) << 23`. Shifting `k + 127` unclamped builds `0xFF800000` = -inf once the argument goes past about -88, and the row sum comes back NaN; `ops::nn::softmax::test::test_softmax_prop` catches it on a quantized case.

Green on aarch64 (4436 + 275) and on `wasm32-wasip1` under wasmtime (2210 + 274).

kali (approved):

PR #2692, opened 2026-08-21
(perf) core/cnn: fuse depthwise deconvolution instead of an outer-product einsum

czoli1976:
`wire_with_deconv_sum` lowers `ConvTranspose` to `EinSum("mk,kn->mn")` + `DeconvSum`, where `k` is input-channels-per-group. For a **depthwise** deconvolution `k` is 1, so the einsum degenerates to an outer product: it materialises `[N, C, HkWk, HW]`, larger than either operand, which `DeconvSum` then scatters into the output.

Because the result is bigger than both inputs, every optimised binary path is correctly rejected — they all require the output to share operand A's shape — so the multiply runs through the generic N-D broadcast walk. GTCRN builds `[1, 16, 9, 363]`, 3267 products per channel, to produce `[1, 16, 1, 33]`, 33 per channel. `DeconvSum`'s own loop already skips out-of-range positions, so most of those products are computed and never read.

This folds the multiply into the accumulation. The rank-2 path mirrors `main_loop_2d`'s pointer walk, so the geometry bounds are still tested once per `(kx, ix, ky, iy)`, with channels outer and the contiguous axis inner. No intermediate is allocated.

### Numbers

GTCRN's three depthwise ConvTransposes, summed over a 627-frame run:

| | einsum `Mul` + `DeconvSum` | fused |
|---|---|---|
| total | ~68 ms | ~11.7 ms |

End to end, interleaved against `main`, 5 rounds, medians, aarch64 and wasm32 plain simd128:

| model | native | wasm |
|---|---|---|
| gtcrn | **0.95** | **0.96** |
| DeepFilterNet3 erb_dec | **0.95** | 0.99 |
| the other nine in the set | 0.99–1.02 | 0.99–1.01 |

The set is five FastEnhancer variants, GTCRN, two DTLN submodels and three DeepFilterNet3 submodels; the nine without a depthwise ConvTranspose sit inside the ±3% run-to-run band.

### Correctness

Output is **byte-identical** to the einsum path — same products, same accumulation order — verified on GTCRN (1.29 MB of output) and on DeepFilterNet3's erb_dec (2.56 MB), comparing the fused path against `TRACT_DISABLE_DEPTHWISE_DECONV=1` in the same binary.

`cargo test --workspace` (macOS, minus `test-tflite` / `tract-proxy` / `tract-proxy-sys` / `test-cuda`), run with the two deconv files at this branch and reverted to `main` in the same tree: 106 suites, 65765 passed both ways, with an identical set of 70 pre-existing failures.

### Scope

Selected only for `group == input_channels == output_channels` at rank 2 — the path with the specialised pointer walk. The generic fused loop builds a coordinate per output element and is far slower than the einsum it would replace, so every other rank and grouping keeps the existing lowering. `TRACT_DISABLE_DEPTHWISE_DECONV` restores it for A/B.

### DIFF
--- a/core/src/ops/cnn/deconv/deconv_sum.rs
+++ b/core/src/ops/cnn/deconv/deconv_sum.rs
@@ -521,7 +521,7 @@ crate::declare_knob!(
 ///
 /// Inputs: `kernel` `[C, HkWk, 1]`, `input` `[N, C, 1, HW]`, `bias` (output-shaped).
 #[derive(Clone, Debug, new, Hash, PartialEq, Eq)]
-pub struct DepthwiseDeconvSum {
+pub struct DepthwiseDeconv {
     pub pool_spec: PoolSpec,
     pub kernel_format: KernelFormat,
     pub input_shape: ShapeFact,
@@ -529,14 +529,14 @@ pub struct DepthwiseDeconvSum {
     pub group: usize,
 }
 
-impl Op for DepthwiseDeconvSum {
+impl Op for DepthwiseDeconv {
     fn name(&self) -> StaticName {
-        "DepthwiseDeconvSum".into()
+        "DepthwiseDeconv".into()
     }
     op_as_typed_op!();
 }
 
-impl EvalOp for DepthwiseDeconvSum {
+impl EvalOp for DepthwiseDeconv {
     fn is_stateless(&self) -> bool {
         true
     }
@@ -580,7 +580,7 @@ impl EvalOp for DepthwiseDeconvSum {
     }
 }
 
-impl DepthwiseDeconvSum {
+impl DepthwiseDeconv {
     /// Rank-2 fused path, mirroring `main_loop_2d`'s pointer walk: the geometry
     /// bounds are tested once per (kx, ix, ky, iy) as there, and the innermost loop
     /// runs over channels, where kernel and input are both contiguous.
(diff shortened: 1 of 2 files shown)


core/src/ops/cnn/deconv/deconv_sum.rs
```diff
@@ -500,3 +500,260 @@ impl_eval! {
...
+/// For a depthwise ConvTranspose the einsum's contraction dim is 1, so it degenerates
+/// to an outer product whose result is larger than either operand — and most of it is
+/// discarded: GTCRN materialises `[1, 16, 9, 363]` (3267 products per channel) to
+/// produce `[1, 16, 1, 33]` (33 per channel). Roughly 91% of those products fall
+/// outside the output and are thrown away. Folding the multiply inside the bounds test
+/// skips them and never allocates the intermediate.
+///
+/// Inputs: `kernel` `[C, HkWk, 1]`, `input` `[N, C, 1, HW]`, `bias` (output-shaped).
+#[derive(Clone, Debug, new, Hash, PartialEq, Eq)]
+pub struct DepthwiseDeconvSum {
```
kali: Just `DepthwiseDeconv` maybe ? or `FusedDepthwiseDeconv` ? The "Sum" in DeconvSum name was because it was "doing the sum in the deconv".

czoli1976: right, changing to DepthwiseDeconv

czoli1976:
?

Il giorno mer 26 ago 2026 alle ore 09:41 github-actions[bot] <
***@***.***> ha scritto:

czoli1976:
@kali , your wish has been granted, rename in place

kali:
when in doubt...

czoli1976:
Always ask the Oracle

Il giorno mer 26 ago 2026 alle ore 11:06 Mathieu Poumeyrol <
***@***.***> ha scritto:

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2693, opened 2026-08-21
Feat/softmax x86 accurate

kali:


kali:
@czoli1976 I've killed the fast compact approximation, it was opt-in so nothing was using it, in favor or the poly you introduced recently for expf. Do you want to check this does not degrade wasm in significant any way?

czoli1976:
OK, noted, will check

Il giorno ven 21 ago 2026 alle ore 14:19 Mathieu Poumeyrol <
***@***.***> ha scritto:

czoli1976:
@kali does not degrade wasm

PR #2697, opened 2026-08-22
(perf) linalg/wasm: add a 4x16 f32 GEMM kernel and pick it by N

czoli1976:
`ops.mmm_f32` ignored `m`, `k` and `n` and returned `wasm_f32_8x8` for every f32 GEMM, so one square tile served every shape. A square tile is not the best shape here: each k-step splats MR values from A and loads NR/4 vectors from B, and **every splat feeds NR/4 multiply-adds**, so widening NR amortises the splat while widening MR does not.

`wasm_f32_4x16` keeps the same 16 v128 accumulators, laid out 4 rows × 4 vectors instead of 8 × 2.

### Numbers

Full tiled GEMM under node, plain `+simd128`, medians of 3 interleaved rounds, against `wasm_f32_8x8`:

| shape | ratio | shape | ratio | shape | ratio |
|---|---|---|---|---|---|
| 64×64×192 | 1.12 | 256×64×64 | 1.10 | 64×32×64 | 1.09 |
| 64×64×64 | 1.10 | 512×64×64 | 1.10 | 64×48×64 | 1.08 |
| 64×128×64 | 1.09 | 1024×64×64 | 1.10 | 256×256×256 | 1.13 |
| 64×256×64 | 1.10 | 1×64×128 | 2.19 | | |

It only pays when N fills the tile. At **N=8 a 16-wide tile wastes half its columns and runs at 0.58×**; N=24 pays that on its second tile (0.86×). So `mmm_f32` gates on `N % 16 == 0` and leaves everything else — including the case where N is not known at optimisation time — on 8×8. An `N >= 16` gate is not sufficient: it cost two models 4% each.

Large M does not regress: MR=4 needs twice the row-tiles of MR=8 and therefore re-reads the B panel twice as often, but the ratio holds at 1.10 through M=1024.

End to end on a speech-enhancement set (five FastEnhancer variants, GTCRN, two DTLN stages, three DeepFilterNet3 submodels), wasm32 plain simd128, five interleaved rounds:

| model | vs before | model | vs before |
|---|---|---|---|
| fastenhancer_m | **0.94** | fastenhancer_t | 0.98 |
| fastenhancer_l | **0.94** | fastenhancer_b | 0.98 |
| dfn3 erb_dec | **0.97** | dfn3 enc | 0.99 |
| fastenhancer_s | **0.97** | GTCRN, DTLN 1 & 2 | 1.00 |

No model regresses on either target. Native is untouched — the module is behind `cfg(all(target_family = "wasm", target_feature = "simd128"))`.

### Testing

`cargo test --target wasm32-wasip1 -p tract-linalg -p tract-core` under wasmtime, both `+simd128` and `+simd128,+relaxed-simd` as `.travis/cross.sh` runs them: all green, including the `packed_packed` proptests against the new kernel. The wasm backend picks its multiply-add form at compile time, so both configurations matter.

`cargo test --workspace` on macOS (minus `test-tflite` / `tract-proxy` / `tract-proxy-sys` / `test-cuda`): 106 suites, 65761 passed, with the same 70 pre-existing failures as `main` — the changed files are not compiled on a native target.

`dispatch_tests.rs` gains a trace asserting which kernel each shape lands on, in the style of the existing `dfn3_shapes` test: N of 64/32/48 route to 4×16, N of 8/12/24/36 stay on 8×8.

🍍

### DIFF
--- a/linalg/src/wasm/dispatch_tests.rs
+++ b/linalg/src/wasm/dispatch_tests.rs
@@ -281,6 +281,7 @@ fn add_row_col_products_and_add_mat_mul_agree_on_fusion() {
     check_madd_pairing(&*crate::wasm::wasm_f32_16x1);
     check_madd_pairing(&*crate::wasm::wasm_f32_32x1);
     check_madd_pairing(&*crate::wasm::wasm_f32_8x8);
+    check_madd_pairing(&*crate::wasm::wasm_f32_4x16);
 }
 
 /// `wasm_f32_4x4` is registered at `TargetOptimized` while every other kernel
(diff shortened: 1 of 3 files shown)


kali:
I had "the help" rebase that one for you.

czoli1976:
Thanks!

Il giorno dom 23 ago 2026 alle ore 23:10 github-actions[bot] <
***@***.***> ha scritto:

czoli1976:
Rebasing this onto current main and re-verifying before pushing an update — will follow up once the full workspace suite is green.

kali:
how do you like the new linalg ?

kali:
i mean, it's not there yet, i still have one branch of refactoring active. if it's confusing leave it to me.

czoli1976:
Have my imaginary friend rebase some, may help, may not

czoli1976:
Rebased onto current main and pushed. The 6 failing test-suite targets I saw are all confirmed pre-existing on a clean main (same failures, same counts) — a doctest fence-annotation issue in wasm.rs's own top comment and 3 unrelated tensorflow/prost doctest failures. Nothing here is from this PR.

kali:
Rebased it again, hopefully the declarative logic still works ?

kali:
I'm done with breaking linalg for now :) This kernel was not in my "dont change picks" list, and its declaration was rebased/ported multiple times, so I think it's worth a check if you have a minute.

czoli1976:
Checked. Good to go.

Kernel Intact, Routing Behaviour preserved.

PR #2699, opened 2026-08-22
onnx: honour allowzero in Reshape

### LINKED ISSUE #2698: Reshape ignores the allowzero attribute
`tract` does not read the `Reshape` attribute `allowzero`. A `0` in the target shape is always
treated as "copy the corresponding input dimension", which is the `allowzero=0` behaviour. With
`allowzero=1` it should mean a literal zero-length dimension.

## What ONNX specifies

From the [`Reshape` reference](https://onnx.ai/onnx/operators/onnx__Reshape.html):

> "If 'allowzero' is set, and the new shape includes 0, the dimension will be set explicitly to
> zero (i.e. not taken from input tensor)."

`allowzero` was added in opset 14.

## Reproduction

```
Reshape, opset 22, attribute allowzero = 1
  input  a : float32[3, 0]   (empty, zero elements)
  input  b : int64[1] = [0]  (initializer, the target shape)
  output   : float32[0]
```

```
onnx.reference (onnx 1.22.0)  ->  ok, float32[0], empty
onnxruntime 1.29.0            ->  ok, float32[0], empty
tract 0.23.4                  ->  fails during analysis
```

tract's full error chain names the computed shape, which is what identifies the cause:

```
0: Failed analyse for node #2 "y" Reshape
1: Infering facts
2: Applying rule GivenRule { (inputs[0].shape, inputs[1]) }
3: Reshaping [Val(3), Val(0)] to [Val(0)]
4: Reshape volume mismatch: input [Val(3), Val(0)] (vol=0) vs shape [Val(3)] (vol=3)
```

The requested target was `[0]`. tract resolved it to `[3]`, the input's dimension 0, then rejected
the model because 3 elements were requested against 0 held. With `allowzero=1` the target should
stay `[0]`, which is what `onnx.reference` and ONNX Runtime produce.

### Steps to reproduce

Needs `pip install onnx`. Writes a complete single-node model; loading it with tract is the whole
reproduction.

```python
import numpy as np, onnx
from onnx import TensorProto as T, helper as h, numpy_helper as nh
from onnx.reference import ReferenceEvaluator

shape = nh.from_array(np.array([0], np.int64), "shape")
g = h.make_graph([h.make_node("Reshape", ["a", "shape"], ["y"], allowzero=1)], "reshape",
                 [h.make_tensor_value_info("a", T.FLOAT, [3, 0])],
                 [h.make_tensor_value_info("y", T.FLOAT, [0])], [shape])
m = h.make_model(g, opset_imports=[h.make_opsetid("", 22)]); m.ir_version = 10
onnx.checker.check_model(m); onnx.save(m, "repro.onnx")

r = ReferenceEvaluator(m).run(None, {"a": np.zeros((3, 0), np.float32)})[0]
print(r.shape, r.size)   # (0,) 0
```

## Where it happens

The ONNX Reshape registration ignores the node, so no Reshape attribute is read.
([`tract-onnx/src/ops/array/mod.rs`](https://github.com/sonos/tract/blob/main/onnx/src/ops/array/mod.rs)):

```rust
reg.insert("Reshape", |_, _| Ok((expand(array::Reshape::default()), vec![])));
```

`Reshape` is a unit struct with nowhere to carry the flag
([`tract-hir/src/ops/array/reshape.rs`](https://github.com/sonos/tract/blob/main/hir/src/ops/array/reshape.rs)),
and the substitution is unconditional in `compute_shape_with_tf_rules`
([`tract-core/src/ops/change_axes.rs`](https://github.com/sonos/tract/blob/main/core/src/ops/change_axes.rs)):

```rust
// Replace 0s with corresponding input dims (positional, per ONNX/TF spec)
for (i, s) in shape.iter_mut().enumerate() {
    if *s == 0.into() {
        *s = input.get(i)...
    }
}
```

That comment is accurate for ONNX only when `allowzero` is unset. The string `allowzero` does not
appear anywhere in `tract-onnx`, `tract-hir` or `tract-core`.

## A regression test already exists

ONNX ships `test_reshape_allowzero_reordered` as part of its node test data. It is not in
`test-rt/suite-onnx/node.txt`, while `test_reshape_zero_dim` and
`test_reshape_zero_and_negative_dim`, both `allowzero=0`, are. So the `allowzero=1` path looks
untested in the ONNX node suite rather than deliberately unsupported.

## Version

tract-onnx 0.23.4, macOS ARM64, opset 22; also reproduced on `main` at 159d3cc0f, where the
relevant code is unchanged. Found by differential testing of single-node models
against `onnx.reference`, ONNX Runtime and candle-onnx. Ranks 2, 3 and 4 behave the same way, at
every element type tried.

keevin-xu:
fixes #2698

### Description

`Reshape` never read the `allowzero` attribute, so a `0` in the target shape was always replaced by
the input dimension at that position. That is the `allowzero=0` rule; with the attribute set, ONNX
gives a `0` its literal meaning of a zero-length dimension, "not taken from input tensor".

The registration discarded the node, and `Reshape` was a unit struct with nowhere to carry the flag:

    reg.insert("Reshape", |_, _| Ok((expand(array::Reshape::default()), vec![])));

`compute_shape_with_tf_rules` and `to_axis_ops_with_tf_rules` now delegate to `allowzero`-aware
versions and keep their signatures, so the TensorFlow path is unchanged. `Reshape` carries the flag
through `rules` and `wire`, and a `0` alongside a `-1` is rejected, which ONNX calls invalid.

Honouring the attribute is not sufficient on its own. `to_axis_ops_*` decomposes a reshape into a
sequence of primitive axis operations, pairing contiguous runs of the input and output shapes whose
volumes are equal. A zero-length dimension defeats that test, because every run containing it has
volume 0. For `[0,3,4]` to `[3,4,0]` it paired the input run `[0]` with the output run `[3,4,0]`,
both volume 0, producing `3,4,0,3,4` and then failing with
`Removing non-trivial axis #4 of dim: 3,4,0,3,4`.

A tensor with a zero-length dimension holds no elements, so there is no data movement to sequence,
only a change of shape metadata. The decomposition is therefore skipped in that case:

    if final_output.iter().chain(input_orig.iter()).any(|d| *d == 0.into()) {
        return Ok(if *input_orig == *final_output {
            tvec!()
        } else {
            tvec!(AxisOp::Reshape(0, input_orig.into(), final_output.clone()))
        });
    }

Without this second part, `test_reshape_allowzero_reordered` still fails after the attribute is
honoured.

### Motivation + Context

Two behaviours change: models that were wrongly rejected now load: `float32[3,0]` reshaped to `[0]`
with `allowzero=1` returns `[0]` instead of failing analysis with a volume mismatch. And models that
ONNX calls invalid are now rejected instead of silently mis-computed: `float32[2,3]` to `[0,3]` and
to `[0,-1]`, both with `allowzero=1`, previously came back as `[2,3]` full of data. `onnx.reference`
and ONNX Runtime agree with the new behaviour in every case.

This enables ONNX's own `test_reshape_allowzero_reordered`, which was absent from
`test-rt/suite-onnx/node.txt` while the two `allowzero=0` zero-dim cases were present, so the path
looks untested rather than deliberately unsupported. It fails without both changes above and passes
in all four suite variants with them. Unit tests cover the literal zero, the reordered case, the
`0`-with-`-1` rejection and the unchanged copy rule.

Verified on macOS ARM64: full ONNX conformance suite 4532 passed / 0 failed; tract-core 285,
tract-hir 67, tract-onnx 15 all green; tract-tensorflow compiles. Separately checked 17 single-node
models by hand against `onnx.reference` and ONNX Runtime, comparing shapes and values: all 17 agree,
where 8 disagreed before.

### DIFF
--- a/core/src/ops/change_axes.rs
+++ b/core/src/ops/change_axes.rs
@@ -1039,20 +1039,47 @@ pub fn perm_to_ops(input: &[usize]) -> TVec<AxisOp> {
 }
 
 pub fn compute_shape_with_tf_rules(input: &[TDim], shape_spec: &[TDim]) -> TractResult<TVec<TDim>> {
+    compute_shape_with_onnx_rules(input, shape_spec, false)
+}
+
+/// Resolve a `Reshape` target shape, honouring ONNX's `allowzero`.
+///
+/// With `allowzero` clear, a 0 in the target copies the input dimension at the same position.
+/// This is the TensorFlow rule and the ONNX default, and is what `compute_shape_with_tf_rules`
+/// asks for. With `allowzero` set, ONNX gives a 0 its literal meaning: a zero-length dimension,
+/// "not taken from input tensor". `allowzero` was added in ONNX opset 14.
+pub fn compute_shape_with_onnx_rules(
+    input: &[TDim],
+    shape_spec: &[TDim],
+    allowzero: bool,
+) -> TractResult<TVec<TDim>> {
     let mut shape: TVec<TDim> = shape_spec.into();
-    // Replace 0s with corresponding input dims (positional, per ONNX/TF spec)
-    for (i, s) in shape.iter_mut().enumerate() {
-        if *s == 0.into() {
-            *s = input
-                .get(i)
-                .with_context(|| {
-                    format!("Reshape: 0 at position {i} but input only has {} dims", input.len())
-                })?
-                .clone();
+    if !allowzero {
+        // Replace 0s with corresponding input dims (positional, per TF and ONNX allowzero=0)
+        for (i, s) in shape.iter_mut().enumerate() {
+            if *s == 0.into() {
+                *s = input
+                    .get(i)
+                    .with_context(|| {
+                        format!(
+                            "Reshape: 0 at position {i} but input only has {} dims",
+                            input.len()
+                        )
+                    })?
+                    .clone();
+            }
         }
     }
     let input_vol: TDim = input.iter().product();
     if let Some(pos) = shape.iter().position(|d| *d == (-1).into()) {
+        // ONNX: "If the attribute 'allowzero' is set, it is invalid for the specified shape to
+        // contain both a zero value and -1, as the value of the dimension corresponding to -1
+        // cannot be determined uniquely."
+        if allowzero && shape.iter().any(|d| *d == 0.into()) {
+            bail!(
+                "Reshape: allowzero is set, so the target shape {shape_spec:?} may not contain both 0 and -1"
+            )
+        }
         let shape_vol: TDim = shape.iter().filter(|d| **d != (-1).into()).product();
         let div = input_vol.maybe_div(&shape_vol)?;
         if div.1 != 1 {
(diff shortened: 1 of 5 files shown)


kali (approved):

kali:
Looks pretty good, thanks !

PR #2708, opened 2026-08-23
onnx: reject Reshape target shapes that inference cannot resolve

### LINKED ISSUE #2707: Reshape hangs or panics on a target shape whose negative dimension is never resolved
`Reshape` can return a shape that still contains a negative dimension. When it does, tract either 
spins indefinitely or panics, instead of rejecting the model cleanly. `onnx.checker` accepts
every model below, and `onnx.reference` and ONNX Runtime both reject them promptly.

## What ONNX specifies

From the [`Reshape` reference](https://onnx.ai/onnx/operators/onnx__Reshape.html):

> "At most one dimension of the new shape can be -1. In this case, the value is inferred from the
> size of the tensor and the remaining dimensions."

`-1` is the only negative the specification gives a meaning to, so any other negative left in a
resolved shape is not an unhandled sentinel but an invalid dimension.

## Reproduction

| input | target | onnx.reference 1.22.0 | onnxruntime 1.29.0 | tract `38b383fbd` |
|---|---|---|---|---|
| `f32[2,3,4]` | `[-1]` | `[24]` | `[24]` | `[24]`, correct |
| `f32[2,3,4]` | `[-1,-1]` | rejects | rejects | **hangs**, 100% CPU |
| `f32[2,3,4]` | `[-1,-1,-1]` | rejects | rejects | **hangs**, 100% CPU |
| `f32[2,3,4]` | `[2,-1,-1]` | rejects | rejects | **hangs**, 100% CPU |
| `f32[2,3,4]` | `[-1,-2]` | rejects | rejects | **panics** |
| `f32[2,3,4]` | `[-2,-1]` | rejects | rejects | **panics** |
| `f32[3,0]` | `[-2,0]` | rejects | rejects | **panics** |

Not affected and deliberately outside this report: `[-2]`, `[-3]`, `[2,-2]`, `[-2,-2]` and
`[-2,-3]` on `f32[2,3,4]` all give a clean volume-mismatch error today.

Hangs were seen at 8s, 35s and separately 6 minutes, with flat memory. The panic is:

```
thread 'main' panicked at core/src/model/fact.rs:21:
assertion failed: self.dims.iter().all(|d| d.as_i64().map(|d| d >= 0).unwrap_or(true))
```

A plain `assert!` rather than `debug_assert!`, so the panic is present in release builds too. Both failures happen in
`into_typed()`; `model_for_path` succeeds.

### Steps to reproduce

Needs `pip install onnx`. **Note: run the tract half under a timeout**, since some do not return.

```python
import numpy as np, onnx
from onnx import TensorProto as T, helper as h, numpy_helper as nh

for name, ishape, target, out in [("two_neg1",  [2, 3, 4], [-1, -1], [24, 1]),
                                  ("neg1_neg2", [2, 3, 4], [-1, -2], [12, 2]),
                                  ("zerovol",   [3, 0],    [-2,  0], [1, 0])]:
    shape = nh.from_array(np.array(target, np.int64), "shape")
    g = h.make_graph([h.make_node("Reshape", ["a", "shape"], ["y"])], "reshape",
                     [h.make_tensor_value_info("a", T.FLOAT, ishape)],
                     [h.make_tensor_value_info("y", T.FLOAT, out)], [shape])
    m = h.make_model(g, opset_imports=[h.make_opsetid("", 22)]); m.ir_version = 10
    onnx.checker.check_model(m)          # all three accepted
    onnx.save(m, f"{name}.onnx")
```

```rust
// two_neg1.onnx  -> does not return
// neg1_neg2.onnx -> panics in ShapeFact::compute_concrete
let _ = tract_onnx::onnx().model_for_path("two_neg1.onnx")?.into_typed()?;
```

## Where it happens

`core/src/ops/change_axes.rs`, in the shape computation:

```rust
if let Some(pos) = shape.iter().position(|d| *d == (-1).into()) {
    let shape_vol: TDim = shape.iter().filter(|d| **d != (-1).into()).product();
    ...
    shape[pos] = div.0;
} else {
    let shape_vol: TDim = shape.iter().product();
    if input_vol != shape_vol { bail!("Reshape volume mismatch: ...") }
}
```

The returned shape is never checked for negative dimensions, and both branches can let one through.
`position` resolves only the **first** `-1` while `filter` removes **every** `-1` from the product,
so a second negative survives a branch that performs no volume check: `[2,3,4]` to `[-1,-1]` gives
`[24, -1]`. And the volume check in the other branch compares products, which a zero-length
dimension makes equal regardless of sign, so `[3,0]` to `[-2,0]` passes it.

In the reproduced multiple-`-1` cases, the leftover `-1` sends `to_axis_ops_*` into an `Add(1)/Rm(1)` oscillation that never converges;
 any other leftover negative reaches `ShapeFact` and trips the assertion above.

## Version

tract main at `38b383fbd`, macOS ARM64, opset 22. Independent of #2699: reproduces with no
attribute set, confirmed both before that fix and after it merged. Found by differential testing of
single-node models against `onnx.reference` and ONNX Runtime.

keevin-xu:
fixes #2707

Reshape could return a shape still holding a negative dimension, which then either hangs the axis-op search or triggers the ShapeFact assertion. 

Two commits: reject a negative that survives inference, and reject a -1 whose remaining volume is zero, since it is undetermined rather than wrong. 🍍

### DIFF
--- a/core/src/ops/change_axes.rs
+++ b/core/src/ops/change_axes.rs
@@ -1094,6 +1094,14 @@ pub fn compute_shape_with_onnx_rules(
             );
         }
     }
+    // Neither branch above rules a negative out: -1 resolves only the first, and comparing
+    // volumes accepts one when a dimension is zero, as -2 * 0 == 3 * 0.
+    if let Some(bad) = shape.iter().position(|d| d.as_i64().map(|d| d < 0).unwrap_or(false)) {
+        bail!(
+            "Reshape: dimension {bad} of target shape {shape_spec:?} is still negative after \
+             inference (resolved to {shape:?}); at most one dimension may be -1"
+        )
+    }
     Ok(shape)
 }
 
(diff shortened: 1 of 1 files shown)


kali (changes requested):
Please make sure your code bot reads the instructions in CLAUDE.md (even if it's not claude, they're generic). Specifically about the role and purpose of comments on this project.

keevin-xu:
Just made changes. I shortened the in-file comments and the commit messages.

Sorry about this!

kali:
> Just made changes. I shortened the in-file comments and the commit messages.
> 
> Sorry about this!

No problem :) we are learning how to tame these things.

kali (approved):

Issue #2629, opened 2026-08-12
Fusing a GRU Scan into one op: what should happen to tract-managed cross-call state?


czoli1976:
@kali — a design question before I write any more code, following the GRU fusion PRs
(#2626, #2628). It is about cross-call Scan state, so it touches the same ground as
#2157 and #2605.

## What I measured

Two streaming speech-enhancement models, GTCRN and UL-UNAS, both built from GRUs that
scan the **frequency** axis (9, 17 or 33 subbands) rather than time. Feeding one frame
of identical input repeatedly, without changing anything between calls:

| | drift in output norm over 4 identical calls |
|---|---|
| onnxruntime | 0.0 — identical every call |
| tract | GTCRN 6.5e-01, UL-UNAS 6.9e+00 |

Both models pass `initial_h` explicitly on every GRU node, and neither exposes the
corresponding `Y_h` as a model output. So `declutter_single_loop`'s reachability check
classifies the state as tract-managed and it is carried across calls, which is what the
`external_state` doc describes and what #2157 is about. The models declare their own
state as explicit cache inputs and outputs, so from the caller's side there is hidden
state they cannot see or reset.

I am **not** filing this as a bug — the behaviour is documented and deliberate, and for a
time-axis recurrence it is clearly the right default. My question is what it should mean
for a fused op.

## The design question

Profiling says the remaining gap to onnxruntime on these models is dispatch count:
onnxruntime runs one `GRU` kernel per layer looping every timestep internally (28 calls
on UL-UNAS) where tract runs 524 scan iterations. Collapsing the body helped —
GTCRN 13.8 ms to 1.63 ms and UL-UNAS 7.3 ms to 2.49 ms across #2621 to #2628, against
onnxruntime's 1.29 and 1.53 — but the per-iteration dispatch remains, and only replacing
the Scan removes it.

That is where the state question bites. A whole-sequence op that takes `initial_h` as a
plain input is stateless, which matches the ONNX contract and onnxruntime, but drops the
carry tract currently provides — on exactly the models the fusion is for. Guarding the
fusion on "no state carried across calls" would mean it never fires here.

So, three questions:

1. For a GRU scanning a non-time axis, is carrying state across calls intended, or an
   artifact of the reachability check not being able to tell the two cases apart? For
   these models the state at the end of a frame is the last subband's, which does not
   obviously mean anything on the next frame.

2. If a fused whole-sequence op is welcome, should it carry state itself — an `OpState`
[...]

PR #2717, opened 2026-08-24
(perf) core,onnx: GruSeq: run a GRU as one op instead of a per-timestep scan

czoli1976:
Replaces the `Scan`-lowered GRU with a single whole-sequence op (`GruSeq`) when the GRU is eligible (`linear_before_reset`, sigmoid/tanh gates, hidden <= 64, seq_len >= 4), so a recurrence over a non-time axis stops paying per-timestep dispatch overhead. The recurrent weight is packed once per model run instead of once per Scan iteration; arithmetic is byte-identical to the existing Scan+GruEpilogue path via a shared row-kernel, and cross-call state semantics are unchanged (see sonos/tract#2629 for the open question of whether that's the right long-term contract — this PR doesn't touch it).

🍍


czoli1976:
Pushed a follow-up commit on this branch: **stop reallocating GruSeq's packed state every timestep**.

While profiling CEVA's [DPDFNet](https://github.com/ceva-ip/DPDFNet) (a Dual-Path-RNN
DeepFilterNet2 variant — 32 `GruSeq` nodes per frame) I found that `GruSeq` was
`prepare_one`-ing the hidden state on every step. The buffer's size depends only on
`hidden` and `batch`, so that is a fresh allocation per timestep for a shape that never
changes, plus a transposed temporary to feed it. `tract -O … dump --profile` put `GruSeq`
at 57% of the frame, and bucketing a `sample` profile showed only 42% of leaf time was
actually in the MMM kernels — malloc/free alone was 14.4%.

The commit packs the state once into a buffer refilled in place (new
`PackedFormat::{new_packed_buffer, repack_tensor_view}`), reads it by stride so the
transposed temporary goes away, and stores the recurrent product straight into its final
`[batch, 3*hidden]` layout instead of copying it out of a `[3*hidden, batch]` temp.
It moves allocations and copies only; the arithmetic is untouched.

### Accuracy

Dumping every output frame and comparing whole streams: this branch is **bit-identical
to the `Scan` it replaces** — max abs difference exactly `0` on every model, both for
`GruSeq` alone and with this commit. Against onnxruntime the two agree to f32 rounding.

| model | A (no GruSeq) vs C | C vs ORT, max abs | vs ORT, rel. to output range |
|---|---|---|---|
| DPDFNet `baseline` | 0 (bit-identical) | 2.5e-07 | 2.1e-06 |
| dpdfnet2 | 0 (bit-identical) | 1.6e-07 | 1.3e-06 |
| dpdfnet8 | 0 (bit-identical) | 1.8e-07 | 2.9e-06 |
| dpdfnet8_48khz_hr | 0 (bit-identical) | 8.9e-07 | 1.7e-06 |
| gtcrn | 0 (bit-identical) | 6.5e-06 | 1.4e-05 |

Cosine similarity to ORT is 1.000000000 (f64 accumulation) in every row.

Profile shift, DPDFNet-8 48 kHz (leaf samples, `/usr/bin/sample`):

| bucket | before | after |
|---|---|---|
| MMM inner kernels | 42.7% | 41.6% |
| malloc / free | 14.4% | **8.0%** |
| memmove / memset | 8.8% | 7.0% |
| packing | 5.2% | 4.3% |
| ndarray copies | 4.1% | **0.6%** |

### Speed, this branch's two steps side by side

p50 ms/frame, median of 4 interleaved runs, single-threaded, M-series. **A** = this
branch's parent (GRU as a `Scan`), **B** = `GruSeq` as it stood before this commit,
**C** = with it.

| model | A: no GruSeq | B: GruSeq | C: B + this | A→B | B→C | **A→C** |
|---|---|---|---|---|---|---|
| dpdfnet2 | 1.383 | 1.180 | 1.139 | +14.7% | +3.4% | **+17.6%** |
| dpdfnet4 | 2.231 | 1.885 | 1.788 | +15.5% | +5.2% | **+19.9%** |
| dpdfnet8 | 4.082 | 3.330 | 3.095 | +18.4% | +7.1% | **+24.2%** |
| dpdfnet2_8khz | 1.374 | 1.210 | 1.153 | +11.9% | +4.7% | **+16.0%** |
| dpdfnet8_8khz | 3.897 | 3.221 | 3.019 | +17.4% | +6.2% | **+22.5%** |
| dpdfnet2_48khz_hr | 2.566 | 2.158 | 1.998 | +15.9% | +7.4% | **+22.1%** |
| dpdfnet8_48khz_hr | 6.841 | 5.118 | 4.735 | +25.2% | +7.5% | **+30.8%** |
| gtcrn | 0.670 | 0.460 | 0.411 | +31.4% | +10.6% | **+38.6%** |
| gtcrn_simple | 0.670 | 0.460 | 0.410 | +31.3% | +10.9% | **+38.8%** |
| DPDFNet `baseline` (no GRU — control) | 0.515 | 0.523 | 0.527 | −1.6% | −0.6% | −2.2% |

The `baseline` row is DPDFNet's GRU-free ablation and is flat, as it should be.

### Against onnxruntime

Same models, ORT and all three tract builds measured **interleaved in one process** so
they see the same machine state. ORT 1.x, `CPUExecutionProvider`,
`intra_op_num_threads=1`, `inter_op_num_threads=1`, `ORT_ENABLE_ALL`; both engines run
one 10 ms hop per call with state threaded between calls.

| model | ORT | A: no GruSeq | B: GruSeq | C: B + this | A/ORT | B/ORT | **C/ORT** |
|---|---|---|---|---|---|---|---|
| dpdfnet2 | 1.011 | 1.354 | 1.173 | 1.130 | 1.34× | 1.16× | **1.12×** |
| dpdfnet4 | 1.631 | 2.245 | 1.879 | 1.777 | 1.38× | 1.15× | **1.09×** |
| dpdfnet8 | 2.931 | 4.024 | 3.314 | 3.084 | 1.37× | 1.13× | **1.05×** |
| dpdfnet2_8khz | 0.995 | 1.362 | 1.209 | 1.155 | 1.37× | 1.22× | **1.16×** |
| dpdfnet8_8khz | 2.494 | 3.470 | 2.920 | 2.741 | 1.39× | 1.17× | **1.10×** |
| dpdfnet2_48khz_hr | 1.436 | 2.417 | 1.993 | 1.908 | 1.68× | 1.39× | **1.33×** |
| dpdfnet8_48khz_hr | 3.902 | 6.780 | 5.092 | 4.714 | 1.74× | 1.30× | **1.21×** |
| gtcrn | 0.619 | 0.676 | 0.464 | 0.414 | 1.09× | 0.75× | **0.67×** |
| gtcrn_simple | 0.616 | 0.674 | 0.464 | 0.412 | 1.09× | 0.75× | **0.67×** |
| DPDFNet `baseline` (no GRU — control) | 0.405 | 0.466 | 0.466 | 0.466 | 1.15× | 1.15× | 1.15× |

On GTCRN this branch turns a loss into a win: 1.09× slower than ORT before `GruSeq`,
**1.49× faster** after both commits. On DPDFNet tract is still behind, but the mean over
the seven GRU models goes 1.47× → 1.22× → 1.15×, and dpdfnet8 lands at 1.05×.

The `baseline` control is pinned at 1.15× in all three columns — it has no GRU, so
nothing here touches it. That is roughly the floor this model family currently costs
tract for non-GRU reasons, which is why the DPDFNet rows converge toward it rather than
past it: there is not much GRU-specific headroom left in them. The two 48 kHz models are
the exception and still the worst rows, and they also started worst.

Numbers are one machine; the DPDFNet rows and the GTCRN rows come from two separate
interleaved runs (ORT and all three tract builds interleaved within each), so compare
ratios down a column rather than absolute ms across the two groups. Treat them as
indicative rather than exact.

Full workspace suite green (15 suites, 0 failures), `cargo fmt` and `cargo clippy` clean.
Adds no `unsafe` outside linalg — it removes one `unsafe` from core.

🍍

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali (dismissed):

czoli1976:
@kali should I look at failures or hold a bit?

kali:
let me deal with them.

czoli1976:
on it .....

kali:
ok :)

czoli1976:
> ok :)

not the most patient person

czoli1976:
@kali go again please

kali (dismissed):

czoli1976:
@kali can you finish fixing the CI now?

kali:
> @kali can you finish fixing the CI now?

These ones look PR induced to me ?

PR #2725, opened 2026-08-25
(fix)core/cnn: fuse ONNX QDQ Conv into int8 GEMM instead of running it in float

czoli1976:
Found this while trying to replicate [faster-enhancer.c](https://github.com/kdrkdrkdr/faster-enhancer.c)'s W8A8 int8 speedup for FastEnhancer. That attempt didn't pay off on this hardware — no AMX-int8 path on Apple Silicon means int8 doesn't beat tract's existing f32 GEMM here regardless of how correctly it's wired — but along the way I found tract was silently running a QDQ-quantized `Conv` in plain float the whole time, paying the Quantize/Dequantize tax for nothing. Worth fixing on its own even though it didn't get me the 2x I was chasing.

## What was wrong

`DequantizeLinear -> Conv -> QuantizeLinear` (the standard `onnxruntime.quantization.quantize_static` QDQ output) never got recognized as quantizable: the generic `DequantizeLinear::declutter()` chain-walk relies on `Op::quantize()`, which `Conv` doesn't implement — and that mechanism only ever wires a single input per op anyway, so it couldn't fit `Conv`'s 3 inputs (x/kernel/bias) regardless. The middle `Conv` ran in plain float while still paying the Quantize/Dequantize cast at every layer boundary — strictly worse than not quantizing at all.

## The fix

Added a `Conv`-specific declutter rule that matches the sandwich directly, re-quantizes the (already constant-folded, plain float) weight and bias into the `q_params` layout `wire_as_quant_im2col` already expects for `ConvInteger`/`QLinearConv`, and fuses it into a single quantized `Conv`.

One real bug found along the way: `QuantizeLinearU8`/`I8`'s own `scale` field is `1/y_scale` (the ONNX importer inverts it to match this op's multiply-based eval) — the opposite convention from `DequantizeLinear`'s `scale`. Missing that inversion silently saturated every output to the zero point.

## Verification

Isolated single-conv unit test: fused output matches the unfused float graph within 1e-2 (cos_sim 0.9988 on a broader synthetic check), and the requantization multiplier matches a hand-derived reference to 8 significant figures. Weight and bias requantization are bit-exact matches to onnxruntime's own calibrated values on a real model.

I'll be honest about where this leaves the model I actually cared about: on the full FastEnhancer pipeline (12 quantized convs + GRU/attention), accuracy is still short of production quality with the calibration set I had on hand (cos_sim ~0.6 vs ~0.98 for dynamic quantization on the same input) — that looks like a calibration-difficulty problem, not a plumbing bug, since the isolated test is exact. And even correctly quantized, this hardware doesn't have a fast int8 path, so the fused conv lands close to f32 parity rather than faster, both natively and on wasm+relaxed-simd. faster-enhancer.c's own reported PESQ (3.054 vs 3.060 fp32) suggests it's much more forgiving of quantization than this pipeline turned out to be, for reasons I don't have a full explanation for yet.

Filing regardless because the bug itself is real and hardware-independent: any model exported via `quantize_static`'s QDQ format currently gets silently mis-optimized on any target, not just this one.

## Testing

`cargo test --release --workspace` (macOS, minus `test-tflite`/`tract-proxy`/`tract-proxy-sys`/`test-cuda`): same failure set as a clean `origin/main` baseline (gelu-f16 doctest fence-annotation issue, unrelated tensorflow/prost doctests, one known MFA probe-test failure already tracked in #2546). Re-verified the tract-metal failures in isolation (109 passed, 1 failed — the tracked MFA issue) since the full-suite run showed transient extra failures from concurrent GPU contention on my end.

🍍

### DIFF
--- a/core/src/ops/cnn/conv/conv.rs
+++ b/core/src/ops/cnn/conv/conv.rs
@@ -895,6 +895,82 @@ impl Conv {
         Ok(Some(patch))
     }
 
+    /// Fuses a QDQ-style `DequantizeLinear(x) -> Conv(x, w, bias) -> QuantizeLinear(y)`
+    /// sandwich into a single `q_params`-set `Conv`, reusing `wire_as_quant_im2col`'s
+    /// int8 GEMM path instead of running the middle `Conv` in float.
+    ///
+    /// `w`/`bias` arrive as plain float32 constants (any weight/bias-side
+    /// `DequantizeLinear` is already constant-folded away by the time this runs), so
+    /// they are independently re-quantized here: `w` per-tensor symmetric int8 from
+    /// its own value range, `bias` to int32 using the standard `x_scale * w_scale`
+    /// convention `wire_as_quant_im2col` assumes for its accumulator.
+    fn declutter_qdq_fuse(
+        &self,
+        model: &TypedModel,
+        node: &TypedNode,
+    ) -> TractResult<Option<TypedModelPatch>> {
+        rule_if!(self.q_params.is_none());
+        let prec = model.node(node.inputs[0].node);
+        rule_if_some!(dq = prec.op_as::<ops::quant::DequantizeLinearF32>());
+        rule_if_let!(&[succ_outlet] = &*node.outputs[0].successors);
+        let succ = model.node(succ_outlet.node);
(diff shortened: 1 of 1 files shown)


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2728, opened 2026-08-26
(perf) core: fuse a squeezed unit matmul axis and a unicast Mul+Add pair

czoli1976:
Three changes that cut copies and node count in the optimised graph. All are
pure plumbing — outputs are **bit-identical** on every model tested.

### `absorb_squeeze` refused to drop a unit m or n axis

`OptMatMul::absorb_squeeze` bailed on any reshape that removed the m or n axis.
But a pure squeeze only ever removes unit axes, and the store already handles a
`None` axis: `compute_strides` reads a zero stride for it and the extent falls
back to 1, which is exactly what a squeezed unit axis means. So the guard was
turning down reshapes it could have absorbed.

It costs two nodes, not one. A `[1, m]` GEMV output reshaped to `[m]` kept its
`IntoShape`, and that node then sat between the matmul and its bias add,
blocking the `OptBinUnicast(Add)` → `AddUnicast`-in-the-store fusion that would
otherwise have fired.

The guard is dropped, both axes now go through a `squeeze_c_axis` helper, and
`rm_c_axis` tracks the change into the fused specs' own store specs — including
`AddUnicast`'s, which previously was not updated at all (unreachable while the
guard stood).

### A unicast `Mul` feeding a unicast `Add` walked the accumulator twice

`x * g` then `+ b` reads and writes the whole tensor twice for arithmetic that
fits in one traversal. `OptBinUnicast::fuse` now folds the pair into
`OptMulAddUnicast` when both unicast operands share a period, so a single walk
covers them. It is codegen-only, so nothing new has to serialise.

This is the affine tail of a normalisation and the gate update of a GRU cell —
it fires on `LayerNormalization.y_internal` and `grucell/Add_2` among others.

### `MoveAxis` copied the tensor twice

`AxisOp::change_tensor`'s `Move` arm did `tensor.clone().move_axis(..)`. But
`move_axis` consumes its input and goes through `permute_axes` → `Tensor::from_datum`,
which allocates a fresh natural-strided tensor and copies into it. The clone was a
second, redundant full copy on every `MoveAxis` node.

Unlike the two rules above this one is not shape-specific — it applies to any
transpose in any model, which is why `dtln` and the DeepFilterNet3 submodels move
at all here.

### Measured

p50 ms/frame, median of 5 interleaved runs, single-threaded, aarch64, against
this PR's merge base.

| model | main | this PR | |
|---|---|---|---|
| dpdfnet2_8khz | 1.449 | 1.380 | **+4.8%** |
| dpdfnet2 | 1.329 | 1.277 | **+4.0%** |
| baseline (DPDFNet, no GRU) | 0.464 | 0.446 | **+3.8%** |
| dpdfnet8 | 3.940 | 3.794 | **+3.7%** |
| dpdfnet4 | 2.198 | 2.130 | **+3.1%** |
| dpdfnet8_8khz | 3.791 | 3.684 | **+2.8%** |
| dpdfnet8_48khz_hr | 6.695 | 6.516 | **+2.7%** |
| dpdfnet2_48khz_hr | 2.555 | 2.497 | **+2.3%** |
| gtcrn | 0.683 | 0.680 | +0.5% |
| DeepFilterNet3 enc | 0.1263 | 0.1258 | +0.4% |
| dtln_model_2 | 0.0529 | 0.0527 | +0.4% |
| dtln_model_1 | 0.0312 | 0.0312 | 0.0% |
| DeepFilterNet3 erb_dec | 0.1057 | 0.1057 | 0.0% |
| DeepFilterNet3 df_dec | 0.0984 | 0.0991 | −0.7% |
| gtcrn_simple | 0.685 | 0.696 | −1.6% |

The last two rows are noise, not regressions: `gtcrn_simple` is the same graph as
`gtcrn` (+0.5%) and `df_dec` runs in 98 µs, where a 0.7 µs swing is a rounding
artifact of the median.

The wins concentrate on [DPDFNet](https://github.com/ceva-ip/DPDFNet), whose
Gemm-based GRU cells emit exactly the `[1, m]`-GEMV-then-reshape-then-bias chain
the first rule unblocks. On `dpdfnet8_48khz_hr` that is 1901 nodes down to 1103,
with `OptAddUnicast` 148 → 50 and `IntoShape` 184 → 154. GTCRN and DTLN do not
form the chain and land inside noise, as expected.

Outputs compared whole-stream, every frame: **max absolute difference exactly
`0`** on all nine models above, including the two that gain nothing.

Full workspace suite green, `cargo fmt` and `cargo clippy` clean. No new
`unsafe`.

🍍

### DIFF
--- a/core/src/ops/binary.rs
+++ b/core/src/ops/binary.rs
@@ -909,7 +909,117 @@ impl EvalOp for OptBinUnicast {
     }
 }
 
+/// `a * b + c` in one pass, where `b` and `c` are unicast operands sharing a
+/// period: each has leading unit axes that `a` repeats over, and matches `a`
+/// past them. Formed by [`OptBinUnicast::fuse`] from a `Mul` feeding an `Add`,
+/// which would otherwise walk `a` twice.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct OptMulAddUnicast;
+
+impl Op for OptMulAddUnicast {
+    fn name(&self) -> StaticName {
+        "OptMulAddUnicast".into()
+    }
+    op_as_typed_op!();
+}
+
+impl OptMulAddUnicast {
+    fn eval_t<T>(a: &mut Tensor, b: &Tensor, c: &Tensor) -> TractResult<()>
+    where
+        T: Datum + Copy + std::ops::Mul<Output = T> + std::ops::Add<Output = T>,
+    {
+        let b = b.try_as_plain()?;
+        let b = b.as_slice::<T>()?;
+        let c = c.try_as_plain()?;
+        let c = c.as_slice::<T>()?;
+        let mut a = a.try_as_plain_mut()?;
+        for chunk in a.as_slice_mut::<T>()?.chunks_mut(b.len()) {
+            for ((v, b), c) in chunk.iter_mut().zip(b).zip(c) {
+                *v = *v * *b + *c;
+            }
+        }
+        Ok(())
+    }
+
+    fn natural(t: &Tensor) -> bool {
+        t.len() == t.shape().iter().product::<usize>()
+            && t.strides() == &*Tensor::natural_strides(t.shape())
+    }
+}
+
+impl EvalOp for OptMulAddUnicast {
+    fn is_stateless(&self) -> bool {
+        true
+    }
+
+    fn eval(&self, inputs: TVec<TValue>) -> TractResult<TVec<TValue>> {
+        let (a, b, c) = args_3!(inputs);
+        // Same natural-strides contract as OptBinUnicast: the chunked walk sizes
+        // its slices from the declared shape, which only matches the storage
+        // when the strides are C-order.
+        if !Self::natural(&a) || !Self::natural(&b) || !Self::natural(&c) {
+            let dt = a.datum_type();
+            let m = Mul.eval(a, b, dt)?;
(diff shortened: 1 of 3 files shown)


czoli1976:
Not bad at all, @kali

kali (dismissed):

czoli1976:
@kali , sorry, folding in an improvement

czoli1976:
Added a third commit, `core: drop a redundant copy in MoveAxis`.

`AxisOp::change_tensor`'s `Move` arm did `tensor.clone().move_axis(..)`, but
`move_axis` consumes its input and goes through `permute_axes` →
`Tensor::from_datum`, which allocates a fresh natural-strided tensor and copies
into it. So the clone was a second full copy of the data on every `MoveAxis`,
83 of them per frame on the model I was profiling.

It is smaller than it looks: the clone is a flat memcpy while the copy that
remains is the strided one, so `MoveAxis` goes 0.256 → 0.219 ms/i rather than
halving. But unlike the other two commits it is not shape-specific — it applies
to any transpose, which is why the DTLN and DeepFilterNet3 rows move at all in
the table.

The perf table in the description is re-measured with all three commits in.

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

czoli1976:
@copilot review this PR

Issue #2724, opened 2026-08-25
ONNX import: fractional literals in dim_param are parsed with integer division, silently changing the shape expression (blocks all paddle2onnx models)


TarikAlHadethi:
## Summary

`parse_tdim` evaluates a fractional literal such as `1/2` using integer division, yielding `0`. Because paddle2onnx emits dimension names of the form `floor(X/2 - 1/2) + 1`, where the `1/2` is a rational constant and the `floor()` exists precisely because the arithmetic is real-valued, tract silently reduces that expression to `X/2 + 1`.

That is off by one for every even `X`. The mis-parsed shape then contradicts tract's own (correct) convolution shape rule, and the model cannot be loaded.

No error or warning is produced at parse time. The expression is simply not the one written in the file.

## Reproduction, without any model file

```rust
use tract_onnx::prelude::*;

fn main() {
    let s = SymbolScope::default();
    println!("{}", s.parse_tdim("1/2").unwrap());                 // 0
    println!("{}", s.parse_tdim("W/2 - 1/2").unwrap());           // (W)/2
    println!("{}", s.parse_tdim("floor(W/2 - 1/2) + 1").unwrap()); // 1+(W)/2
    println!("{}", s.parse_tdim("(W-1)/2 + 1").unwrap());          // (W+1)/2   <- correct
}
```

Evaluated:

| W | `floor(W/2 - 1/2) + 1` as written | tract's parse `1 + W/2` |
|---|---|---|
| 4 | 2 | 3 |
| 5 | 3 | 3 |
| 6 | 3 | 4 |
| 320 | 160 | 161 |
| 321 | 161 | 161 |

## Reproduction with a model

`min_with_vi.onnx` (763 bytes, attached) is a single `Conv`, kernel 3x3, stride 2, pads [1,1,1,1], input `[1, 3, 48, DynamicDimension.1]`, with one `value_info` entry declaring the output width as `floor(DynamicDimension.1/2 - 1/2) + 1`. That is exactly what paddle2onnx writes.

```rust
tract_onnx::onnx().model_for_path("min_with_vi.onnx")?.into_optimized()?;
```

```
Failed analyse for node #2 "Conv.0" ConvHir

Caused by:
    0: Infering facts
    1: Applying rule outputs[0].shape == 1,4,24,(DynamicDimension.1+1)/2
    2: Unifying shapes 1,4,24,1+(DynamicDimension.1)/2 and 1,4,24,(DynamicDimension.1+1)/2
[...]

PR #2730, opened 2026-08-26
onnx: let the facade pass loader options

TarikAlHadethi:
Adds `load_with_options` and `load_buffer_with_options` to `OnnxInterface`, so a
caller going through the `tract` facade can reach `with_ignore_value_info` and set
assertions on the model symbols. `load` and `load_buffer` keep their signatures and
delegate to the new methods, per your note about the facade being the semver
commitment. Closes the API half of #2724.

Options are a `HashMap<String, String>`, lowered to C as two parallel arrays of
null-terminated strings, so the FFI and Python entry points get it too:

* `ONNX_IGNORE_VALUE_INFO` takes `1`, `yes`, `true`, `on` for true and `0`, `no`,
  `false`, `off` for false, in any case.
* `ASSERTIONS` is split on `;` and each piece goes to the assertion parser
  unchanged, so the facade holds no opinion about what an assertion may say.
* An unrecognised key, or a value that does not parse, is an error. That is the
  point of the exercise: #2724 exists because a misread `dim_param` was silent, and
  a stringly-typed option map that ignores a typo would put the same failure one
  layer up.

Both PP-OCR models from the issue now load through the public facade alone, each
checked against the same call without the option so the option is doing the work:

```
recognition model, stock file, nothing stripped
  no options              fails: Impossible to unify Add([Val(1), Div(Sym(DynamicDimension.1), 2)]) with ...
  ONNX_IGNORE_VALUE_INFO  loads

detection model, input 1,3,32*h,32*w
  no options              fails: Impossible to unify MulInt(8, Div(Add([Val(31), MulInt(32, Sym(h))]), 32)) with ...
  ASSERTIONS h>=0; w>=0   loads
```

The detection case is your reparameterisation suggestion from the issue, and it
needs the assertions because the rule that reduces `(32h+31)/32` to `h` is gated on
the symbol being provably non-negative. Neither model needs anything outside the
facade any more.

Seven tests, in `api/tests/mobilenet/mod.rs` so both the native and proxy backends
run the same ones. I checked the unknown-key test can actually fail by making the
parser ignore unknown keys and watching it go red.

Two things worth flagging rather than hiding:

I could not build `tract-proxy` locally, because `tract-proxy-sys` runs bindgen and
there is no libclang on this machine, so the proxy implementation and the two
declarations I added by hand to `api/proxy/sys/tract.h` are unverified here. They
mirror the existing `tract_onnx_load` pattern exactly, and the shared tests will
exercise them on CI. I hand-edited that header rather than regenerating it with
cbindgen, to keep the diff to the two new functions.

The PP-OCR models are 10 MB and 21 MB, so I have not added them as a regression
case. If you want them in the S3 bucket the tests pull from, say the word and I
will send them.

`cargo fmt --all` and `cargo clippy` are clean on the crates I touched. The two
clippy warnings I see are in `tract-linalg` and predate this.

I left `floor()` in the TDim parser alone, as you asked.

🍍

### DIFF
--- a/api/ffi/src/lib.rs
+++ b/api/ffi/src/lib.rs
@@ -2,13 +2,12 @@
 
 use anyhow::{Context, Result};
 use std::cell::RefCell;
-use std::collections::HashMap;
 use std::ffi::{CStr, CString, c_char, c_void};
 use tract::{State, Tensor};
 use tract_api::{
     AsFact, DatumType, DimInterface, FactInterface, InferenceModelInterface, ModelInterface,
-    NnefInterface, OnnxInterface, RunnableInterface, RuntimeInterface, StateInterface,
-    TensorInterface,
+    NnefInterface, OnnxInterface, OnnxOptions, OnnxOptionsSpec, RunnableInterface,
+    RuntimeInterface, StateInterface, TensorInterface,
 };
 
 /// Used as a return type of functions that can encounter errors.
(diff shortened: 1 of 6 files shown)


kali:
I apologize, seeing the PR code, I realize the HashMap was not the right call here. I think we should consider using a json serialization, just as we did for the ModelTransform. I thought the HashMap idiom was already used in the API, but I got rid of it in favor or json serde. This has all the benefits of exposing the OnnxOption structure in the API, make ignore_value_info an actual boolean, the assertions an array of strings, saving us the split on ';'. We can mimick the ModelTransform spec principle, plug in serde json, and bam, done.

I will understand if the back and forth is too much, I can do the changes soon if you don't feel like it.

kali (approved):

kali:
Thank you for your patience. It feels much better.

TarikAlHadethi:
Thanks, and no patience needed. The json version is better, and it made the ffi simpler rather than just tidier: two parallel string arrays plus a length became a single `const char *`.

One thing worth putting on the record, since the bench comment is still sitting there red. I do not think those numbers are from this PR. The bench workflow builds tract-cli and no harness crate depends on the api crates, so this change is not in the path those numbers measure. The same report also carries a 12.7% improvement alongside the regressions, which reads more like variance between two runs than a direction.

I also stopped the loader cloning `Onnx` on every load while I was in there. `with_ignore_value_info` takes `self` by value, so the obvious spelling copies the 182 entry op register whether or not an option is set, and a caller passing nothing now does exactly the work it did before. Measured at 5.6 us against a 19 ms load, so tidiness rather than a fix.

Still cannot build tract-proxy on this machine, no libclang, so the proxy impl and the header edit are unverified locally and I am leaning on CI for that half.

kali:
yeah, don't worry about these red benches, there're not yours. benching tract is expensive and difficult, and there's always some noise.

kali (approved):

TarikAlHadethi:
Thanks for merging.

That is step 1 of the three from #2724 done. Do you want me to take step 2, the separate ONNX shape expression parser that reduces to a single fraction and translates into TDim?

I still have the analysis from the issue, including the case I think decides the design: `floor(floor(DynamicDimension.1/2 - 1/2)/2) + 1`. It is a nested floor rather than a single fraction, two stacked stride-2 convolutions produce it, so it shows up in essentially any real CNN export, and the reduce-to-one-fraction step does not reach it without `floor(floor(x)/n) == floor(x/n)` for integer n > 0 first.

Happy to leave floor() in the TDim parser where you parked it and keep this to the ONNX side.

kali:
Yeah, my understanding is, the extra options unlocked your use-case, but we're not fixing the core issue.

So, yes, I'm happy to look at a PR fixing the onnx/tract semantics mismatch. And kill the TDim ugly floor.

PR #2734, opened 2026-08-27
onnx: parse dim_param as rational, and kill the TDim floor

TarikAlHadethi:
Fixes the remaining two steps of the plan in #2724: the ONNX shape expression parser and simplifier, and removing `floor` from the TDim parser. #2730 only routed around the bug.

## The problem

An ONNX `dim_param` is a rational expression. TDim is integral. The importer handed the string straight to `parse_tdim`, so `1/2` was evaluated with integer division and became `0`:

```
floor(W/2 - 1/2) + 1   read as   W/2 + 1
```

One too big for every even W, no error anywhere. Odd widths agreed, which is why it survived so long.

## What this does

A new `dim_expr` module in tract-onnx keeps the two algebras apart, as you suggested:

1. parse the expression as rational
2. reduce so every `floor` applies to a single fraction
3. drop those floors, because a floored single fraction is exactly what TDim's truncating division computes over non-negative values
4. return a TDim, or an error naming what could not be translated

Step 2 is the part that earns step 3. `floor` of a sum cannot be dropped; `floor` of a single fraction can.

The importer keeps its existing fallback of treating an untranslatable dimension as unknown and letting tract infer it, so the torch dynamo models that relied on that still load. The reason is logged instead of swallowed.

## The nested case

`floor(floor(W/2 - 1/2)/2) + 1` comes from two stacked stride-2 convolutions, so it turns up in essentially any real CNN export rather than being a quirk of the model that started this. It is not a single fraction as written, and reducing the inner floor alone does not make it one. It needs

```
floor(floor(x)/n) == floor(x/n)     for integer n > 0
```

applied first, after which it collapses to `(W+3)/4`. Without that step the result is still correct, since nested truncating divisions agree with nested floors on non-negative values, but it is a tower of divisions that tract then has to unify against the shape it computes itself.

## Removing the TDim floor

Swept for other users first. The only two references were the parser arm and its own test. The `floor` in `resize.rs` is a nearest-neighbour rounding mode and the `floor` in NNEF graphs is the element-wise op, both unrelated. No tract-written NNEF can carry `floor(` in a dim, because TDim's Display never emitted it. Its test is now a negative control: the string is refused rather than silently discarded.

## Testing

20 tests.

The arithmetic ones evaluate the ONNX reading in exact rational arithmetic with real flooring, then require the translated TDim to agree at every width. Agreement is evidence rather than a tautology, since the two are computed by different code over different number systems. Widths include the 4, 5, 6, 318, 319, 320, 321 I checked against the ONNX spec on the real model.

Three of them build a model and run it through the real front end. That is deliberate: this bug shipped for years behind a perfectly good TDim parser, because the wrong parser was the one being called, and testing the translation alone would not have caught that.

`cargo fmt` and `cargo clippy` clean on both crates. tract-data 160 tests and tract-onnx 35 pass. `cargo check --workspace --all-targets` fails here on `core-foundation`, which is macOS only and fails identically on main, so I have not been able to check that target myself.

## What I could not verify locally

The PP-OCR models I have are the copies with `value_info` already stripped, so I cannot use them as an end-to-end fixture. The offer from last time still stands: if you want the two originals in the S3 bucket the tests pull from, say the word and I will send them.

🍍


czoli1976:
@TarikAlHadethi, out of curiosity, which models are you testing? which beckends?

TarikAlHadethi:
PP-OCRv6, the detection and recognition pair, exported with paddle2onnx. That exporter is where the `floor(DynamicDimension.1/2 - 1/2) + 1` dim_params come from, so it is what started #2724 in the first place.

Backend is plain tract on CPU, x86_64 Windows, no GPU. It runs inside Clarus, a Windows screenshot tool I build, which does the OCR locally so nothing has to leave the machine. Detection loads symbolically at `1,3,32*h,32*w` with `h>=0, w>=0` assertions, which is the pair of things #2730 exposed through the facade.

I benchmarked tract against ort and candle on the recognition model a few weeks ago if the numbers are of any use to you. One thing worth knowing regardless: tract runs single-threaded by default, and turning multithreading on was worth about 3.5x on that model, so anyone comparing runtimes out of the box is putting one core up against everyone else's many.

czoli1976:
Thank you, Tarik, mind sharing a link to the models ?

czoli1976:
Thanks — we picked up the official Hub ONNX exports ([collection](https://huggingface.co/collections/PaddlePaddle/pp-ocrv6)) and ran them here on Apple Silicon.

This PR is the import fix; once it loads, **rec is already at or under ORT**, **det is not**. On this Mac, 1 thread, f32, `1×3×48×320` rec / `1×3×640×640` det:

| model | ORT ms | tract ms | tract/ORT |
|---|---:|---:|---:|
| tiny_rec | 6.9 | 7.1 | 1.03× |
| small_rec | 33 | 29 | 0.87× |
| medium_rec | 139 | 58 | 0.42× |
| tiny_det | 77 | 157 | **2.03×** |
| small_det | 181 | 320 | **1.76×** |
| medium_det | 1082 | 1389 | 1.28× |

tiny_det profile: `OptMatMul` 32% at ~25 GF/s, **`DepthWiseConv` 31% at ~1.7 GF/s**, then nearest Resize / `MultiBroadcastTo` / `MaxPool` 2×2 s=1. So the leftover is scalar 2-D depthwise (`3×3`/`5×5`/`7×7`/`9×9`), not the dim_param parse.

AMX vs MLAS will change that ranking on your box. Before we write kernels I need the same 1-thread table from you.

### What to run

PR branch, **one thread**, ORT **CPU EP only** with `intra_op_num_threads=1` (and `inter_op=1`). Models: `PaddlePaddle/PP-OCRv6_{tiny,small,medium}_{det,rec}_onnx` → `inference.onnx`.

```
set RAYON_NUM_THREADS=1

tract -O inference.onnx -i 1,3,48,320,f32 bench --allow-random-input --warmup-loops 8 --max-loops 40
tract -O inference.onnx -i 1,3,640,640,f32 bench --allow-random-input --warmup-loops 3 --max-loops 10
```

Rec shape is `48×320`, det is `640×640`. If analyse still trips on the graph output fact, add `--onnx-ignore-output-shapes` (value_info should be fine on this PR).

ORT side, same shapes, `GraphOptimizationLevel.ORT_ENABLE_ALL`, `providers=['CPUExecutionProvider']` only.

Then the profile dump for **tiny_det** and **tiny_rec** (the “Most time consuming operations” block is enough):

```
tract -O inference.onnx -i 1,3,640,640,f32 dump --profile --cost --allow-random-input
tract -O inference.onnx -i 1,3,48,320,f32 dump --profile --cost --allow-random-input
```

### Please send back

1. CPU (and whether AVX2 / AVX-512)
2. ORT version
3. the 6-row table: ORT p50, tract ms, ratio
4. those two profile summaries
5. the H×W Clarus actually feeds detection, if it is not 640

If you ship with `--threads`, a second table at that count vs ORT `intra_op=N` is useful — but the **1-thread** table is what we need to pick the first kernel PR. No WASM / GPU numbers needed.

The depthwise inner loop we would share is portable (contiguous W, f32, N = 9/25/49 taps). I can do the NEON side here; an AVX2 8-wide `fmadd` on your side is the matching piece, but only after we see whether x86 is DW-bound or GEMM-bound.

TarikAlHadethi:
Numbers below. Short version: on x86 both models are **GEMM-bound, not depthwise-bound**, the ratios are much worse than yours across the board, and the matmul kernels themselves are fine. It is the shapes.

## Machine

- **i7-12700K** (Alder Lake, 8P+4E, 20 threads). **AVX2 + FMA, no AVX-512**, fused off on consumer Alder Lake.
- **onnxruntime 1.29.0**, CPU EP only, `intra_op_num_threads=1`, `inter_op=1`, `ORT_ENABLE_ALL`.
- tract from this PR branch, `RAYON_NUM_THREADS=1`, release.
- Models: the same `PaddlePaddle/PP-OCRv6_{tiny,small,medium}_{det,rec}_onnx` Hub exports.

## The table

Same shapes as yours, rec `1,3,48,320`, det `1,3,640,640`. ORT is p50 over 40 runs (rec) and 10 (det); tract is its own bench mean.

| model | ORT ms | tract ms | tract/ORT | yours (M-series) |
|---|---|---|---|---|
| tiny_rec | 3.9 | 11.8 | **2.99x** | 1.03x |
| small_rec | 20.8 | 47.7 | **2.30x** | 0.87x |
| medium_rec | 101.8 | 294.6 | **2.90x** | 0.42x |
| tiny_det | 45.5 | 304.4 | **6.70x** | 2.03x |
| small_det | 108.9 | 761.6 | **6.99x** | 1.76x |
| medium_det | 746.7 | 4144.4 | **5.55x** | 1.28x |

Loop counts where tract's time budget cut it short: medium_rec 18, small_det 7, medium_det 2. The rest ran the full count.

Worth noting the absolute ORT numbers are all faster than yours (tiny_det 45.5 vs 77, medium_det 747 vs 1082), so this box is not slow. The gap is tract-side.

The recognition row is the one I did not expect. You have tract at or ahead of ORT there; here it is 2.3x to 3x behind.

## Profiles

**tiny_det**, entire network 318.659 ms/i:

```
OptMatMul          68 nodes  115.514 ms/i  36.3%
DepthWiseConv      17 nodes   68.984 ms/i  21.7%
MultiBroadcastTo    7 nodes   37.558 ms/i  11.8%
OptMatMulPack      63 nodes   26.171 ms/i   8.2%
Pad                 5 nodes   15.984 ms/i   5.0%
OptMaxPool          1 nodes   10.746 ms/i   3.4%
OptMulByScalar     57 nodes    9.305 ms/i   2.9%
DeconvSum           2 nodes    7.226 ms/i   2.3%
```

**tiny_rec**:

```
OptMatMul          29 nodes    8.385 ms/i  64.9%
DepthWiseConv      10 nodes    2.220 ms/i  17.2%
Erf                10 nodes    0.678 ms/i   5.2%
OptMatMulPack      27 nodes    0.361 ms/i   2.8%
OptMulByScalar     30 nodes    0.329 ms/i   2.5%
```

## What that says about the kernel question

You asked whether x86 is DW-bound or GEMM-bound before deciding what to write. It is GEMM-bound, in both models:

- **det**: OptMatMul 36.3% plus OptMatMulPack 8.2% is 44.5%, against DepthWiseConv 21.7%. Your Apple split was 32% / 31%.
- **rec**: OptMatMul 64.9% against DepthWiseConv 17.2%.

And the per-node throughput is the part I would look at first. From the det profile:

```
Conv.0.matmatmul   8.005 ms/i   44,236,800 FMA    5.526 GF/s
Conv.1.matmatmul   6.744 ms/i   52,428,800 FMA    7.774 GF/s
Conv.2.matmatmul   8.164 ms/i   52,428,800 FMA    6.422 GF/s
Conv.3.matmatmul   6.802 ms/i  117,964,800 FMA   17.344 GF/s
```

5.5 to 17.3 GF/s on those, against the ~25 GF/s you measured on Apple.

Before reading anything into that: **the AVX2 kernels are working.** I checked, because a build that had quietly fallen back to generic would make this whole table worthless. `medium_rec`'s matmuls reach **61 GF/s** on the same binary, and the peak inside tiny_det itself is **32 GF/s**, so the FMA path is live and fast when the shapes suit it. That is a 6x spread within one model.

So the read is not that the x86 matmul is slow, it is that PP-OCR's detection shapes starve it. `OptMatMulPack` at 63 nodes and 8.2% points the same way, as do `MultiBroadcastTo` at 11.8% and `Pad` at 5.0% for ops that only move memory.

So an AVX2 depthwise kernel would still be worth having, but on x86 it addresses about a fifth of detection rather than a third, and it would leave the larger half untouched. Happy to write it if you still want it, but I would rather ask first whether the matmul path is the better place to spend the effort here, since that is where this box is losing.

## On threads

You offered a second table at `--threads`. Not needed, and the reason is useful: **Clarus does not use tract's internal threading.** It runs tract single-threaded and parallelises in its own code instead, across recognition batches, with a deliberately small batch size so there are more chunks to spread over cores.

So the 1-thread table above is the shipping configuration rather than a lab setting, which I think makes it the right one for picking a kernel anyway.

## Two smaller things

**The flags.** These exports would not analyse with a concrete input: the value_info declares a symbolic batch, so `1,3,640,640` fails with `Impossible to unify Sym(DynamicDimension.0) with Val(1)` at the first Conv. `--onnx-ignore-output-shapes` alone does not clear it, and `--onnx-ignore-value-info` alone then trips on the output fact instead. Both together work. Worth knowing for anyone reproducing this.

**The PR does its job on these files.** Before I added the flags, the analyse dump showed tract carrying `(DynamicDimension.1+1)/2` and `(DynamicDimension.1+3)/4` through the graph, which are exactly the reduced forms of `floor(W/2 - 1/2) + 1` and the stacked-stride-2 `floor(floor(W/2 - 1/2)/2) + 1`. That is the end-to-end confirmation on the official exports that I said in the PR description I could not produce locally, since the copies I had were already stripped.

**And the H x W Clarus actually feeds detection is not 640.** It caps the longest side at 960, keeps the aspect ratio, then pads up to a multiple of 32. A 2560x1440 screenshot goes in at **960x544**, which is 2.25x the pixels of the 640x640 in the table above.

czoli1976:
You’re right it’s GEMM-bound on x86, and the FMA kernels are fine — 61 GF/s on medium_rec is the tell. The four convs you listed are these im2col GEMMs:

```
Conv.0  m=16  k=27   n=102400   5.5 GF/s
Conv.1  m=8   k=64   n=102400   7.8 GF/s
Conv.2  m=16  k=32   n=102400   6.4 GF/s
Conv.3  m=16  k=288  n=25600   17.3 GF/s
```

On this Mac Conv.0–2 miss AMX too (16×4 / 8×8). Conv.3 lands on apple_amx_mmm_f32_32x32. tiny_rec here is 80% of FMA volume on AMX (tiny_det 59%), which is why rec looks fine here and 3× behind there. medium_rec is the extreme because it’s the most AMX-shaped of the three.

So I would not start with the AVX2 depthwise. ~22% of your det, and it leaves the half you’re actually losing. I’ll still do the NEON side here — Apple det is 31% DW.

If you have a minute, same binary, one thread:

`tract hwbench 16,27,102400 8,64,102400 16,32,102400 16,288,25600 320,160,240`

The <-- pick vs the fastest kernel at those shapes tells us whether the FMA picker is also mis-selecting, or whether they’re just pack-bound.

TarikAlHadethi:
Ran it. **The picker is mis-selecting on two of the four conv shapes**, and it misses in both directions, so it does not look like a simple bias toward one tile.

Same binary, `RAYON_NUM_THREADS=1`, machine idle (21% at launch, the release build had finished and settled).

| shape | picked | GF/s | fastest available | GF/s | |
|---|---|---|---|---|---|
| 16,27,102400 | `fma_mmm_f32_16x6` | 7.93 | `fma_mmm_f32_16x6` | 7.93 | correct |
| **8,64,102400** | `fma_mmm_f32_24x4` | **6.35** | `fma_mmm_f32_8x8` | **12.12** | **1.91x** |
| 16,32,102400 | `fma_mmm_f32_16x6` | 9.14 | `fma_mmm_f32_16x6` | 9.14 | correct |
| **16,288,25600** | `fma_mmm_f32_8x8` | **22.20** | `fma_mmm_f32_16x6` | **32.71** | **1.47x** |
| 320,160,240 | `fma_mmm_f32_16x6` | 33.17 | `fma_mmm_f32_32x3` | 33.81 | within noise |

The two misses invert each other. On `8,64,102400` it should have taken `8x8` and took `24x4`; on `16,288,25600` it should have taken `16x6` and took `8x8`. So it is not "always too wide" or "always too narrow", it is picking `8x8` where it does not belong and skipping it where it does.

On `8,64,102400` the chosen kernel was **seventh** in the list. `8x8` and `avx_8x8` are both ahead of everything else by nearly 4 GF/s, so the gap is not marginal at that shape.

## But it does not explain the 6.7x on its own

Being straight about the size of it, since it would be easy to over-read. Those two nodes in the tiny_det profile were 6.744 ms and 6.802 ms of a 318.659 ms network. Even at the full ratio, picking correctly on both saves single-digit milliseconds, call it 2% of detection.

**What I cannot tell you from this** is whether the same mis-selection is happening across the other 64 `OptMatMul` nodes, which together are the 36.3%. Five shapes is what you asked for and five shapes is what I measured. If the picker is wrong at this rate generally then it matters a great deal more than 2%; if these two are unlucky then it is a real bug worth fixing and not the thing that closes the gap.

Happy to run the full set of conv shapes out of tiny_det if that would settle it. It is the same command with a longer argument list.

## Machine context from the same run

```
20 logical / 12 physical cores
1-thread  L1 : 382.3 GB/s
1-thread  L∞ :  25.8 GB/s
```

The f16 rows are not interesting here: only two f16 kernels are ever offered, so there is nothing to mis-pick, and PP-OCR is f32 throughout.

Happy to paste the full output, losing kernels and all, if that is useful.

czoli1976:
Thanks. Two misses, inverted, and ~2% of det if it's just those two nodes — agreed, that does not close 6.7×.
 
Yes, run the rest. tiny_det's other `OptMatMul`s (n≥2, same binary, one thread):
 
```
tract hwbench 16,576,25600 25600,32,64 16,288,25600 25600,64,32 6400,48,96 6400,96,48 8,64,102400 16,32,102400 1600,64,128 1600,128,64 16,27,102400 400,160,320 400,320,160 25600,64,16 25600,16,64 6400,64,48 6400,48,64 25600,16,32 6400,32,64 1600,96,64 400,128,160 1600,48,96 1600,64,64 6400,64,16 102400,16,4 400,160,64 400,64,128 1600,64,16 20,64,16
```
 
If the picker is wrong at that rate on the big `25600×32×64` / `1600×64×128` tiles, it starts to matter. If not, we are back to packing / huge-N skinny-K, not the FMA kernels.
 
**Full output welcome**; the `<--` column is enough if it's long.

kali (approved):

data/src/dim/parse.rs
```diff
@@ -85,7 +85,15 @@ fn atom<'i>(symbol_table: &SymbolScope, i: &'i str) -> R<'i, TDim> {
...
-        map(|i| func(symbol_table, "floor", i), |xs| xs[0].clone()),
+        // ⚠⚠ There is deliberately NO `floor` here. It used to parse and then
+        // throw itself away, on the reasoning that flooring is a no-op in
+        // integer arithmetic. That is true of a TDim that is ALREADY integral,
+        // and false of the thing the expression was describing: ONNX writes
+        // rational shape expressions, so `floor(W/2 - 1/2)` was read as
+        // `W/2 - 0` and every even W came out one too big, silently. See #2724.
+        // ONNX expressions are now parsed as rational and translated
+        // explicitly, in `tract-onnx`'s `dim_expr` module, which is the only
+        // thing that ever relied on this arm.
```
kali: please avoid narrative comments.

onnx/src/dim_expr.rs
```diff
@@ -0,0 +1,837 @@
...
+/// Deliberately `i64` rather than a bignum: shape expressions are small, and a
+/// dependency is a poor trade for a case that does not occur. Every operation
+/// is checked, and an overflow is reported rather than wrapped — a silently
+/// wrong shape is the exact failure this module exists to end.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+struct Rat {
+    num: i64,
+    den: i64,
+}
+
```
kali: would https://docs.rs/num-rational/latest/num_rational/ work?

onnx/src/dim_expr.rs
```diff
@@ -0,0 +1,837 @@
...
+        }
+        self.terms.retain(|(c, _)| c.num != 0);
+        Ok(self)
+    }
+
+    /// The single fraction this form reduces to: `(integer_terms, denominator)`
+    /// such that the value is `sum(coeff * atom) + konst` over `denominator`,
+    /// with every coefficient an integer and the denominator strictly positive.
+    ///
+    /// This is the step that earns the right to drop a `floor`.
```
kali: last line is narrative.

onnx/src/dim_expr.rs
```diff
@@ -0,0 +1,837 @@
...
+// ---------------------------------------------------------------------------
+
+/// A hand-written recursive-descent parser.
+///
+/// `tract_data`'s `TDim` parser uses nom; this one does not, on purpose. The
+/// grammar is a dozen productions, `tract-onnx` carries no parser dependency
+/// today, and adding one to express *this* would be the larger change. The
+/// grammar it accepts is a superset of what the `TDim` parser accepts for the
+/// same strings, minus the `#` broadcast operator, which no exporter emits.
+struct Parser<'a> {
```
kali: please use nom, you can pick a lot of things from the tdim and nnef parser.

onnx/src/dim_expr.rs
```diff
@@ -0,0 +1,837 @@
...
+
+// ---------------------------------------------------------------------------
+// Normalisation
+// ---------------------------------------------------------------------------
+
+/// `floor(floor(x)/n) == floor(x/n)` for integer `n > 0`.
+///
+/// Applied before linearisation, so the reduction below sees one fraction
+/// rather than a floor divided by something. This is what reaches the nested
+/// case that two stacked stride-2 convolutions produce; without it the result
```
kali: last part is narrative again

onnx/src/dim_expr.rs
```diff
@@ -0,0 +1,837 @@
...
+    to_tdim_inner(&e, scope).with_context(|| format!("translating ONNX dim_param {s:?}"))
+}
+
+/// Evaluate a parsed expression over concrete symbol values, in EXACT rational
+/// arithmetic with real flooring.
+///
+/// This is what the tests check the `TDim` translation against: it is the ONNX
+/// reading of the expression, computed without ever touching tract's integer
+/// division, so agreement between the two is evidence and not a tautology.
+#[cfg(test)]
```
kali, line 282: test in a module, with a single cfg() annotation

kali (changes requested):

kali:
PR is doing the right thing, but it is a bit AI-ish to my taste. Please make sure claude is aware of the comment policy from claude.md. See if we can use num_rationale, and use a compact nom parser instead of a the handrolled one.

kali:
you can ignore the deny red, it's there on main, i'll deal with it.

data/src/dim/parse.rs
TarikAlHadethi: Done, cut it back to just the constraint.

onnx/src/dim_expr.rs
TarikAlHadethi: Yeah, works fine. `Rat` is `Ratio<i64>` now and the hand-rolled arithmetic is gone. Doesn't pull anything new in either, num-integer and num-traits were already there.

onnx/src/dim_expr.rs
TarikAlHadethi: Gone.

onnx/src/dim_expr.rs
TarikAlHadethi: Done, picked up the `bin!` macro and the stag/identifier/numeric helpers from the tdim parser.

Couple of things I did differently. `/` takes an atom on the right instead of just a numeric, so `W/N` parses and then gets refused by name rather than dying as a parse error. And floor checks its arity now, so `floor(a,b)` is rejected instead of quietly turning into `floor(a+b)`.

onnx/src/dim_expr.rs
TarikAlHadethi: Removed.

onnx/src/dim_expr.rs
TarikAlHadethi: Moved it inside `mod test`.

TarikAlHadethi:
All six done in 94aa088.

num-rational and nom both went in. On the comments, fair enough. I'd written them the way I do in my own repo, which is pretty much the opposite of what claude.md asks for, so I've cut them right back to the constraints and dropped the history and issue numbers. The section banners went too.

One extra thing while I was in there: floor checks its arity at parse time, so `floor(a,b)` gets rejected instead of quietly becoming `floor(a+b)`.

fmt and clippy are clean on tract-onnx and tract-data and the tests pass. Workspace clippy I've left to CI, core-foundation won't build here on Windows.

TarikAlHadethi:
@czoli1976 ran the full list. Short version: the two tiles you named are picked correctly, so this is packing / huge-N skinny-K rather than the FMA kernels.

```
25600x32x64    picked fma_16x6 10.68   fastest fma_24x4 11.01   1.03x
1600x64x128    picked fma_16x6 19.66   fastest fma_32x3 20.54   1.04x
```

Both inside 4%.

It's wrong on 12 of 29 overall though, which is higher than I expected, and it clusters at narrow N: 9 of 17 mis-picked at N<=64 against 3 of 12 above. 24x4 wins 9 of those 12 and gets chosen for one. 8x8 gets chosen for 5 and wins one. Worst is still the m=8 case at 2.00x, where 8x8 is the exact fit and it takes 24x4 and throws away 16 of 24 rows.

```
shape             picked        GF/s   fastest       GF/s   loss
16x576x25600      8x8          28.20   16x6         42.87   1.52x
25600x32x64       16x6         10.68   24x4         11.01
16x288x25600      8x8          26.31   16x6         37.02   1.41x
25600x64x32       16x6         16.97   24x4         19.80   1.17x
6400x48x96        16x6         15.85   16x6         15.85
6400x96x48        16x6         25.39   16x6         25.39
8x64x102400       24x4          6.88   8x8          13.75   2.00x
16x32x102400      16x6          9.88   16x6          9.88
1600x64x128       16x6         19.66   32x3         20.54
1600x128x64       16x6         30.82   24x4         31.77
16x27x102400      16x6          8.64   16x6          8.64
400x160x320       16x6         35.95   16x6         35.95
400x320x160       16x6         48.22   24x4         48.58
25600x64x16       16x6         17.11   24x4         19.53   1.14x
25600x16x64       16x5          5.22   24x4          6.30   1.21x
6400x64x48        16x6         19.97   32x3         20.31
6400x48x64        16x6         15.39   24x4         16.20   1.05x
25600x16x32       16x5          4.87   24x4          6.31   1.30x
6400x32x64        16x6         11.31   24x4         11.75
1600x96x64        16x6         25.62   24x4         26.86
400x128x160       16x6         31.93   16x6         31.93
1600x48x96        16x6         16.44   32x3         16.76
1600x64x64        16x6         19.54   24x4         20.27
6400x64x16        16x6         17.63   24x4         20.35   1.15x
102400x16x4       24x4          6.21   24x4          6.21
400x160x64        8x8          27.64   24x4         35.27   1.28x
400x64x128        16x6         19.70   24x4         20.15
1600x64x16        8x8          14.83   24x4         19.81   1.34x
20x64x16          8x8          10.33   24x4         15.04   1.46x
```

All `fma_*`. The `avx_*` variants run 3 to 8% behind their FMA twins everywhere and never win.

Same caveat as before, and you made the point yourself: I can't weight these by wall clock, so I don't know whether the 41% sits on hot nodes or cold ones. Doesn't change your conclusion either way.

f16 left out, only two f16 kernels are offered so there's nothing to mis-pick, and this model is f32 throughout.

kali:
> All six done in [94aa088](https://github.com/sonos/tract/commit/94aa088ed59e083a5168e933daa6b71cd9f5e080).
> 
> num-rational and nom both went in. On the comments, fair enough. I'd written them the way I do in my own repo, which is pretty much the opposite of what claude.md asks for, so I've cut them right back to the constraints and dropped the history and issue numbers. The section banners went too.
> 
> One extra thing while I was in there: floor checks its arity at parse time, so `floor(a,b)` gets rejected instead of quietly becoming `floor(a+b)`.
> 
> fmt and clippy are clean on tract-onnx and tract-data and the tests pass. Workspace clippy I've left to CI, core-foundation won't build here on Windows.

thanks !

czoli1976:
@TarikAlHadethi please open an issue for the Tract vs ORT PErformance Gap as observed on your system (Do not compare with my numbers, Apple Silicon has AMX/SME so not directly comparables, just TRACT vs ORT on yours)

kali:
i'm ready to merge this folks, I suggest the performance conversation move some place else

czoli1976:
@TarikAlHadethi draft with the NCHW spatial kernels (DirectSpatial, DW along-W, MaxPool, nearest upsample, deconv unpack) for you to try on the i7: https://github.com/sonos/tract/pull/2771

Same 1-thread recipe as before, plus `--onnx-ignore-value-info --onnx-ignore-output-shapes`. Will not close 6.7× — leftover is still packing / skinny-K FMA. Tiny_det dump should show three `DirectSpatialConv` (Conv.0/1/2).

PR #2735, opened 2026-08-27
(perf) core/cnn: vectorise depthwise along contiguous W

czoli1976:
`DepthWiseConv`'s inner loop (`process_zone_n`) was scalar. On NCHW with W innermost and unit output stride that is a contiguous FIR along W: splat each kernel tap and FMA a 4-wide (then 8-wide) load of X.

DPDFNet 48 kHz-hr encoder DW is that shape with **stride 2 or 3** along frequency (`kw=3`, `group=C=64`, `W=481`). The stride-1 path never ran there; those taps use `vld2q` / `vld3q` to de-interleave. Padded / non-unit output-stride zones stay on the scalar loop. `BlockedConv` is untouched (wasm / `kw=1` / small `ocg`).

Does not apply to FastEnhancer or DTLN (no `DepthWiseConv`). DFN3 encoder/erb have the same DW family but it is ~1% of FMA and does not move the frame.

### Measured

p50 ms/frame, median of 3 interleaved rounds, single-threaded, M1 Pro. Same tree with and without this patch; binaries hash-checked as distinct. Median **and** min: a row whose signs disagree is drift.

| model | base | this PR | median | min |
|---|---|---|---|---|
| dpdfnet2_48khz_hr | 1.926 | 1.865 | **+3.2%** | **+3.2%** |
| dpdfnet2 | 1.276 | 1.193 | +6.5% | +0.9% |
| gtcrn | 0.605 | 0.603 | +0.2% | −0.1% |
| dfn3-enc | 4.886 | 4.720 | +3.4% | +0.1% |
| fe-tiny | 0.107 | 0.110 | −2.9% | +0.1% |
| dtln-2 | 0.052 | 0.052 | +0.2% | 0.0% |

The 48 kHz-hr row is the real one (min and median agree). dpdfnet2's p50 is one slow base round; min is ~1%. Everything else is noise or has no DepthWise op.

Vs ONNX Runtime CPU EP (intra=1, inter=1) after this patch: dpdfnet2_48khz_hr is still **1.33×** ORT. This is a few percent of tract, not a catch-up to FusedConv.

### Correctness

500-frame `dpdfnet2_48khz_hr` stream, skip-only vs this PR: max abs **8.9e-07**, max rel 1.0e-05, no frame-index drift. Not bit-identical (`vfmaq` vs scalar `*`/`+`); zero samples above 1e-6.

Same dump vs ORT: max abs 2.3e-06 (skip-only was 1.9e-06). Under the 3e-05 bar.

Unit test `depthwise_contig_w_matches_reference` covers `W=481`, `kw=3`, Valid/SameUpper, stride 1/2/3. `test-unit-core` conv filter: 501 passed.

🍍

### DIFF
--- a/cli/src/routines.rs
+++ b/cli/src/routines.rs
@@ -55,6 +55,7 @@ fn family(factory: &RoutineFactory) -> (u8, &'static str) {
         RoutineFactory::RmsNormF32 { .. } => (4, "fused row-wise"),
         RoutineFactory::LutU8 { .. } => (5, "look-up table"),
         RoutineFactory::BinF32 { .. } | RoutineFactory::BinF16 { .. } => (6, "binary"),
+        RoutineFactory::DepthwiseWF32 { .. } => (7, "depthwise"),
     }
 }
 
--- a/core/src/ops/cnn/conv/depth_wise.rs
+++ b/core/src/ops/cnn/conv/depth_wise.rs
@@ -206,6 +206,25 @@ macro_rules! impl_eval {
                     while !visitor.done {
                         let iptr = iptr.offset(visitor.input_center_offset);
                         let optr = optr.offset(visitor.output_offset);
+                        #[cfg(target_arch = "aarch64")]
+                        if std::any::TypeId::of::<T>() == std::any::TypeId::of::<f32>()
+                            && visitor.inner_loop_output_stride == 1
+                            && visitor.inner_loop_input_full_stride >= 1
+                        {
+                            let k_f32 = *(&k as *const [T; N] as *const [f32; N]);
+                            let bias_f32 = *(&bias as *const T as *const f32);
+                            neon_depthwise_w_f32::<N>(
+                                iptr as *const f32,
+                                optr as *mut f32,
+                                &k_f32,
+                                &ioffset,
+                                bias_f32,
+                                visitor.inner_loop_len,
+                                visitor.inner_loop_input_full_stride,
+                            );
+                            visitor.next_non_inner_axis();
+                            continue;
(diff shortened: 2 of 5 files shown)


czoli1976:
@kali , am I pretty?
image

kali:
Any structural reason not to move this to a routine in linalg ?

czoli1976:
> ✅ **Bench vs main — no regressions**
> 
> _Reference: 2026-09-07 morning nightly run (0d old) · full report → [run](https://github.com/sonos/tract/actions/runs/34152502484)_
> 
> **Speed** — evaltime · prefill · decode
> 
> _no inference-speed regressions_

Wow

kali:
@czoli1976 do you want to address the red on wasm ?

kali:
Thank you, cutting a release this morning, trying to fast track this one.

kali (approved):

czoli1976:
Done — it was the test, not the kernel. The `kw=1` case is f32 NCHW with unit W stride and `ocg == 1`, which is exactly what `try_blocked_conv` claims, and blocked conv is on by default on wasm. So that shape lowered to `BlockedConv` and the `expected DepthWiseConv` assert aborted the run. The test now accepts `BlockedConv` for `kw=1` and leaves it to `blocked_conv_matches_reference`; the escape is scoped to `kw=1`, so native still asserts `DepthWise` there and a `kw=3` routing regression would still fail. Verified locally on both CI configs (`+simd128` and `+simd128,+relaxed-simd`, linalg/core/test-unit-core) before pushing.

PR #2409, opened 2026-06-24
fix: SimplifiedLayerNormalization mapped to RMSNorm instead of LayerNorm

jaweed3:
## Problem

`SimplifiedLayerNormalization` (LayerNorm without bias) was registered to the `RMSNorm` handler (`rms_norm::rms_normalization`). RMSNorm computes `X / RMS(X) * Scale`, while SimplifiedLayerNorm computes `(X - μ) / σ * Scale` (mean-subtracting LayerNorm without the bias term). The math is incorrect.

## Solution

Re-register to a new handler that constructs `LayerNorm` with `have_bias: false` and no optional mean/invstddev outputs. The existing `LayerNorm` expansion in `layer_norm.rs` already supports this configuration.

## Changes

- `onnx/src/ops/nn/layer_norm.rs`: add `LayerNorm::new()` constructor
- `onnx/src/ops/nn/simplified_layer_norm.rs` (new): handler that creates `LayerNorm` with `have_bias: false`
- `onnx/src/ops/nn/mod.rs`: register new handler, remove incorrect mapping

## Notes

This is a pure ONNX-level decomposition fix. No core ops, GPU kernels, or NNEF serialization changes needed.

Closes #1775.

PR #2749, opened 2026-08-30
onnx: lower SimplifiedLayerNormalization to RMS norm, not LayerNorm

### LINKED ISSUE #2646: ONNX import: SimplifiedLayerNormalization lowers to LayerNorm, not RMS norm
`SimplifiedLayerNormalization` currently lowers to `LayerNorm`, which subtracts the mean. The operator is RMS normalization: it scales by the root mean square and does **not** centre the input. Models importing cleanly and running to completion therefore produce wrong activations, with no error anywhere.

Found while bringing an ORT-GenAI fused export (Qwen3 architecture) through tract: tract's logits diverged from onnxruntime, and bisecting the intermediates put the first divergence on this node.

## Repro

A single node, with an input whose mean is deliberately non-zero so the two formulas separate:

```python
import numpy as np, onnx, onnxruntime
from onnx import helper, TensorProto
vi = helper.make_tensor_value_info
node = helper.make_node("SimplifiedLayerNormalization", ["X", "W"], ["Y"], axis=-1, epsilon=1e-5)
g = helper.make_graph([node], "sln",
                      [vi("X", TensorProto.FLOAT, [3, 8]), vi("W", TensorProto.FLOAT, [8])],
                      [vi("Y", TensorProto.FLOAT, [3, 8])])
m = helper.make_model(g, opset_imports=[helper.make_opsetid("", 17),
                                        helper.make_opsetid("com.microsoft", 1)])
m.ir_version = 10
onnx.save(m, "sln.onnx")

r = np.random.RandomState(5)
d = dict(X=(r.randn(3, 8) * 2 + 1).astype(np.float32), W=(r.rand(8) + 0.5).astype(np.float32))
np.savez("sln_in.npz", **d)
y = onnxruntime.InferenceSession("sln.onnx").run(None, d)[0]

x, w = d["X"].astype(np.float64), d["W"].astype(np.float64)
rms = x / np.sqrt((x ** 2).mean(-1, keepdims=True) + 1e-5) * w
ln = (x - x.mean(-1, keepdims=True)) / np.sqrt(x.var(-1, keepdims=True) + 1e-5) * w
print("ORT vs RMSNorm  ", np.abs(y - rms).max())   # 2.6e-07
print("ORT vs LayerNorm", np.abs(y - ln).max())    # 1.8e-01
```

```
tract sln.onnx --input-facts-from-bundle sln_in.npz run \
  --input-from-bundle sln_in.npz --assert-output-bundle sln_in.npz
```

On `d5d2fb904` tract's output matches the **LayerNorm** expression to 1.5e-07 and differs from onnxruntime by 1.8e-01. First few values of one row:

```
onnxruntime  [-2.1760528  -1.8019617  -0.531172    0.21287762]
tract        [-2.0468385  -1.6825728  -0.4022511  0.375858  ]
```

## Where it comes from

Before #2409 the op was registered on `rms_norm::rms_normalization`, which is correct. #2409 (titled "SimplifiedLayerNormalization mapped to RMSNorm instead of LayerNorm") moved it onto a new `simplified_layer_norm` handler that constructs `LayerNorm`:

https://github.com/sonos/tract/blob/main/onnx/src/ops/nn/simplified_layer_norm.rs

Note `SkipSimplifiedLayerNormalization` right below it in the registry is unaffected — it goes to `rms_norm_contrib` and is correct — so a model that fuses most of its norms into the Skip form shows the discrepancy only where the plain op survives, which makes it easy to miss.

## How to fix

Point the registration back at `rms_normalization` and delete the handler; `LayerNorm::new` then has no callers and can go too. `LayerNormalization` keeps using `layer_norm::layer_norm` and is unaffected.

```diff
--- a/onnx/src/ops/nn/mod.rs
+++ b/onnx/src/ops/nn/mod.rs
@@ -27,7 +27,6 @@ mod reduce;
 mod rms_norm;
 mod rms_norm_contrib;
 mod rotary_embedding;
-mod simplified_layer_norm;
 mod skip_layer_norm;
 
@@ -101,7 +100,7 @@ pub fn register_all_ops(reg: &mut OnnxOpRegister) {
     reg.insert("RMSNormalization", rms_norm::rms_normalization);
     reg.insert("RotaryEmbedding", rotary_embedding::rotary_embedding);
-    reg.insert("SimplifiedLayerNormalization", simplified_layer_norm::simplified_layer_norm);
+    reg.insert("SimplifiedLayerNormalization", rms_norm::rms_normalization);
```

plus `rm onnx/src/ops/nn/simplified_layer_norm.rs` and dropping the now-unused `impl LayerNorm { pub fn new(..) }` block in `onnx/src/ops/nn/layer_norm.rs`.

With that applied, the single-node repro above matches onnxruntime to 7.5e-09.

One wrinkle if this gets a case under `onnx/test_cases/`: tract's and onnxruntime's `rsqrt` disagree in the last f32 ulp, so `--assert-output-bundle` at the default `close` tolerance reports one outlier in 24. `run_all.sh` passes `$OPTIONS` before the `run` subcommand, so `--approx` cannot be threaded through from `vars.sh` as it stands.

Jorge-Polanco-Roque:
**Problem:** `SimplifiedLayerNormalization` (onnxruntime contrib op) is RMS normalization — it scales by the root mean square and does **not** subtract the mean. tract lowered it to `LayerNorm`, which centres the input, so every model using it imported and ran but produced wrong activations, with no warning.

**Fix:** delegate to the existing `rms_normalization` parser (the same op already used for `SkipSimplifiedLayerNormalization`), which normalizes without centring. Removed the now-unused `LayerNorm::new`.

**Tests:** added `simplified_layer_norm_is_rms_norm`, an end-to-end test that builds and runs a single-node model on a non-zero-mean input and checks the RMS-norm result; verified it fails against the old `LayerNorm` lowering. `cargo fmt`, `clippy`, and the `tract-onnx` suite are green.

Closes #2646

### DIFF
--- a/onnx/src/ops/nn/layer_norm.rs
+++ b/onnx/src/ops/nn/layer_norm.rs
@@ -33,19 +33,6 @@ pub struct LayerNorm {
     invstddev_output: Option<usize>,
 }
 
-impl LayerNorm {
-    pub fn new(axis: isize, epsilon: f32, datum_type: DatumType) -> Self {
-        LayerNorm {
-            axis,
-            epsilon,
-            datum_type,
-            have_bias: false,
-            mean_output: None,
-            invstddev_output: None,
-        }
-    }
-}
-
 impl Expansion for LayerNorm {
     fn name(&self) -> StaticName {
         "LayerNorm".into()
--- a/onnx/src/ops/nn/simplified_layer_norm.rs
+++ b/onnx/src/ops/nn/simplified_layer_norm.rs
@@ -72,15 +72,17 @@ mod tests {
         let runnable =
             crate::onnx().model_for_read(&mut &*buf)?.into_optimized()?.into_runnable()?;
 
-        // x has a non-zero mean; scale is all ones.
+        // x has a non-zero mean; scale is non-unit so the test also exercises the
+        // elementwise scale multiply (a lowering that dropped input 1 would fail).
         let x = tensor2(&[[1f32, 2., 3., 4.]]);
-        let scale = tensor1(&[1f32, 1., 1., 1.]);
+        let scale = tensor1(&[0.5f32, 1.5, 2.0, 0.25]);
         let out = runnable.run(tvec!(x.into_tvalue(), scale.into_tvalue()))?;
 
         // RMS norm: rms = sqrt(mean(x^2) + 1e-5) = sqrt(7.5 + 1e-5) ~= 2.7386
-        //   y = x / rms ~= [0.3651, 0.7303, 1.0954, 1.4606]
-        // (LayerNorm would centre first and yield [-1.342, -0.447, 0.447, 1.342].)
-        let expected = tensor2(&[[0.365_15f32, 0.730_30, 1.095_45, 1.460_60]]);
+        //   norm = x / rms ~= [0.3651, 0.7303, 1.0954, 1.4606]
+        //   y = norm * scale ~= [0.1826, 1.0954, 2.1909, 0.3651]
+        // (LayerNorm would centre first and yield a completely different result.)
+        let expected = tensor2(&[[0.182_574f32, 1.095_444, 2.190_889, 0.365_148]]);
         out[0].close_enough(&expected, Approximation::Approximate)?;
         Ok(())
     }
--- a/onnx/src/ops/nn/simplified_layer_norm.rs
+++ b/onnx/src/ops/nn/simplified_layer_norm.rs
@@ -1,14 +1,87 @@
 use crate::model::ParsingContext;
-use crate::ops::nn::layer_norm::LayerNorm;
+use crate::ops::nn::rms_norm::rms_normalization;
 use crate::pb::NodeProto;
 use tract_hir::internal::*;
 
 pub fn simplified_layer_norm(
-    _ctx: &ParsingContext,
+    ctx: &ParsingContext,
     node: &NodeProto,
 ) -> TractResult<(Box<dyn InferenceOp>, Vec<String>)> {
-    let axis = node.get_attr_opt("axis")?.unwrap_or(-1);
-    let epsilon = node.get_attr_opt("epsilon")?.unwrap_or(1e-5);
-    let datum_type = node.get_attr_opt("stash_type")?.unwrap_or(DatumType::F32);
-    Ok((expand(LayerNorm::new(axis, epsilon, datum_type)), vec![]))
+    // `SimplifiedLayerNormalization` is RMS normalization: it scales by the root
+    // mean square and does *not* subtract the mean. Lowering it to `LayerNorm`,
+    // which centres the input, produced wrong activations for every non-zero-mean
+    // input (and silently dropped the optional bias input). Delegate to the RMS
+    // normalization parser instead (#2646).
+    rms_normalization(ctx, node)
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::pb::*;
+    use prost::Message;
+    use tract_hir::internal::*;
+
+    fn float_value_info(name: &str, dims: &[i64]) -> ValueInfoProto {
+        let dim = dims
+            .iter()
+            .map(|d| tensor_shape_proto::Dimension {
+                denotation: String::new(),
+                value: Some(tensor_shape_proto::dimension::Value::DimValue(*d)),
+            })
+            .collect();
+        ValueInfoProto {
+            name: name.to_string(),
+            r#type: Some(TypeProto {
+                denotation: String::new(),
+                value: Some(type_proto::Value::TensorType(type_proto::Tensor {
+                    elem_type: tensor_proto::DataType::Float as i32,
+                    shape: Some(TensorShapeProto { dim }),
+                })),
+            }),
+            doc_string: String::new(),
+        }
+    }
+
+    // `SimplifiedLayerNormalization` is RMS normalization, which does not subtract
+    // the mean. With a non-zero-mean input the previous `LayerNorm` lowering gave a
+    // completely different (centred) result. Regression test for #2646.
+    #[test]
+    fn simplified_layer_norm_is_rms_norm() -> TractResult<()> {
+        let node = NodeProto {
+            name: "sln".to_string(),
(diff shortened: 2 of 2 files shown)


czoli1976:
Thanks for picking this up — I filed #2646 and I've built and validated your branch. The fix is correct. Notes below, plus the numbers in case they're useful for the PR description.

## Validated against onnxruntime

Built `2ff9f3bfe9`, ran single-node `SimplifiedLayerNormalization` models through both ORT 1.19.2 and the tract CLI:

| case | attrs | tract vs ORT | ORT vs RMS norm | ORT vs LayerNorm |
|---|---|---|---|---|
| rank-2 | defaults | 7.5e-09 | 1.7e-07 | 1.06 |
| rank-3 | `axis=2, epsilon=1e-3` | 1.2e-07 | 1.2e-07 | 1.52 |
| rank-3 | defaults | 1.2e-07 | 1.7e-07 | 3.88 |
| rank-3 | `axis=1` (multi-axis reduce, rank-2 scale) | 2.4e-07 | 2.0e-07 | 0.65 |

The `axis=1` case matters because it exercises reduction over the last *two* dims with a rank-2 scale broadcast — that path is fine.

f16 activations (via a Cast-wrapped model, since the CLI can't load f16 npz inputs): 1.8e-3 abs / 9.6e-4 rel vs ORT, i.e. f16 rounding, no systematic error.

Housekeeping on my box: `cargo fmt --check` and `clippy -p tract-onnx --all-targets` clean, `cargo test -p tract-onnx` 30/30, and the ONNX conformance suite `cargo test -p test-onnx-core` **4532 passed / 0 failed**.

## Your test is a genuine regression test

I restored the old `LayerNorm::new` lowering while keeping `simplified_layer_norm_is_rms_norm` in place. It fails as intended:

```
Error: Mismatch. First outlier: Approximate for F32) at [0, 0] -1.3416355 != 0.36515.
Outliers: 4 / 4 = 1.00000 > 0.00000.
```

Good call putting it in-crate rather than under `onnx/test_cases/` — as I noted at the bottom of #2646, tract's and ORT's `rsqrt` disagree in the last f32 ulp, and `run_all.sh` passes `$OPTIONS` before the `run` subcommand so `--approx` can't currently be threaded through from `vars.sh`.

## A perf upside worth adding to the description

This isn't only a correctness fix — it also lets the node fold into the fused kernel. Optimized graph before (i.e. the `LayerNorm` lowering):

```
Reduce<Sum> → Mul → Sub → RmsNorm → AddAxis → Mul       (8 nodes)
```

after:

```
RmsNorm                                                  (1 node)
```

## One correction for the PR body

> The lowering also hard-coded `have_bias: false`, dropping the optional bias input.

`SimplifiedLayerNormalization` has no bias input — ORT rejects a 3-input node at load:

```
Node () has input size 3 not in range [min=2, max=2]
```

So `rms_normalization`'s `have_bias = node.input.len() >= 3` can never fire for this op. The code is right either way; it's just that the sentence would land in the history as an ORT fact that isn't one. (The one that *does* take an extra input is `SkipSimplifiedLayerNormalization` — and ORT >= 1.19 ignores its bias, which is why `rms_norm_contrib` handles it separately.)

## Two optional suggestions

1. **Make `scale` non-unit in the test.** It's currently `[1, 1, 1, 1]`, so a lowering that dropped input 1 entirely would still pass. Something like `[0.5, 1.5, 2.0, 0.25]` with the expected values scaled accordingly closes that hole for free.

2. **`stash_type` is now unread.** The old handler parsed it; `rms_normalization` hard-codes an F32 stash. For the default (`stash_type=1`) that's identical, and `RMSNormalization` already has the same gap, so this is not a regression introduced here — but it's worth a line in the description so it isn't mistaken for an oversight. Happy to file that as its own issue.

Not blocking, purely a taste call: #2646 suggested deleting `simplified_layer_norm.rs` and registering `rms_norm::rms_normalization` directly. Keeping the forwarding module is arguably better — it gives the regression test a home and avoids touching the `mod.rs` registry block, which conflicts constantly.

Separately, and unrelated to this PR: the two-output form (`Y`, `inv_std_var`) still fails at analyse with `Wrong number of outputs. Op says 1, node says 2.` ORT accepts it. Same before and after your change; I'll file it separately.

czoli1976:
Correction to my earlier comment — I tested this rather than assuming, and I had it wrong.

ORT **ignores `stash_type`** for `SimplifiedLayerNormalization` and always accumulates in f32. Same f16 input, `stash_type=1` vs `stash_type=10`, with magnitudes chosen so a real f16 accumulator would overflow (`sum(x^2)` ~ 1.2e6 against f16's 65504 ceiling, which would saturate to `inf` and poison the output):

```
stash_type=1   y[0,:4]: [1.07 1.138 0.945 1.099]   finite
stash_type=10  y[0,:4]: [1.07 1.138 0.945 1.099]   finite
bit-identical: True      both vs f32-stash reference: 4.8e-04
```

So `rms_normalization`'s hard-coded F32 stash matches ORT exactly. **Please disregard suggestion 2** — there is nothing to note in the description, and no follow-up issue to file. Sorry for the detour.

(Suggestion 1 — a non-unit `scale` in the test — still stands, and the correction on the bias sentence still stands: that one I did verify against ORT's loader.)

Separately, I filed #2762 for the optional `inv_std_var` second output, which tract still rejects at analyse. That gap predates #2409 and is unrelated to your change — **it is not something this PR should take on.** I have a verified patch on it and will send it as its own PR; it touches only `rms_norm.rs`, so it will not conflict with yours.

This PR is good as it stands.

Jorge-Polanco-Roque:
Thanks — applied suggestion 1: the test now uses a non-unit `scale` `[0.5, 1.5, 2.0, 0.25]` with the expected values recomputed accordingly, so it genuinely exercises the elementwise scale (a lowering that dropped the scale input would now fail). Appreciate the careful review and the ORT correction on suggestion 2 — and noted that #2762 is separate, won't touch it here.

kali (approved):

kali (approved):

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2761, opened 2026-09-01
GatedDeltaNet recurrent and causal-conv1d-update kernels

JulienBalianSonos:
Split out of #2084, which bundles far more than its title suggests. This
piece is self-contained: GatedDeltaNet (GDN) and causal-conv1d-update
support, the linear-attention core used by Qwen3.5's hybrid decoder. Not
MoE-related, no dependency on the rest of #2084.

What's in it:

- A GQA-aware `GatedDeltaNetRecurrent` op, generalized to an S-generic
  `[b, S, h, w]` layout, with a threadgroup-parallel Metal kernel
  (row-chunked reductions), a chunked prefill kernel, and an f16-state
  variant.
- An S-generic `CausalConv1dUpdate` op with a `[b, C, S]` layout.
- The beta sigmoid folded directly into the GDN kernel dispatch, via a
  Metal rewrite rule.
- Matching CUDA and Metal kernels, plus fixes so both work at any
  sequence length (symbolic S) on the decode graph.

## Benchmarks

I added a criterion bench comparing the pre-existing CPU reference
implementation against the new GPU kernels, on both backends. Both
benches include a full device-completion sync inside the timed region,
so the numbers reflect real GPU latency rather than just command-buffer
encoding time.

**Metal** (`cargo bench -p tract-metal --bench gdn_recurrent`, on a Mac):
the real Qwen3.5-35B geometry — 16 k-heads, GQA groups=2, head width 128.

| | CPU | Metal, old per-token kernel | Metal, this PR's chunked path |
|---|---|---|---|
| Decode (1 step) | ~2.0 ms | – | ~120–160 µs (**13–17x** faster) |
| Prefill (512 tokens) | 302.6 ms | 6.454 ms (**47x** faster) | 5.596 ms (**54x** faster) |

The middle column is the older, pre-chunking threadgroup kernel that this
PR's chunked prefill path builds on top of (forced on via
`TRACT_METAL_DISABLE_GDN_CHUNKED=1` for the comparison). At this
particular length — 512 tokens, 8 chunks — the chunking optimization only
adds about 15% on top of the much larger CPU-to-Metal speedup; I'd
expect that gap to widen at longer prefill lengths, but I haven't
verified that independently.

**CUDA** (`cargo bench -p tract-cuda --bench gdn_recurrent`, on an A10G):
same shapes but ungrouped heads (32 heads, no GQA split) and an f32
state, since that's the only configuration the CUDA kernel supports
today. The CPU numbers differ from the Metal table above because this
ran on different hardware with a slightly different input shape, not
because the CPU op itself changed.

| | CPU | CUDA |
|---|---|---|
| Decode (1 step) | 7.09 ms | 25.0 µs (**~283x** faster) |
| Prefill (512 tokens) | 781.2 ms | 9.787 ms (**~80x** faster) |

Decode is unchanged by this PR — the old CUDA kernel already handled a
single step correctly (its shape math just happens to reduce to the same
thing for S=1). Prefill is the real story here, though: I checked, and
the old kernel doesn't have any real handling of a sequence axis at all.
Feeding it a genuine 512-token call trips its state-shape check (it
infers a bogus head count from the flattened tensor size) and it errors
out, which means any real prefill request would have silently fallen
back to the CPU op entirely — there was no GPU-accelerated prefill on
CUDA before this PR, at any speed. So the CUDA prefill row above isn't
"kernel got faster", it's "GPU prefill exists now, where it didn't
before", via a plain per-step host-side loop (still no chunking — that
would be the natural CUDA follow-up, mirroring the Metal kernel).

🍍

### DIFF
--- a/cuda/Cargo.toml
+++ b/cuda/Cargo.toml
@@ -64,3 +64,7 @@ default = ["cuda-13000"]
 [[bench]]
 name = "cuda_flash"
 harness = false
+
+[[bench]]
+name = "gdn_recurrent"
+harness = false
--- a/cuda/benches/gdn_recurrent.rs
+++ b/cuda/benches/gdn_recurrent.rs
@@ -67,14 +67,17 @@ fn cpu_gdn(crit: &mut BenchmarkGroup<WallTime>, label: &str, inputs: &Inputs) {
     crit.bench_function(&format!("cpu_{label}"), |be| {
         be.iter(|| {
             GatedDeltaNetRecurrent::default()
-                .eval(&EvalContext::out_of_plan(), tvec![
-                    inputs.q.clone().into_tvalue(),
-                    inputs.k.clone().into_tvalue(),
-                    inputs.v.clone().into_tvalue(),
-                    inputs.g.clone().into_tvalue(),
-                    inputs.beta.clone().into_tvalue(),
-                    inputs.state.clone().into_tvalue(),
-                ])
+                .eval(
+                    &EvalContext::out_of_plan(),
+                    tvec![
+                        inputs.q.clone().into_tvalue(),
+                        inputs.k.clone().into_tvalue(),
+                        inputs.v.clone().into_tvalue(),
+                        inputs.g.clone().into_tvalue(),
+                        inputs.beta.clone().into_tvalue(),
+                        inputs.state.clone().into_tvalue(),
+                    ],
+                )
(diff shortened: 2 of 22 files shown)


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2766, opened 2026-09-03
Harden read_tensor against untrusted NNEF string length (CWE-770)

li-jin-quan:
## Summary

`tract-nnef` parses NNEF model files with `read_tensor` (`nnef/tract-nnef/src/tensors.rs`).
For `String`-typed tensors, the per-item byte length `len: u32` is read **directly from the
(untrusted) file**, then used to pre-allocate and `set_len` an unbounded buffer:

```rust
let len: u32 = reader.read_u32::<LE>()?;
let mut bytes = Vec::with_capacity(len as usize);
#[allow(clippy::uninit_vec)]
unsafe { bytes.set_len(len as usize); }   // <- unbounded upfront allocation + uninit memory
reader.read_exact(&mut bytes)?;
```

A malicious NNEF file that advertises a multi-gigabyte string length forces a huge **upfront**
heap allocation *before a single byte is read* → instant memory blow-up / abort on OOM (CWE-770 /
CWE-400 resource exhaustion DoS). The `unsafe set_len` also left the buffer uninitialized until
`read_exact` overwrote it.

### Fix

Allocate from what actually arrives and verify the real length matches the declaration:

```rust
let len: u32 = reader.read_u32::<LE>()?;
let mut bytes = Vec::new();
(&mut reader).take(len as u64).read_to_end(&mut bytes)?;
if bytes.len() != len as usize {
    bail!("NNEF string item declared {len} bytes but only {} readable", bytes.len());
}
*item = String::from_utf8(bytes)?;
```

No API change, no behavior change for well-formed files (verified: a 3-byte string still loads as
`"abc"`). Only the allocation policy at a trust boundary is hardened.

## Details

- **Trust boundary:** the NNEF file is untrusted input (downloaded models, user-supplied graphs).
  Any `len` field in it must be treated as hostile, not as an allocation hint.
- **Amplification:** a 4-byte header value (`u32::MAX` ≈ 4 GiB, or any large value) lets a tiny
  file demand an enormous reservation. On 64-bit this is limited only by `usize`, but even a
  `128 MiB` declaration from a 3-byte body is a 40,000,000× amplification — enough to OOM modest
  runners / WASM edge deployments.
- **Unsafe removal:** the previous `unsafe { set_len(len) }` relied on `read_exact` fully
  overwriting the buffer; now the buffer is always initialized by `read_to_end`, eliminating the
  uninitialized-memory footgun.

## Impact

| Vector | Before | After |
| --- | --- | --- |
| Allocation from untrusted `len` | `Vec::with_capacity(len)` (unbounded) | bounded by bytes actually present |
| Uninitialized memory | yes (`set_len` before `read_exact`) | no |
| DoS on hostile file | pre-allocates ~declared size, may OOM | allocates only real bytes, errors cleanly |
| Well-formed files | unchanged | unchanged |

## Testing

- Added regression test `nnef/tract-nnef/tests/nnef_string_len_alloc.rs` (driven by a counting
  `#[global_allocator]` that records the largest single allocation). It crafts a string tensor
  declaring a **128 MiB** length but shipping only 3 bytes, and asserts the largest single heap
  allocation stays bounded (≤ 16 MiB). On the old code the assertion fails (allocates ~128 MiB);
  on the patched code it passes.
- Local standalone PoC reproducing the exact two code paths confirms the numbers:
  - vulnerable path → largest single alloc = **134,217,728 bytes (128 MiB)**
  - patched path   → largest single alloc = **~100 bytes**

> Note: the in-repo regression test was not executed in CI here because the full `tract` build
> requires an MSVC/MinGW assembler toolchain (`tract-linalg` MASM kernels) that is absent from the
> audit environment. The standalone PoC (`tract_nnef_cwe770_poc`) provides the local empirical
> proof and runs on any host with a Rust toolchain. Maintainers can run
> `cargo test --test nnef_string_len_alloc` in a normal dev environment.

## Notes for maintainers

- This is a drop-in safety fix; no public API/signature changed.
- The crafted-header constants in the test mirror `Header`'s 128-byte layout (magic `0x4E 0xEF`,
  version 1.0, `bits_per_item = u32::MAX` String sentinel, `item_type = 0x1000`); if the internal
  header parsing is tightened later, only the fixture needs adjustment — the assertion logic is
  the important part.

## Suggested commit / branch

- branch: `fix/nnef-string-len-unbounded-alloc`
- applies to: `nnef/` (file `nnef/src/tensors.rs`; test `nnef/tests/nnef_string_len_alloc.rs`)

### DIFF
--- a/nnef/src/tensors.rs
+++ b/nnef/src/tensors.rs
@@ -185,10 +185,7 @@ pub fn read_tensor(mut reader: impl Read) -> TractResult<Tensor> {
             let mut bytes = Vec::new();
             (&mut reader).take(len as u64).read_to_end(&mut bytes)?;
             if bytes.len() != len as usize {
-                bail!(
-                    "NNEF string item declared {len} bytes but only {} readable",
-                    bytes.len()
-                );
+                bail!("NNEF string item declared {len} bytes but only {} readable", bytes.len());
             }
             *item = String::from_utf8(bytes)?;
         }
--- a/nnef/src/tensors.rs
+++ b/nnef/src/tensors.rs
@@ -176,12 +176,20 @@ pub fn read_tensor(mut reader: impl Read) -> TractResult<Tensor> {
         let mut plain = tensor.try_as_plain_mut()?;
         for item in plain.as_slice_mut::<String>()? {
             let len: u32 = reader.read_u32::<LE>()?;
-            let mut bytes = Vec::with_capacity(len as usize);
-            #[allow(clippy::uninit_vec)]
-            unsafe {
-                bytes.set_len(len as usize);
-            };
-            reader.read_exact(&mut bytes)?;
+            // SECURITY: `len` is read from the (untrusted) NNEF file. Do NOT pre-allocate or
+            // `set_len` an unbounded buffer from it (CWE-770): a malicious file advertising a
+            // multi-gigabyte string length forces a huge *upfront* allocation (abort on OOM, or
+            // memory exhaustion) before any byte is read, and the previous `unsafe set_len` also
+            // left the buffer uninitialized. Read at most `len` bytes, growing the buffer from what
+            // actually arrives, then verify the real length matches the declaration.
+            let mut bytes = Vec::new();
+            (&mut reader).take(len as u64).read_to_end(&mut bytes)?;
+            if bytes.len() != len as usize {
+                bail!(
+                    "NNEF string item declared {len} bytes but only {} readable",
+                    bytes.len()
+                );
+            }
             *item = String::from_utf8(bytes)?;
         }
         Ok(tensor)
--- a/nnef/tests/nnef_string_len_alloc.rs
+++ b/nnef/tests/nnef_string_len_alloc.rs
@@ -21,7 +21,8 @@ static MAX_SINGLE_ALLOC: AtomicUsize = AtomicUsize::new(0);
 unsafe impl GlobalAlloc for CountingAllocator {
     unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
         let size = layout.size();
-        let _ = MAX_SINGLE_ALLOC.fetch_update(Ordering::Relaxed, Ordering::Relaxed, |c| Some(c.max(size)));
+        let _ = MAX_SINGLE_ALLOC
+            .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |c| Some(c.max(size)));
         System.alloc(layout)
     }
     unsafe fn dealloc(&self, ptr: *mut u8, layout: Layout) {
@@ -76,7 +77,9 @@ fn string_item_len_must_not_drive_unbounded_alloc() {
     let _ = read_tensor(Cursor::new(bytes));
 
     let max_alloc = MAX_SINGLE_ALLOC.load(Ordering::Relaxed);
-    println!("largest single allocation during read_tensor = {max_alloc} bytes (malicious declared len = {DECLARED_LEN})");
+    println!(
+        "largest single allocation during read_tensor = {max_alloc} bytes (malicious declared len = {DECLARED_LEN})"
+    );
     assert!(
         max_alloc <= MAX_ACCEPTABLE_SINGLE_ALLOC,
         "read_tensor pre-allocated {max_alloc} bytes from an untrusted string length (CWE-770)"
--- a/nnef/tests/nnef_string_len_alloc.rs
+++ b/nnef/tests/nnef_string_len_alloc.rs
@@ -1,0 +1,84 @@
+//! Regression test for CWE-770 in NNEF string-tensor loading.
+//!
+//! `read_tensor` used to read a per-item `len: u32` straight from the (untrusted) NNEF file and
+//! then `Vec::with_capacity(len)` + `unsafe { set_len(len) }` — a malicious file advertising a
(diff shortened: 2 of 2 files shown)


kali (dismissed):

kali:
Thanks.

li-jin-quan:
Hi @kali, thanks again for the quick merge.

I'd like to ask whether this one warrants a security advisory. Some context I gathered while preparing the write-up:

- The unbounded `Vec::with_capacity(len)` + `unsafe { set_len(len) }` in the `DatumType::String` branch of `read_tensor` was introduced in **0.12.2** (2021-01-05): v0.12.1 is clean, v0.12.2 already carries it.
- It is still present in the latest release of **all three maintained lines** — `0.21.17`, `0.22.3`, `0.23.6`. That is roughly 5 years 8 months and ~90 releases.
- 0.23.6 shipped on 09-02 and this merged on 09-03, so **no released version contains the fix yet** — downstream users are still exposed.
- Measured with a counting global allocator: a 3-byte payload declaring a 128 MiB string length forces a single **134,217,728-byte (128 MiB)** allocation, versus **102 bytes** after the fix.
- This is distinct from RUSTSEC-2026-0217 / CVE-2026-55093. That one is an integer overflow in the shape/dimension handling (fixed in 0.23.1); this is an unbounded allocation in the String branch and is still reproducible on 0.23.6, so it is not covered by that advisory.
- Impact is availability only. I'd suggest `CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:H` (6.5 Medium) to stay consistent with RUSTSEC-2026-0217's AV:L / UI:R treatment; a service that accepts user-supplied models would be AV:N / 7.5.

Would you like me to open a draft GHSA on this repo, or would you prefer to file it yourself? Happy to supply the full write-up, the PoC and a ready-to-submit RustSec TOML. I can also help with backports to the 0.21 and 0.22 lines if that is useful.

Reporter credit: Li Jinquan — github.com/li-jin-quan

kali:
Hey ! If you want the CVE, I'll play along. TBH, I think it's pretty low value for tract, most people use ONNX, nobody in their right mind accept models they don't control. The bug is fixed on the main tree, the rest is mostly churn from my POV. But I do value independent security research, so make it easy for me, I'll do my part, I'll backport the fix and do the releases.

li-jin-quan:
Hi @kali, perfect — thank you. Since I'll be the one getting the credit anyway, let me make this as close to copy-paste as I can. Everything below is API-verified.

**Suggested GHSA title**

```
Unbounded allocation in tract-nnef NNEF tensor parser allows denial of service on untrusted model load (CWE-770)
```

**Suggested description**

> `tract_nnef::tensors::read_tensor` parses NNEF tensor data. For tensors of `DatumType::String`, it reads a 32-bit length prefix directly from the (untrusted) NNEF file and allocates that many bytes **before reading any of them** (`Vec::with_capacity(len)` + `unsafe { set_len(len) }`, then `read_exact`). A crafted NNEF file can therefore declare a string length anywhere in the `u32` range (up to 4 GiB) while supplying only a handful of bytes: the huge allocation happens first and only then does `read_exact` fail.
>
> Impact is availability only (memory exhaustion / process abort) whenever an application loads an NNEF model from an untrusted source — e.g. a model fetched from a public hub, received from a third party, or uploaded by an end user of a hosted service. Measured with a counting global allocator: a 3-byte payload declaring a 128 MiB string length forces a single **134,217,728-byte (128 MiB)** allocation, versus **102 bytes** after the fix — roughly a 4×10⁷ amplification from a 3-byte input.
>
> The fix (PR #2766, merge commit `cd14c552`) reads at most `len` bytes via `take(len).read_to_end`, verifies the real length against the declaration, and as a side effect removes the `unsafe { set_len(..) }` block (`clippy::uninit_vec`). No public API change. Not to be confused with RUSTSEC-2026-0217 / CVE-2026-55093, which is an integer overflow in shape/dimension handling (fixed in 0.23.1); this one is still reproducible on 0.23.6.

**Form fields**

- Package: `tract-nnef`
- Affected: `>= 0.12.2, <= 0.23.6` (introduced in v0.12.2, 2021-01-05 — v0.12.1 is clean; the latest releases of all three maintained lines, `0.21.17` / `0.22.3` / `0.23.6`, are all affected)
- Patched: whatever you cut from main (I'll fill in the exact versions once you release)
- CWE-770 (Allocation of Resources Without Limits or Throttling); CWE-400 as secondary
- Severity: `CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:H` — 6.5 Medium, kept consistent with RUSTSEC-2026-0217's AV:L / UI:R treatment. (A service accepting user-uploaded models would justify AV:N / 7.5, but I'd rather stay conservative.)
- References: PR #2766, commit `cd14c552f05f6142587cf5a2003d32decab89cd5`
- Reporter credit: Li Jinquan — github.com/li-jin-quan

**Division of labor** — you create the draft GHSA on this repo (that permission is maintainer-only) and GitHub will request the CVE for you; once the IDs exist, I'll prepare and submit the RustSec advisory PR myself so there's zero extra work on your side beyond the releases/backports you already offered. If you'd rather I put the full write-up + PoC into a gist or the advisory body, say the word.

Thanks again for handling the backports and releases.

kali:
i think you can submit the draft through the github interface actually.

PR #2768, opened 2026-09-03
harness: the causal-conv1d-update case declared input [1, 1, 4] and s…

kali:
…tate [4, 4], which the op rejects since it reads input as [b, C, S] and state as [b, C, k], and the weight [4, 4] pins C and k at 4. Declare input [1, 4, 1] and state [1, 4, 4] and reshape the bundle to match; the recorded output and final state already follow that reading, so the values stand.


kali:
ping @JulienBalianSonos . can you check this ? the check only run on the nightly PR so we missed it.

JulienBalianSonos (approved):
Yes: the old fixture was invalid against the op's own contract (state must be [b, C, k], rank 3 :  the old `state = [4, 4]` was rank 2, and the old `input = [1, 1, 4]` had C=1 mismatching the weight's C=4). The new shapes (`input = [1, 4, 1]`, `state = [1, 4, 4]`) line up with what causal_conv1d_update actually asserts. This is exactly the case this harness check exists for.
Thanks for catching it, sorry it slipped through on the original PR since nightly-only coverage missed it there.

PR #2770, opened 2026-09-03
RMSNorm residual-absorption and scaled-norm fusion

JulienBalianSonos:
Split out of #2084 (see the dismantling plan; #2761 was the first split, already merged).

- `ScaledRmsNorm`: absorbs a slice-split gamma into one fused scaled norm.
- Residual-add absorption into the RMSNorm kernel dispatch, on both Metal
  and Cuda.
- Cast folding around the norm (in/out), with the redundant standalone
  downcast-roundtrip rule dropped once the residual/scale fusion covers
  the cases it existed for.
- Ports across core (op definition), NNEF (ser/deser), and the gpu/cuda/
  metal kernels.
- End-to-end fusion tests (`RmsNormFusionProblem`) that run a real
  Add/RmsNorm/Mul graph through the actual Metal/Cuda transform
  pipeline, on top of the existing kernel-level tests.

## Benchmarks

Before this PR, a residual add feeding an RMSNorm followed by a learned
gamma multiply was three separate kernel dispatches per layer (Add,
RmsNorm, Mul); this PR collapses that into one. 32 stacked norm layers,
hidden size 4096:

| Backend | Shape | Before | After | Speedup |
|---|---|---|---|---|
| Metal | Decode, b=1 s=1 | 915 µs | 625 µs | 1.46x |
| Metal | Prefill, b=1 s=32 | 2.08 ms | 1.43 ms | 1.45x |
| Cuda | Decode, b=1 s=1 | 3.08 ms | 2.21 ms | 1.39x |
| Cuda | Prefill, b=1 s=32 | 3.13 ms | 2.26 ms | 1.38x |

Reproduce with `cargo bench -p tract-metal --bench rms_norm_fusion` /
`cargo bench -p tract-cuda --bench rms_norm_fusion`.

🍍


kali:
Unknown `/ci` target: `example`.

Usage: `/ci full`, `/ci large-models` (aliases: `large`, `llm`), `/ci wheels` (aliases: `python`, `py`), `/ci examples` (alias: `ex`), `/ci bench` (alias: `benches`), or `/ci bench-mt-ladder [ladder]` (default ladder `2,4,0`).

kali:
Kudos, this is looking pretty good.

kali:
Found two precision bugs, fixed them, the third one is more tricky. WDYT ?
https://claude.ai/code/artifact/fa14487c-a733-4d1e-bd96-b18cc6c2464a

kali:
Please ignore the preprocessor benches.

JulienBalianSonos:
Thanks for the fixes.
About point 3, I agree, and I would go with Option A over B. 

I do not think we need a new field though. Both fuse_rms_norm_scale and fuse_rms_norm_split_scale match directly off a plain RmsNorm node, and that op's output_facts hardcodes its output dtype to match its input dtype. So the pre-scale rounding target is always exactly in_dt, not something separate that needs recording. The simpler fix is to round the f32 accumulator to InT and back right before the scale multiply in both kernels, the same pattern used in the residual fix, with no new op field or NNEF change. 

That keeps the fusion with zero extra dispatches.

PR #2771, opened 2026-09-03
(perf) core/cnn: NCHW spatial kernels (direct conv, maxpool, nearest, deconv)

czoli1976:
Draft for @TarikAlHadethi to try on the i7-12700K against PP-OCRv6. Includes #2735 (depthwise along W) plus the rest of the NCHW spatial path.

## What this is

Small-`k` group-1 2-D conv stays a spatial FIR along W (`DirectSpatialConv`, `k = Cin·kh·kw ≤ 64`) instead of im2col GEMM. Shared along-W kernel is NEON and AVX2 FMA. Also:

- NCHW MaxPool 2×2 along W
- integer nearest upsample in one pass (no `Tile` → `MultiBroadcastTo`)
- 2×2 stride-2 deconv unpack along W
- Relu fused into DirectSpatial / DepthWise; channel-scale Mul fused when the other input is `[N,C,1,1]`
- DirectSpatial **off on wasm** (would replace simd128 GEMM with a scalar FIR)

## Apple (M1 Pro, 1 thread, this branch)

PP-OCRv6 Hub ONNX, `RAYON_NUM_THREADS=1`, ORT 1.29 CPU EP `intra=1`.

| model | tract | ORT p50 | tract/ORT |
|---|---:|---:|---:|
| tiny_det `1×3×640×640` | **61.6 ms** | ~66 | **0.93×** |
| tiny_rec `1×3×48×320` | **4.8 ms** | 5.6 | **0.85×** |

tiny_det was ~2× ORT at the start of this work. Same input vs ORT: max abs **2.4e-5** det / **5e-6** rec (ops are `Validation::Rounding`, not bit-identical).

## What should move on x86

Your tiny_det profile was GEMM-bound, but these nodes are not the AMX-shaped 1×1s:

| op (your 318 ms profile) | why this PR |
|---|---|
| MultiBroadcastTo 37.6 ms | nearest upsample + fast broadcast copy |
| DepthWiseConv 69 ms | AVX2 along-W, stride 1 |
| OptMaxPool 10.7 ms | AVX 2×2 along W |
| Conv.1 / Conv.2 im2col (`8×64×102400`, `16×32×102400`) | DirectSpatial, AVX2 oc4, **s=1** |
| Conv.0 3×3 **s=2** | DirectSpatial, but AVX2 oc4 is **scalar at stride 2** |
| DeconvSum 7.2 ms | W-inner unpack (scalar on x86, still not C-inner) |

Will **not** close 6.7×. Leftover is still packing / huge-N skinny-K FMA (and the picker misses). Rec here is AMX; yours will stay GEMM.

## How to run

Hub `PaddlePaddle/PP-OCRv6_tiny_{det,rec}_onnx`. Both ignore flags (value_info batch vs concrete `-i`):

```
set RAYON_NUM_THREADS=1

tract -O inference.onnx --onnx-ignore-value-info --onnx-ignore-output-shapes -i 1,3,640,640,f32 bench -R --warmup-loops 3 --max-loops 10
tract -O inference.onnx --onnx-ignore-value-info --onnx-ignore-output-shapes -i 1,3,48,320,f32 bench -R --warmup-loops 8 --max-loops 40
```

Dump should show `DirectSpatialConv` (3 nodes on tiny_det: Conv.0/1/2), `NearestUpsample`, no `Resize.*.tile` MultiBroadcastTo.

ORT: same shapes, CPU EP only, `intra_op=1`, `inter_op=1`, `ORT_ENABLE_ALL`.

Please send back: the 6-row table if you can, or at least tiny_det + tiny_rec, and the tiny_det “Most time consuming operations” block.


TarikAlHadethi:
Ran it on the i7-12700K, on `4ee49e4`. Windows, `RAYON_NUM_THREADS=1`, both flags as you gave them.

Baseline is the merge-base `722b9ce`, built and measured on the same machine tonight rather than my older 318 ms figure, so before and after are the same binary settings on the same day.

### It still does not build on x86_64

`cargo build --release -p tract-cli` fails on this branch:

```
error[E0432]: unresolved import `std::arch::aarch64`
  --> linalg/src/arm64/arm64simd/erf.rs
```

`erf.rs` uses `std::arch::aarch64` with no `cfg` guard (two places on the current head). That is fine on aarch64, but `linalg/src/lib.rs:69` compiles the whole `arm64` tree when `foreign-inventory` is on:

```rust
#[cfg(any(target_arch = "aarch64", feature = "foreign-inventory"))]
pub mod arm64;
```

and `cli/Cargo.toml:88` turns that feature on for the CLI. So the CLI does not compile on the architecture you asked me to test. The merge-base builds clean, so it comes in with this PR.

I got the numbers by gating the two lines in `arm64simd/mod.rs`:

```rust
#[cfg(target_arch = "aarch64")]
mod erf;
#[cfg(target_arch = "aarch64")]
pub use erf::arm64simd_erf_f32_8n;
```

Nothing outside `erf.rs` references `arm64simd_erf_f32_8n`, so that is the whole of it. I can open it separately if that is easier than folding it in.

### Numbers

ORT 1.29.0, CPU EP only, `intra_op=1`, `inter_op=1`, `ORT_ENABLE_ALL`, p50.

| model | base | this PR | ORT p50 | base/ORT | PR/ORT |
|---|---|---|---|---|---|
| tiny_det 1x3x640x640 | 298.324 ms | **199.515 ms** | 52.106 ms | 5.73x | **3.83x** |
| tiny_rec 1x3x48x320 | 12.750 ms | **11.682 ms** | 4.308 ms | 2.96x | **2.71x** |

tiny_det 1.50x, tiny_rec 1.09x.

Worth flagging: on `2d8d6c8` I measured tiny_rec at 12.757 ms, which is nothing at all against the 12.750 baseline. The scale fuse in `4ee49e4` is what moved it. So rec does benefit here, just not from the spatial kernels.

### tiny_det, most time consuming operations

| op | base | this PR | delta |
|---|---|---|---|
| OptMatMul | 114.118 | 99.943 | -14.2 |
| OptMatMulPack | 25.518 | 27.637 | +2.1 |
| DepthWiseConv | 55.755 | 17.398 | **-38.4** |
| MultiBroadcastTo | 38.966 | - | **gone** |
| Pad | 16.968 | 10.378 | -6.6 |
| OptMaxPool | 9.416 | 3.134 | -6.3 |
| DeconvSum | 5.961 | - | gone |
| Erf | 4.023 | 5.972 | +1.9 |
| DirectSpatialConv | - | 30.228 | new |
| NearestUpsample | - | 2.479 | new |

Every op you predicted would move, moved. MultiBroadcastTo leaves the profile entirely, DepthWiseConv drops 38.4 ms, and DirectSpatialConv costs 30.2 ms against the im2col GEMM it replaces.

You were right that it would not close the gap. `OptMatMul` plus `OptMatMulPack` is 127.6 ms, which tract's own profile puts at **56% of the run** (43.9% + 12.1%), so the remaining distance to ORT is where you said it was. `OptMatMulPack` is up 2.1 ms and `Erf` up 1.9 ms.

### Structural checks

- `DirectSpatialConv`: 3 nodes, on `Conv.0`, `Conv.1`, `Conv.2`
- `NearestUpsample`: 6 nodes, zero `Resize.*.tile`
- One `MultiBroadcastTo` node survives in the graph, but it is `ConvTranspose.0.broadcast_bias` and no longer shows in the timing table

Say if you want the rec profile or another shape.

czoli1976:
Pushed `3973c01` for the two leftovers your dump showed.

AVX2 along-W now does stride 2/3 (oc4 too), so Conv.0 should not be scalar. Without AMX, DirectSpatial takes fat 3×3 up to k=576 — dump should show more than the three stem nodes. Apple still keeps those on im2col.

Checked under qemu-system-x86_64 Haswell (AVX2+FMA): fat 3×3 lowers to DirectSpatial and matches the reference; stem s=2 and DW s=2/3 pass.

CLI should also build without the erf cfg patch (`bf9c63c`).

TarikAlHadethi:
Ran `3973c01` on the i7-12700K, same setup as before.

**The CLI builds clean with no patch**, so `bf9c63c` does it.

The dump shows **5 DirectSpatialConv nodes**, not 3: `Conv.0`, `Conv.1`, `Conv.2`, `Conv.3`, `Conv.82`. So the fat 3x3s do lower here without AMX, as you expected.

### Numbers

| model | base `722b9ce` | `3973c01` | ORT p50 | base/ORT | now/ORT |
|---|---|---|---|---|---|
| tiny_det 1x3x640x640 | 298.324 ms | **181.675 ms** | 52.106 ms | 5.73x | **3.49x** |
| tiny_rec 1x3x48x320 | 12.750 ms | **11.170 ms** | 4.308 ms | 2.96x | **2.59x** |

tiny_det is 1.64x now, tiny_rec 1.14x.

Every commit has moved it on this machine:

| | tiny_det |
|---|---|
| base `722b9ce` | 298.324 ms |
| `2d8d6c8` | 224.660 ms |
| `4ee49e4` | 199.515 ms |
| `3973c01` | 181.675 ms |

### tiny_det profile

| op | `4ee49e4` | `3973c01` | delta |
|---|---|---|---|
| OptMatMul | 99.943 | 74.991 | **-25.0** |
| DirectSpatialConv | 30.228 | 33.412 | +3.2 |
| OptMatMulPack | 27.637 | 25.392 | -2.2 |
| DepthWiseConv | 17.398 | 13.485 | -3.9 |
| OptMaxPool | 3.134 | 2.645 | -0.5 |
| Erf | 5.972 | 4.067 | -1.9 |

`OptMatMul` gives up 25 ms for 3.2 ms on `DirectSpatialConv`, which is the two extra convs coming off im2col. `Pad` is out of the profile entirely and `DeconvSum` is down to 1.658 ms.

`OptMatMul` plus `OptMatMulPack` is 100.4 ms, 54.6% of the run, so the remaining distance to ORT is still packing and skinny-K as you said. Down from 56% but it is the same shape.

Happy to run anything else you want checked.

czoli1976:
@kali a ride on the horse for this one please

kali:
Happy to put a number on the progress, but this will need work before it merges.

TarikAlHadethi:
Reproduced all three red checks locally at 3973c01. Host is Windows x86_64, so the wasm side is a cross build.

fmt

One line, in the new AVX2 path. core/src/ops/cnn/conv/along_w.rs:223:

let x = avx2_load8(iptr.offset(ioffset[n]).offset(i as isize * in_stride), in_stride);
cargo fmt wraps it across three lines. Nothing else in the tree is dirty.

test-nnef-cycle

2465 passed, 1 failed:

---- nnef_cycle::onnx::node::v1_13_0::test_resize_upsample_scales_nearest stdout ----
Error: Translating model to proto_model
Caused by:
    0: Translating model to AST
    1: translating node #1 "Y.nearest" NearestUpsample
    2: No serializer found for node #1 "Y.nearest" NearestUpsample
The one pass integer nearest upsample lowers to a NearestUpsample node that has no NNEF serializer, so any model containing it stops round tripping through NNEF.

wasm32-wasi

This one is not a build break. Every step of .travis/cross.sh wasm32-wasi builds clean for wasm32-wasip1 here:

cargo check --target wasm32-wasip1 --features getrandom-js -p tract-onnx -p tract-tensorflow
cargo build --target wasm32-wasip1 --tests -p tract-linalg -p tract-core -p test-unit-core with RUSTFLAGS='-C target-feature=+simd128'
the same with +simd128,+relaxed-simd
Those are full codegen builds, not cargo check, so intrinsics are actually being lowered. That points at a run time failure under wasmtime rather than a compile guard, which makes it a different problem from the x86_64 one earlier in this PR. I do not have wasmtime set up locally to name the failing test, but I can if that is useful.

czoli1976:
> 🔴 **Bench vs main — 27 speed regression(s)** · ⚠️ 1 secondary
> 
> _Reference: 2026-09-04 morning nightly run (0d old) · full report → [run](https://github.com/sonos/tract/actions/runs/33843793820)_
> 
> **Speed** — evaltime · prefill · decode
> 
> Δ	metric	device	main → PR
> 🔴 **+79.4%**	mobilenet_v2_1
> evaltime · pass_mt	`cortex-a53`	516 ms → 926 ms
> 🔴 **+77.6%**	mobilenet_v1_1
> evaltime · pass	`cortex-a53`	437 ms → 775 ms
> 🔴 **+62.9%**	mobilenet_v2_1
> evaltime · pass	`cortex-a53`	647 ms → 1.05 s
> 🔴 **+61.1%**	mobilenet_v1_1
> evaltime · pass	`cortex-a55`	291 ms → 470 ms
> 🔴 **+50.0%**	mobilenet_v2_1
> evaltime · pass_mt	`cortex-a7`	938 ms → 1.41 s
> +22 more regression(s)
> Δ	metric	device	main → PR
> 🔴 **+40.0%**	mobilenet_v1_1
> evaltime · pass	`cortex-a7`	1.11 s → 1.55 s
> 🔴 **+34.9%**	inceptionv1q
> evaltime · pass	`apple-m1-max`	39.6 ms → 53.4 ms
> 🔴 **+32.4%**	mobilenet_v2_1
> evaltime · pass	`cortex-a7`	1.49 s → 1.97 s
> 🔴 **+32.1%**	mobilenet_v1_1
> evaltime · pass	`orangepi-rv2`	413 ms → 546 ms
> 🔴 **+30.6%**	mobilenet_v2_1
> evaltime · pass_mt	`cortex-a55`	352 ms → 460 ms
> 🔴 **+25.6%**	mobilenet_v2_1
> evaltime · pass_mt	`cortex-a9`	956 ms → 1.2 s
> 🔴 **+24.6%**	mobilenet_v2_1
> evaltime · pass_mt	`beaglev-ahead`	503 ms → 627 ms
> 🔴 **+23.6%**	mobilenet_v2_1
> evaltime · pass	`cortex-a55`	419 ms → 518 ms
> 🔴 **+18.4%**	inceptionv1q
> evaltime · pass	`i9-11900kb_rtx-4060`	49.4 ms → 58.4 ms
> 🔴 **+16.3%**	mobilenet_v1_1
> evaltime · pass	`beaglev-ahead`	912 ms → 1.06 s
> 🔴 **+14.6%**	mobilenet_v2_1
> evaltime · pass	`cortex-a9`	1.46 s → 1.68 s
> 🔴 **+14.2%**	inceptionv1q
> evaltime · pass	`cortex-a7`	1.64 s → 1.87 s
> 🔴 **+13.9%**	inceptionv1q
> evaltime · pass	`orangepi-rv2`	754 ms → 859 ms
> 🔴 **+13.0%**	mobilenet_v2_1
> evaltime · pass	`apple-m1-max`	31.1 ms → 35.2 ms
> 🔴 **+12.6%**	mobilenet_v1_1
> evaltime · pass	`apple-m1-max`	18.1 ms → 20.4 ms
> 🔴 **+12.6%**	mobilenet_v2_1
> evaltime · pass_mt	`orangepi-rv2`	514 ms → 579 ms
> 🔴 **+11.9%**	mobilenet_v2_1
> evaltime · pass	`beaglev-ahead`	1.15 s → 1.28 s
> 🔴 **+11.6%**	inceptionv1q
> evaltime · pass	`cortex-a9`	1.94 s → 2.16 s
> 🔴 **+11.0%**	mobilenet_v1_1
> evaltime · pass	`cortex-a9`	1.04 s → 1.15 s
> 🔴 **+9.3%**	inceptionv1q
> evaltime · pass	`cortex-a53`	915 ms → 1 s
> 🔴 **+9.2%**	mobilenet_v2_1
> evaltime · pass	`orangepi-rv2`	643 ms → 702 ms
> 🔴 **+7.3%**	inceptionv1q
> evaltime · pass	`cortex-a55`	629 ms → 675 ms
> ⚠️ 1 secondary regression(s)

Nice, very indicative

czoli1976:
Pushed five commits. Each of the 27 bench regressions traced to a mechanism; all
of them the same shape — a lowering that wins in one regime, taken in every regime.

**`OptMaxPool`** — `try_nchw_2d` returned `Some` for every 2-D NCHW pool, so
everything that is not 2×2 f32 went down a naive scalar walk with a bounds test
per tap, replacing `visit_output`'s zone partition. Falls through now.
Covers the 7 `inceptionv1q` rows.

**`DepthWiseConv`** — the channel-scale work put three loop-invariant decisions
in the innermost per-element loop (the relu flag, whether a scale exists, the
scale value), and applied the scale *through the output pointer*: store → reload
→ multiply → store, per element. That stopped the const-N body vectorising —
which is the path NHWC models take, since they never reach the along-W kernel at
all. Hoisted, and a constant per-channel scale now folds into the weights
(`relu(conv(x, w) + b)·s == relu(conv(x, w·s) + b·s)`, `s > 0`), so no scale
survives to eval. Covers the 20 mobilenet rows.

**`DirectSpatialConv`** — `can_direct_spatial` had no shape term, so it also
claimed short rows, where the per-zone tap repack cannot amortise. Gated on the
output row.

**Also**: SIMD gate is a runtime check on x86_64 (AVX2 is not baseline, so a
`cfg` let a pre-AVX2 CPU into the scalar fallback the gate exists to avoid);
`MIN_ALONG_W_LEN` dropped, nothing measured supported it; `NearestUpsample`
lowered at codegen, since NNEF round-trips the decluttered model and has no
serializer for it.

Measured here, apple m1 pro, 1 thread, against the merge-base:

| | was | now |
|---|---|---|
| inceptionv1q | +34.9% | **−10.6%** |
| mobilenet_v1 | +12.6% | **−2.3%** |
| mobilenet_v2 | +13.0% | **−3.8%** |
| mobilenet_v2 `pass_mt` | | **−15.7%** |
| PP-OCRv6 tiny_det | 171.6 ms | **60.0 ms** (2.86×) |

Speech-enhancement models are clean (GTCRN, DTLN, FastEnhancer). DFN3's encoder
keeps a **~1%** residual against the merge-base — real, but the within-run
spread on that harness is ~5%, and it is not `DirectSpatialConv`: disabling that
op entirely leaves the same ~1%.

`cargo test --workspace` (macOS, less `test-tflite`/`tract-proxy*`/`test-cuda`):
109 suites, 70160 passed, 70 failed — all 70 pre-existing, none in conv, pooling,
resize or linalg. The wasm job will **stay red** on `rms_norm::tests::eval_with_f16_eps_and_f16_input`,
which fails identically on clean `origin/main` (4bcf90f) — not this PR.

Three CI failures fixed: fmt, `test-nnef-cycle` (2466/0), and the wasm depthwise
test, which asserted `DepthWise` where the blocked convolution owns that lowering
on wasm and aborted the whole run before anything else could report.

@TarikAlHadethi — would you re-run the i7 numbers? Especially whether tiny_det
holds and whether the `OptMatMulPack` / `Erf` lines moved.

@kali — could you kick off a bench-vs-main? The cortex-a53/a55/a7/a9 and riscv
rows are the ones I cannot measure here; the depthwise defect above is
architecture-independent and in-order cores should have suffered most from it,
but that is inference until the fleet says otherwise.

czoli1976:
solving conflicts first

czoli1976:
@kali let's go again
image

czoli1976:
Pushed `3b651fe` — three fixes from a review pass over this branch. Two are
correctness, and one changes which pools take the new path, so the numbers in my
previous comment predate it.

**relu fused in the wrong order relative to a channel scale.** `fuse()` took a
trailing Relu onto an op that had already absorbed a `[1,C,1,1]` Mul, but eval
applies relu *before* the scale. So `Conv -> Mul -> Relu` — the unfolded
Conv-BN-ReLU order — computed `relu(conv(x)) * s` where the graph means
`relu(conv(x) * s)`. Those differ wherever a scale is negative, which BatchNorm
gamma can be: at `s = -2` over a conv output of `-1`, the graph says 2 and the op
said 0. Both `DirectSpatialConv` and `DepthWise` decline the relu now once a
scale is carried. DepthWise's constant fold is unaffected — it bakes the scale
into the weights while `relu` is still false, which is why this only ever bit the
runtime-scale path and DirectSpatial, which has no fold.

**The along-W stride-2/3 blocks read past the input.** A de-interleaving load
takes a whole register: `vld2q` spans eight floats to use four, `vld3q` spans
twelve to use four, and the AVX2 stride-2 pair spans sixteen to use eight — while
the caller only guarantees `(len-1)*in_stride`. On the last block of the last row
of the last channel that is one to two f32 past the tensor. Silent in practice,
since the allocator has slack, but it is UB and would fault at a page boundary or
under ASAN. Each vector block now reserves one more output of slack and the scalar
tail takes the remainder. Stride 1 loads exactly what it consumes and is
unchanged, as is the AVX2 stride-3 arm, which gathers exact indices. It costs the
last few outputs of each stride-2/3 row; widening the contract to demand
caller-side padding would be the faster fix but a much larger change.

**The 2×2 maxpool gate admitted any stride.** `maxpool_2x2_f32` only vectorises
stride 1, so plain 2×2/2 downsampling was taking the scalar per-tap
bounds-tested walk rather than `visit_output` — the same shape that made the 3×3
pools regress inceptionv1q. The gate requires stride 1 now, as the comment above
it already claimed. **This is the one that moves numbers**: any model with an
ordinary 2×2/2 MaxPool goes back to `visit_output`. I expect that to be neutral
or better, but I have not measured it, so please treat the bench run as the
authority rather than my earlier table.

`cargo test --workspace` on macOS (less `test-tflite`, `tract-proxy*`,
`test-cuda`) is 109 suites / 70160 passed / 70 failed, all 70 pre-existing and
none in conv, pooling, resize or linalg.

One thing I could not check here: the AVX2 halves of the second fix are
inspection-only, because cross-compiling x86_64 fails in `tract-linalg`'s build
script on this host for want of a cross assembler. @TarikAlHadethi if you are
re-running anyway, the stride-2/3 along-W paths are the ones worth a correctness
look as well as a timing one.

czoli1976:
@kali when you next look at this — could you approve the workflow run for
`bc3b914` when you get a chance? Everything since `3b651fe` is sitting at
`action_required`, so neither the embedded targets nor the bench have started.

The one run that did execute caught a real miss of mine, now fixed in `bc3b914`.
Eight `direct_spatial::tests` failed on armv6, armv7 and riscv64: they asserted
DirectSpatialConv was present unless the target was wasm, which was right while
wasm was the only exclusion, but `can_direct_spatial` now declines wherever
there is no along-W SIMD kernel. The tests ask `has_simd_kernel()` now — the
same gate the lowering asks — rather than restating one case of it. Locally
that is direct_spatial 8/8 and tract-core 316/0, but armv7 and riscv are
precisely what I cannot run here, so CI is the check that matters.

The bench is the other thing I am waiting on. My numbers are single-machine
(apple m1 pro) and they predate the 2×2 maxpool stride fix in `3b651fe`, which
sends ordinary 2×2/2 downsampling back to `visit_output` — so please treat the
fleet run as the authority over my earlier table, particularly for the
cortex-a53/a55/a7/a9 and riscv rows, which I have no hardware for.

czoli1976:
I can see the prize now ....

czoli1976:
@kali sorry to ask again — could you approve a run on `de46085` when you have a
moment? Thank you for the last one; it was worth far more than the local numbers
I had, and it found something I would not have.

The bot showed the split clearly: mobilenet gained on the out-of-order cores
(i9 −13 to −23%, m1-max −6 to −13%) and lost badly on the in-order ones,
cortex-a53 up to +57%. That is one dispatch change. This branch added 9 and 25
arms, so a 3×3 depthwise — every mobilenet depthwise layer — moved off
`visit_output` onto the const-N monomorph.

What I had missed is that mobilenet is NHWC, so none of that work could help it
in the first place: the inner axis is C, not W, so the monomorphs cannot
vectorise and the along-W handoff cannot fire. It was pure cost, and the small
cores paid the most. `de46085` gates all of it on a contiguous output W, which
puts NHWC back on main's exact path — same `visit_output` dispatch, and
`inner_loop` back to the bare store, since nothing reaching it can carry a relu
or a scale any more. The two fusions are gated with it, or `inner_loop` keeps
paying per output element.

What is left over main there is the constant scale fold, which removes an
`OptMulUnicast` outright, so the NHWC path now does strictly less work than main
on identical kernels. On an m1 pro that is mobilenet_v1 18.385 → 17.703 and
mobilenet_v2 31.467 → 30.763 ms/i, but the structural point matters more than my
one machine: it should not be able to lose on a53 either. That is the claim I
would like the fleet to check.

On the other three rows, and I would rather tell you what I could not reproduce
than let them stand:

- `parakeet_…_preprocessor` +45%: not reproducible here. main and this branch
  are 0.441/0.440/0.440 against 0.439/0.441/0.440 ms over ~11,300 iterations
  each. Either x86-specific — I only have aarch64 — or noise on a 0.3 ms
  workload.
- `nemotron_…_preprocessor` load +19.4%: eval is identical (0.397 vs 0.397), and
  nothing here touches NNEF deserialisation.
- `en_tdnn_15M` RSS +6.3%: this one cannot be us. The optimised graph carries no
  DepthWiseConv, DirectSpatialConv, MultiBroadcastTo or MaxPool, so no code in
  this PR runs on it at all.

No rush on any of this, and thank you for the CI time.

kali:
the preprocessors are all over the place, i'm making progress, but they're reacalcitrant. The tdnn is more worrying. I have found recently multiple occurencies of compilation variability triggered by unrelated areas of the code. If this one keep showing up, I can investigate it myself with direct access to devices.

czoli1976:
> the preprocessors are all over the place, i'm making progress, but they're reacalcitrant. The tdnn is more worrying. I have found recently multiple occurencies of compilation variability triggered by unrelated areas of the code. If this one keep showing up, I can investigate it myself with direct access to devices.

Yeah I am often puzzled by these, device access makes the difference here

czoli1976:
Thank you — 27 → 16 → 1. The mobilenet rows are all gone, including cortex-a53,
which is the one I could not measure and was least sure of.

On the row that is left, `parakeet_…_preprocessor_1s` +47% on the i9, I have
something concrete but not a fix, and I would rather show the evidence than
guess again — I have already been wrong about this row twice, first calling it
noise and then blaming the new `fast_broadcast`.

**No code from this PR runs on that model.** Its optimised graph is STFT 41.6%,
Pad 18.5%, Reduce<Sum>, Iff, OptMatMul, Slice, Ln. Grepping it for every op this
PR can produce — DepthWiseConv, DirectSpatialConv, MultiBroadcastTo, MaxPool,
NearestUpsample, DeconvSum, Resize — matches zero. The six `tile`s I suspected
fold away during optimisation, so `MultiBroadcastTo` never even exists, which
kills the `fast_broadcast` theory.

On aarch64 the two binaries are indistinguishable: 0.441/0.440/0.440 against
0.439/0.441/0.440 ms over ~11,300 iterations each.

So the +47% is real and repeatable but cannot be this PR's logic executing. What
fits both facts is code layout: ~3000 lines land in tract-core and shift
placement, which moves I-cache and branch-prediction behaviour for unrelated hot
loops. That is deterministic per build, which is why it reproduces across runs
without being caused — and a 0.319 ms STFT-bound workload is about the most
layout-sensitive thing in the suite. Worth noting `en_tdnn_15M` RSS and the
nemotron load rows, which I flagged the same way, have both gone this run.

@TarikAlHadethi if you have a moment: does this reproduce on your i7 outside CI?
The bench is
`tract <parakeet preprocessor.nnef.tgz> -t transformers_detect_all --nnef-tract-transformers --set B=1 --set A=16000 -O bench`
against main and this branch. If it does not reproduce on another x86 box, that
would settle it as a property of that runner rather than the change. I have only
aarch64 here, which is why I am asking.

There is also a new secondary I think genuinely is mine: `hey_snips_v31`
load+optimize +5.6% on cortex-a9. `can_direct_spatial` now asks for
`pool_spec.output_shape()` on every conv that gets that far, which is real work
added at codegen. I can make that lazier if the load time matters on the small
cores — say the word.

TarikAlHadethi:
Ran the parakeet preprocessor bench on the i7. It does not reproduce here: the two builds are indistinguishable, and the sign of the difference flips between batches.

Compared `7ef965c` (this branch's merge base) against `de46085`, rather than current `main`, so nothing from main's later commits is folded in. Both built the same way, MSVC release, then run alternating base/PR rather than all of one and then all of the other, and pinned to P cores. That last part matters on this box: it is a 12700K, so an unpinned 0.3 ms workload can land on an E core and read as a large regression that is really the scheduler.

ms/i, `-O bench --allow-random-input`, same args as `benches.toml`:

| batch | base median | PR median | delta |
|---|---|---|---|
| 1, n=8 each | 0.3825 | 0.393 | +2.7% |
| 2, n=14 each | 0.381 | 0.378 | −0.8% |

Full range across all 44 runs was 0.363 to 0.431 on both builds, so roughly 16% spread run to run, and both medians sit inside it. Combined means are 0.387 base against 0.391 PR, about +1%, with the two batches disagreeing on direction.

One caveat I would not want to overstate this past. My toolchain and CPU are both different from the runner, MSVC against your Linux build and a hybrid 12700K against the 11900KB. If the cause really is code layout, that is a property of a particular binary from a particular compiler, so a null result here does not disprove it happening on that runner. What it does say is that the +47% is not inherent to the change: on another x86 host the same two commits are the same speed.

Happy to run the other x86 rows if useful, or to try this again with the layout hypothesis in mind, for instance the same two commits built with `-C llvm-args=-align-all-functions=6` on both sides to see whether forcing alignment collapses the gap on your runner.

czoli1976:
And are you happy with the OCR Models Performance now ? Have you tried
again ST and MT ? I know this is not closing the gap entirely

TarikAlHadethi:
Happier, yes. Re-ran both models on the i7-12700K at `de46085` against `7ef965c`, the merge base, and did ST and MT this time.

Best of 3, ms/i, both builds pinned to P-cores (hybrid part, and a run landing on an E-core is not a measurement of anything):

| model | mode | base | PR | speedup |
|---|---|---|---|---|
| tiny_det 1x3x640x640 | ST | 291.286 | 177.444 | **1.64x** |
| tiny_det 1x3x640x640 | MT | 292.868 | 178.702 | **1.64x** |
| tiny_rec 1x3x48x320 | ST | 11.192 | 9.369 | 1.20x |
| tiny_rec 1x3x48x320 | MT | 11.165 | 9.437 | 1.18x |

The thing I most wanted to know is that **1.64x on tiny_det survived the whole regression pass**. It was 1.64x at `3973c01` too, and the five commits since gate a lot of the new lowering, so it could easily have been given back. It was not, and tiny_rec improved slightly, 1.14x to 1.20x.

**ST and MT are the same number.** Neither model gains anything from threads here, on either build, so nothing in this PR changes thread scaling because there is not any to change. If that is unexpected to you it might be worth a separate look, but it is equally true of the base, so it is not this PR.

On ORT, a caveat before the numbers: my ORT baseline today is faster than the one in my earlier table, 41.883 against 52.106 on tiny_det, same script and same 1.29.0. I pinned it this time, which the earlier run did not, so please do not read my old and new "x ORT" columns against each other. Both pinned, ST:

| model | base | PR | ORT p50 | base/ORT | PR/ORT |
|---|---|---|---|---|---|
| tiny_det | 291.286 | 177.444 | 41.883 | 6.95x | **4.24x** |
| tiny_rec | 11.192 | 9.369 | 3.757 | 2.98x | **2.49x** |

You asked earlier whether `OptMatMulPack` and `Erf` moved. They did not. tiny_det profile, base against PR, ms/i:

| op | base | PR | delta |
|---|---|---|---|
| OptMatMul | 128.220 | 85.442 | **−42.8** |
| DepthWiseConv | 72.711 | 17.039 | **−55.7** |
| MultiBroadcastTo | 52.658 | 1.215 | **−51.4** |
| DirectSpatialConv | 0.000 | 50.457 | +50.5 |
| OptMatMulPack | 27.610 | 28.756 | +1.1 |
| Pad | 18.607 | 0.000 | −18.6 |
| OptMaxPool | 12.072 | 3.273 | −8.8 |
| DeconvSum | 7.941 | 2.170 | −5.8 |
| Erf | 5.806 | 6.011 | +0.2 |

`MultiBroadcastTo` all but disappearing is the constant scale fold, and `Pad` is gone entirely. `DirectSpatialConv` picking up 50 ms to take 43 off `OptMatMul` plus everything im2col dragged with it is a good trade.

`OptMatMul` plus `OptMatMulPack` is 114.2 of 227.0 in the profile, so it has gone from 43% of the run to 50%. Everything else got faster and packing did not, so it is the same conclusion as last time, only more so: the remaining distance is packing and skinny-K.

Profile totals run high against the bench numbers, 227 against 177, which is just the profiler's own overhead. The deltas are the part worth reading.

czoli1976:
@TarikAlHadethi thank you, that is a much better-controlled run than I could have
done — merge base rather than current main, alternating rather than batched, and
P-core pinning, which on a 12700K is exactly the thing that would have
manufactured a fake +47% on a 0.3 ms workload.

And your caveat is the right one, not a hedge: MSVC on a hybrid 12700K is a
different binary on a different core from Linux on an 11900KB, so a null result
cannot disprove a layout effect on that particular runner. What it does settle is
that the regression is not inherent to the change, which is what I could not
establish from aarch64 alone. Two independent x86 hosts disagreeing, on a row
where the model provably runs none of this PR's code, is about as far as I can
take it without the runner itself.

Your `-align-all-functions` idea is the right experiment, and I think it only
means anything run on the runner that shows the effect — forcing alignment on
both sides there and watching the gap close would confirm layout; doing it on a
box where the gap is already absent cannot. That is kali's call on CI time, and
I would not spend yours on it.

If you do have an appetite for one more, the more valuable target is the
tiny_det/tiny_rec pair on this branch. The last numbers you posted were `3973c01`,
and there have been six commits since, including one that changes which
convolutions take DirectSpatial at all — it now needs an output row of at least
128, which keeps your three stem nodes but drops the short ones. Locally that is
171.6 → 60.0 ms on an m1 pro, but yours is the machine where the DirectSpatial
gains were real in the first place.

Separately, `12dc4b8` takes out the `hey_snips_v31` load+optimize regression on
cortex-a9: `can_direct_spatial` was computing a full output shape per convolution
at codegen, and the input's W bounds the output's, so the short rows can be
rejected without it.

TarikAlHadethi:
Ran the tiny_det/tiny_rec pair. I had numbers for `de46085` ready when you pushed `12dc4b8`, so I rebuilt and re-ran everything at the head rather than post the stale ones. Base throughout is `7ef965c`, the merge base, not current main.

Best of 3, ms/i, both builds pinned to P-cores:

| model | mode | base `7ef965c` | head `12dc4b8` | speedup |
|---|---|---|---|---|
| tiny_det 1x3x640x640 | ST | 286.888 | 175.959 | **1.63x** |
| tiny_det 1x3x640x640 | MT | 297.595 | 182.708 | **1.63x** |
| tiny_rec 1x3x48x320 | ST | 11.400 | 9.658 | 1.18x |
| tiny_rec 1x3x48x320 | MT | 11.613 | 9.646 | 1.20x |

So **1.63x on tiny_det has survived the whole regression pass**. It was 1.64x at `3973c01` and the six commits since gate a lot of the new lowering, so it could easily have been handed back. tiny_rec went the other way, 1.14x to 1.18x.

**ST and MT are the same number**, on both builds and both models. Nothing here scales with threads, so there was no thread scaling for this PR to change. Equally true of the base, so it is not something this branch introduced, but you may want to know it.

### The 128 gate keeps all five nodes here

You predicted it would keep the stem and drop the short rows. On tiny_det at 640x640 there are no short rows to drop, so all five survive:

| node | output |
|---|---|
| Conv.0 | 1,16,320,320 |
| Conv.1 | 1,8,320,320 |
| Conv.2 | 1,16,320,320 |
| Conv.3 | 1,16,160,160 |
| Conv.82 | 1,16,160,160 |

Same five as at `3973c01`. 160 and 320 both clear 128 comfortably, so the gate is inert on this model at this input size. It would start biting below about W=256.

### `12dc4b8` specifically

No measurable change against `de46085`, on either axis, which is what a cost-only change should look like:

| | `de46085` | `12dc4b8` |
|---|---|---|
| tiny_det ST | 177.444 | 175.959 |
| load+optimize | 247 ms | 245 ms |

The node count is also unchanged at 5, so the cheaper test is selecting the same set.

Load is a real win for the branch overall though, median of 5 with the first dropped, wall time of `-O dump` with stdout to a file:

| | base | head |
|---|---|---|
| load+optimize+dump | 347 ms | **245 ms** |

### Against ORT

A caveat first: my ORT baseline today is faster than the one in my earlier table, 41.883 against 52.106 on tiny_det, same script and same 1.29.0. I pinned it this time and the earlier run did not, so please do not read the old and new "x ORT" columns against each other. Both pinned, ST:

| model | base | head | ORT p50 | base/ORT | head/ORT |
|---|---|---|---|---|---|
| tiny_det | 286.888 | 175.959 | 41.883 | 6.85x | **4.20x** |
| tiny_rec | 11.400 | 9.658 | 3.757 | 3.03x | **2.57x** |

### Profile

tiny_det, base against head, ms/i. Treat these as indicative: the profiler adds enough overhead that its total (193) sits above the bench figure (176), and single-run op costs move a few ms between runs.

| op | base | head | delta |
|---|---|---|---|
| OptMatMul | 128.220 | 77.029 | **−51.2** |
| DepthWiseConv | 72.711 | 14.246 | **−58.5** |
| MultiBroadcastTo | 52.658 | 1.219 | **−51.4** |
| DirectSpatialConv | 0.000 | 33.273 | +33.3 |
| OptMatMulPack | 27.610 | 28.116 | +0.5 |
| Pad | 18.607 | 0.000 | −18.6 |
| OptMaxPool | 12.072 | 2.643 | −9.4 |
| DeconvSum | 7.941 | 2.073 | −5.9 |
| Erf | 5.806 | 4.049 | −1.8 |

You asked earlier whether `OptMatMulPack` and `Erf` had moved. Pack has not, +0.5 ms across the whole run. Erf bounces by a couple of ms between runs and is 4 to 6 either way, so I would call it unmoved as well.

`MultiBroadcastTo` all but disappearing is the constant scale fold, and `Pad` is gone outright. `DirectSpatialConv` taking on 33 ms to remove 51 from `OptMatMul` plus everything im2col dragged behind it is a good trade.

`OptMatMul` plus `OptMatMulPack` is now 54.4% of the run, up from 43.3%. Everything else got faster and packing did not, so it is the same conclusion as last time and a sharper one: what is left is packing and skinny-K.

czoli1976:
@TarikAlHadethi this is the number I most wanted and least expected to survive.
1.63x on tiny_det after six commits of gating, and tiny_rec going 1.14x → 1.18x
rather than backwards. Rebuilding at `12dc4b8` instead of posting the `de46085`
numbers you already had was the right call and I appreciate it.

Two things in your data I think matter more than the speedup.

**ST and MT being the same number.** On both builds and both models, so nothing
this branch did caused it — but it means PP-OCR gets nothing at all from
threads today, on a 12-core machine. That is a larger untapped factor than
anything in this PR, and it is invisible in the single-threaded framing we have
both been using. Worth its own issue; I will file one unless you would rather,
since it is your measurement.

**The 128 gate being inert here.** Exactly right, and it is the outcome I wanted
rather than a lucky one: at 640x640 the rows are 320 and 160 and there is
nothing short to drop, so all five nodes survive. It only engages below about
W=256, which is where DFN3's encoder lives at 32 and 96 — the model it was added
for. A gate that changes nothing on the model it is not aimed at is doing its
job.

The load figure is a win I had not measured at all: 347 → 245 ms for
load+optimize+dump. Most of that will be `MultiBroadcastTo` and `Pad` leaving
the graph rather than anything I did deliberately, but it is a third off model
preparation and it matters for anyone loading these repeatedly.

Your ORT caveat is well taken and I will not compare the old and new x-ORT
columns. Pinned against pinned, 6.85x → 4.20x on tiny_det is the honest read.

On `12dc4b8` showing no measurable change against `de46085`: that is what I
hoped for. It only removes a `pool_spec.output_shape()` call per convolution at
codegen, so identical selection with less work — your unchanged node count of 5
confirms the cheaper test picks the same set, which is the part I could not
verify from here.

czoli1976:
@TarikAlHadethi one more if you have the appetite, and this one is squarely a
question only your machine can answer.

Your tiny_det profile has `OptMatMulPack` at 28.1 ms out of 193 ms — 14.6% of
the run. On an m1 pro the same line is 3.43 ms out of 61 ms — 5.6%. The models
differ by 2.9x overall and the packing differs by 8x. Packing is ~2.6x more
expensive relative to the rest of the work on your box than on mine, and I
cannot see why from here.

So I have two independent attempts at that line, each behind a knob, default
off, on one branch: `czoli1976/tract` branch `perf/matmul-pack-experiments`.
It is PR head `12dc4b8` plus two commits, so the baseline in that build is
exactly the numbers you already posted. One build gives four runs.

```
# baseline (should reproduce your 12dc4b8 numbers)
tract ... -O dump --profile -R
TRACT_LAZY_MATMUL_PACK=1  tract ... -O dump --profile -R
TRACT_PACK_BLOCK_DEST=1   tract ... -O dump --profile -R
TRACT_LAZY_MATMUL_PACK=1 TRACT_PACK_BLOCK_DEST=1 tract ... -O dump --profile -R
```

What I actually need is the per-op summary rather than the total: the
`OptMatMul` / `OptMatMulPack` / `LazyMatMulPack` lines. The totals will move
by a few percent and be hard to read against your run-to-run spread; the pack
line moving or not moving is unambiguous.

**`TRACT_LAZY_MATMUL_PACK`** stops materialising an activation's packed buffer
and has the matmul gather each panel into its own L1 scratch as it needs it.
On aarch64 that is pack 3.432 -> 0.036 ms, `OptMatMul` 30.116 -> 32.340, total
61.333 -> 59.745 — i.e. it does not delete the gather, it relocates it into
cache, and nets 2.6%. If the 14.6% is the same gather being more expensive on
your memory system, this should be worth substantially more there than here.

**`TRACT_PACK_BLOCK_DEST`** is a different theory about the same line.
`pack_mn_major` runs `k` outer, so each pass strides the destination by
`panel_len` across the whole packed buffer and the next `k` walks it again —
for a 1x1 convolution's activation that is tens of MB re-traversed `k` times.
The commit blocks the panes so the bytes one `k` pass revisits stay resident.
Byte-for-byte identical packing, only the write order moves. On m1 pro it is in
the noise, which is consistent with either "no effect" or "this machine's
prefetcher already hides it" — those are indistinguishable from one host.

They are not alternatives; they compose, which is why the fourth run is there.

All four combinations are bit-identical on tiny_det here, and `tract-linalg`'s
pack tests pass with the blocking on and off.

If you only have time for one thing: `TRACT_PACK_BLOCK_DEST=1` on tiny_det,
pack line only. That is the one where I have no prediction at all and yours is
the machine where the answer exists.

czoli1976:
Four commits on top of `12dc4b8`, all x86-64. Three of them speed up the along-W
kernel this PR introduces; the fourth gives the lowering an off switch.

| | change | measured |
|---|---|---|
| `2dbd7380` | four accumulators in `avx2_fma` — the tap loop kept one, so every FMA waited on the previous one's latency (NEON already has two) | DirectSpatialConv −5.5% |
| `d081887e` | `avx2_oc4` runs two W blocks per tap against the same four weight broadcasts, which are what bound the loop | **−10.9%** |
| `9c38d50b` | a zmm path for unit stride, remainder handed to the ymm kernel so the accumulation order is unchanged | −4.8%, AVX-512 only |
| `69d52bdc` | `TRACT_DISABLE_DIRECT_SPATIAL` | — |

All three kernel commits are bit-identical on `tiny_det` and `tiny_rec` and are
inside `#[cfg(target_arch = "x86_64")]`, so aarch64 codegen is untouched by
construction. `cargo fmt` clean, 316 core tests, and both crates `cargo check`
clean for `x86_64-apple-darwin`. The x86 numbers are single-threaded on a Linux
x86-64 box; I cannot execute these kernels on my own machine, so the bench bot's
`i9-11900kb` runner is the real gate — it has AVX-512, so it exercises both the
ymm and the zmm path.

On the knob: `can_direct_spatial` is a purely structural predicate with no way to
turn it off, so a model where the lowering loses can only be diagnosed by
rebuilding. `try_blocked_conv` and `should_use_lazy` both carry one. Given this
PR's earlier regression list, being able to A/B the spatial path per model from a
bench run seems worth the six lines. Default is off, so nothing changes without
setting it.

Two pieces from the same work are independent of this PR and are open separately:
sonos/tract#2786 (x86 had only an AVX-512 `Erf`, so every non-AVX-512 machine ran
a scalar loop — 11.9x on that line, `tiny_rec` −27.7% on the FMA tier) and
sonos/tract#2787 (the exact-GELU detector missed PP-OCR's operand ordering, so all
13 chains stayed unfused).

---

@TarikAlHadethi — please **don't** run the pack experiment I asked you for
yesterday. Both knobs have now been measured on x86 and both are dead:
`TRACT_LAZY_MATMUL_PACK` removes the pack line (18.57 → 0.15 ms) but `OptMatMul`
absorbs +16.8 ms, so it is a net ~4% *worse* — the packing is data movement, not
waste, which is the opposite of what I argued. `TRACT_PACK_BLOCK_DEST` is neutral
(183.0 vs 183.4 ms). I have dropped both. Sorry for the detour.

The threading question I put to you is also answered, and it matters for reading
your ST == MT result: **`--threads 1` is not single-threaded.** In
`chunked_dispatch_rayon` a one-thread pool hits
`pool.is_none_or(|p| p.current_num_threads() <= 1)` and falls through to rayon's
*global* pool, i.e. every core. On an 8-core box, no flag gives `user 5.09 / real
5.30`; `--threads 1` gives `user 8.15 / real 5.22`. So a comparison built on
`--threads 1` compares MT with MT. Pass no flag at all for single-thread, and
check `user ≈ real`. I will open that as its own issue.

With that corrected, ST-vs-ST on `tiny_det` is ~2x, not 4.2x. The 4.2x is a
threading result: ORT uses 3.99 of 4 cores where tract uses 1.14, because only
the matmul parallelises — `DirectSpatialConv`, `DepthWiseConv`, `Pad` and
`MaxPool` are all exactly 1.00x on more cores.

czoli1976:
@kali I would like to pick your brain on how to proceed with this one. I am not pushing for a merge, necessarely, but wondering if and how to slice it (if you want me to do that) and what about that possible regression that may still be there (TBH not sure about that so that is probably the most important ask for you).

kali:
So my quick take: 
1. we need to cut it into pieces. Some part are completely unrelated, we can evaluate the code and merits of each of these bits on its own.
2. I'd like core to stay (mostly) platform-independent, so we should look for ways to move the new specialized stuff into linalg (and consider porting it to other platforms). Whether we can use the new routines registration is an open question.

TarikAlHadethi:
Understood on the pack experiment, I had not started it. Better to have both knobs measured and dropped than left open on my machine.

The threading one is the finding that matters to me, and it is worse than a flag choice: the doc comment directly above `chunked_dispatch_rayon` already says what you measured, that `Some(p)` with a single thread pool or `None` both fall through to `into_par_iter` on the global pool. So `--threads 1` was never single threaded, and my ST == MT result was MT against MT. It should not stand as posted.

Please open both issues, the `--threads 1` one and PP-OCR getting nothing from threads. You have the context for them and I will add my numbers to whichever you file.

If a corrected ST baseline is useful I will re-run tiny_det and tiny_rec with no flag at all and `user ≈ real` checked, on the same box. It is a 12700K, so no AVX-512, which means it exercises the ymm path in the new kernels rather than the zmm one.

czoli1976:
> So my quick take:
> 
> 1. we need to cut it into pieces. Some part are completely unrelated, we can evaluate the code and merits of each of these bits on its own.
> 2. I'd like core to stay (mostly) platform-independent, so we should look for ways to move the new specialized stuff into linalg (and consider porting it to other platforms). Whether we can use the new routines registration is an open question.

think about it for more directions on this so can proceed swiftly with them

TarikAlHadethi:
The `erf.rs` guard is already on the branch, so that one is closed.

On slicing, same shape as the wgpu series. The order I would suggest, cheapest argument first:

1. Nearest upsample in one pass. The substance is the lowering, not a kernel, and it removes the `Tile` into `MultiBroadcastTo` pair outright. It stands completely on its own and it is the easiest of these to judge.
2. MaxPool 2x2 along W. One kernel and one lowering rule. Worth taking before the conv, because it is where the core/linalg seam actually gets decided, and deciding it on a small diff is cheaper than deciding it on DirectSpatial.
3. Deconv 2x2 s=2 unpack.
4. DirectSpatialConv, by which point the seam is already agreed.
5. The fusions last, since they need the ops to exist first.

On core staying platform-independent, the wasm carve-out in this PR is the tell. `DirectSpatial off on wasm` is core making a platform judgement in its own code. It disappears if the lowering asks linalg two questions instead: is there an along-W FIR on this machine, and what does it cost at this k against the best mmm. On wasm the first answer is no, im2col stays, and there is no special case left to write.

On whether the routines registration fits, I think it splits in two.

The leaf kernels fit it. `DepthwiseWF32` is already declared in `routines.rs` on main with the signature this work wants (taps, offsets, in_stride, one run of outputs), so the along-W family has a home there already, and MaxPool-along-W and the deconv unpack are the same shape of thing.

The picker does not fit it, and the file says so itself: "These have no shape to weigh: one kernel wins outright, by what it is written against." DirectSpatial versus im2col does have a shape to weigh. k gates it at 64, it goes to 576 without AMX, and Apple keeps fat 3x3 on im2col regardless. That is the `mmm_routines` case, a pool and a tier ladder, not a single winner. So the kernels go in `routines`, and the choice needs to live wherever the cost question lives.

Happy to be wrong on the second half of that.

Last thing. You said each bit should be evaluated on its own merits, and merit here is a number. I have the 12700K, no AVX-512, so it exercises the ymm path rather than the zmm one, and a merge-base baseline from the same box. If you slice it, I will measure each slice on its own so every piece arrives with its own number instead of all of them sharing one.

kali:
On the process, yes. Codebots have made spikes cheaper than they use to be, and you find yourself with one in your hands before you realize it. We are integrating one about MOE, the one about wgpu, let's treat this as another one. The proposed slicing looks fine.

About the routines form, as you notices, target dependant implementation routines are cheap, input dependant ones are way more expensive. We have exactly one case of this at this point, the matmul family. I am not sure I want to see a second one. The rule of thumb is, I think, "if it fits in the simple routine framework, move it to linalg" (which is what I asked czoli to to about the recent depthwise optimized one). It is not inflexible. Simple strategy logic is allowed in the depthwise / convolution lowering (we already have some, including knobs, it's an option). Let's think about it and discuss it when we see the PR on its own.

About the bench, I have a i9 9990K machine that I could put in bench. Would it be representative of your use case ?

czoli1976:
Go ahead @TarikAlHadethi !

kali:
#2822

TarikAlHadethi:
Yes, the 9900K is representative. It is Coffee Lake, so AVX2 and FMA3 with no AVX-512, which is the same ymm path my 12700K takes. It is arguably a better box for this than mine: its cores are homogeneous, where Alder Lake splits P and E and puts scheduling variance into anything I measure at home. Absolute numbers will differ, but what these kernels turn on is which one wins at which shape, and that is decided by the instruction set and the cache behaviour rather than by the microarchitecture. I see it is already in the fleet, thanks for that.

On the routines question, your rule of thumb settles it. I will keep the leaf kernels in routines, where DepthwiseMF32 already sits, and put the choice in the convolution lowering as strategy logic rather than standing a second input-dependent picker next to mmm_routines. Same behaviour, no second family.

Starting on the nearest upsample. Each slice will arrive with its own numbers.

czoli1976:
@TarikAlHadethi do you have plans to try to shave the rest of the gap and/or look at the MT Scaling ?

TarikAlHadethi:
Both, and the threading half is the one I owe you.

**MT scaling.** The corrected baseline from 7 Sep is still outstanding and it is mine to post. My ST/MT pair used `--threads 1` on the ST side, so it was MT against MT and nothing in it should be read as a scaling result. What I will run on the 12700K is no flag at all for single thread with `user ~= real` checked, `--threads 0` for the pool, and a per-op profile of both, so the table says which ops move with cores and which sit at 1.00x rather than giving only the whole-model number. No AVX-512 on that box, so it exercises the ymm path.

I do not see an issue open for either half, the `--threads 1` one or PP-OCR getting nothing from threads. If you would rather not write them, I will open the first with the repro. It is unchanged on main today: the default executor is `SingleThread`, and `chunked_dispatch_rayon` still sends `Some(p)` with a one-thread pool down `into_par_iter` on the global pool, so no flag is genuinely serial while `--threads 1` is every core. `bench_suite` filters `1` out of its sweep and calls it a global-pool trap, so the harness already knows. Someone timing by hand does not.

**The gap.** Two separate things, and I would rather be straight about which I can take.

The first is the rest of this PR. Slices 1 to 3 are in (#2829, #2830, #2831) and they are the smaller half of the tiny_det win. DirectSpatialConv and the fusions are the rest, and neither is in main. Slice 4 was waiting on the seam question, which kali answered on #2830: platform-dependent kernels may stay in tract-core for now provided they are recorded, which is what #2836 is for. So it is unblocked, and I am happy to take it next on the same terms as the others, its own numbers against a merge-base built the same day and its kernels on the record issue. It is your code, so say if you would rather carry that one yourself.

The second is what is left after that, and my own profile says where it is: `OptMatMul` plus `OptMatMulPack` was 54.4% of tiny_det at `12dc4b8`, and everything else got faster while packing did not. That is packing and skinny-K in linalg rather than a lowering, and it is a bigger piece of work than a slice. I am not going to start rewriting the matmul on my own initiative, but I will measure it if that is useful: pack cost against K, where the skinny-K cliff sits on this machine, and what the tier ladder picks at those shapes. If the numbers point somewhere specific, that is a better argument for doing the work than my opinion is.

czoli1976:
@TarikAlHadethi thanks — that split is the right one.
DirectSpatial + fusions: please take those next if you are happy to. Same
terms as the other slices: own PR, own numbers vs a same-day merge-base on
the 12700K, kernels recorded on #2836.

Suggested shape so we do not re-open the seam debate:
1. DirectSpatialConv first, fusions second (or two commits in one PR if you
prefer, but keep the review surface separable).

2. Kernels in routines, choice in conv lowering as strategy logic — no
second input-dependent picker next to mmm_routines. Gate on “is there an
along-W FIR on this machine, and is the row long enough to beat im2col at
this k”. Wasm / no-SIMD stays on im2col with no special case in core beyond
asking linalg.

3. Caps as they are now: k ≤ 64 when AMX is in play, k ≤ 576 without AMX;
Apple fat 3×3 stays im2col. Require a long enough output row so short
stages (DFN3 W=32/96, MobileNet 14×14 / 7×7) do not pay the tap-repack.

4. Do not hand off short rows or non-SIMD targets to a scalar FIR — that
was the MobileNet / Inception / rv64 / armv7 regression. HAS_SIMD is a
runtime check on x86_64 (AVX2+FMA is not baseline).

5. Fusions after the op exists: bake a const per-channel [1,C,1,1] scale
into weights/bias (relu(conv(x, w) + b) * s → relu(conv(x, w*s) + b*s) when
every s > 0). Do not leave a scalar Mul epilogue in the inner loop; hoist
relu/scale off the per-element path so NHWC still hits the const-N body.
Fall back when the scale is not const-per-channel or not strictly positive
under fused relu.

6. Keep NNEF round-trip intact: anything without a serializer stays a
codegen lowering, not declutter (same lesson as nearest upsample).

If you want a name pass while you are in there, DirectSpatialConv / along-W
fusion flags are fine; just do not introduce a second public name for the
same op. Happy to review the slice as soon as it has a dump
(DirectSpatialConv on the stem + fat 3×3 where expected) and the tiny_det /
tiny_rec table.
--threads 1: please open that issue.

Repro from your last comment is enough: default executor is SingleThread,
chunked_dispatch_rayon still sends Some(p) with a one-thread pool down
into_par_iter on the global pool, so --threads 1 is every core and no flag
is the actually serial path. bench_suite already drops 1 as a global-pool
trap; hand timing does not.

A second issue (or a section in the same one) for “PP-OCR gets nothing from
threads” is useful once you have the corrected ST / --threads 0 / per-op
table.
If you are willing to fix the flag while you are there, the intended
behaviour is: --threads 1 (and Some(1)) must be genuinely serial and must
not touch the global pool. --threads 0 stays “use the pool”. I will not
block the spatial slices on that fix.


I will leave DirectSpatial to you unless you bounce it back :-) The packing
/ skinny-K measurements you offered would help after that; no need to start
a matmul rewrite until those numbers say where the cliff is.

Thanks !

PR #2777, opened 2026-09-04
pulse-opl: cast the pad constant to the datum type it fills

cverrier:
PulsePad, PulseMask and GpuPulsePad handed the tensor PadMode carries straight to a fill that takes its value in the datum type it fills, so an f16 pulse padded by the f32 constant the NNEF deserializer builds could not run. Cast at use, as Pad and GpuPad already do, and drop the Pad case from the float precision translator: a constant keeps the datum type its builder gave it, and the tflite writer, which hands the value to PADV2 in the type of the padded tensor, casts on the way out.

### DIFF
--- a/core/src/floats.rs
+++ b/core/src/floats.rs
@@ -1,6 +1,5 @@
 use crate::internal::translator::Translate;
 use crate::internal::*;
-use crate::ops::array::{Pad, PadMode};
 use crate::ops::binary::TypedBinOp;
 use crate::ops::cast::{Cast, cast};
 use crate::ops::einsum::EinSum;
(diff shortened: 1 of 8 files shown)


kali (dismissed):

cverrier:
Nothing in CI could reproduce this: every pulsified test and harness model is f32, and the f16 runtimes drive suites with no streaming dimension. `harness/nnef-test-cases/conv-then-shape-of-mask` already pads on the streaming axis and runs at `--pulse 4`, and the `pulse` stage runs before `-t`, so an `f32_to_f16` run of that same pulsified graph reaches a PulsePad still carrying the f32 constant. That line fails on the parent commit and passes here. It covers the CPU PulsePad only — the GpuPulsePad change has no equivalent assertion.

kali (approved):

PR #2779, opened 2026-09-04
Element-wise f32 ln and exp went through libm one element at a time, …

kali:
…which is most of what a log-mel featurizer costs. Register them as linalg routines, with the scalar fits and their shared coefficients in generic/{ln,exp}.rs, FMA and AVX-512 kernels on x86_64, NEON on aarch64, and the f32 arms of the core ops dispatching to them.


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

czoli1976:
sweet ..... but not for  WASM, right?

kali:
i can give it a shot, but i'm out of tokens today :)

PR #2786, opened 2026-09-07
linalg/x86_64: add an AVX2/FMA erf kernel

czoli1976:
x86-64 had only an AVX-512 `Erf` kernel, so every machine without AVX-512 — including Alder Lake and the Coffee Lake part the `intel_fma_linear` cost model is fitted on — fell through to `generic_erf_f32_4n`, a scalar loop. This adds the same Abramowitz & Stegun polynomial over four ymm registers gated on FMA, leaving AVX-512 machines on the wider kernel.

Measured on PP-OCRv6 `tiny_rec`, single-threaded, with AVX-512 masked off (`TRACT_CPU_ISA=-avx512f,-avx512vnni,-avx512fp16`): the `Erf` line goes 40.72 → ~2.8 ms (11.9x) and the model is 27.7% faster. No effect on an AVX-512 box, which keeps the existing kernel.

Covered by the auto-generated `erf_frame_tests!` suite, including the property test against the scalar reference.

🍍

### DIFF
--- a/linalg/src/x86_64/erf.rs
+++ b/linalg/src/x86_64/erf.rs
@@ -198,3 +198,81 @@ unsafe fn x86_64_avx512_erf_f32_64n_run(buf: &mut [f32]) {
         );
     }
 }
+
+// AVX2/FMA (ymm, 8-wide) error function kernel. Same polynomial and shape as
+// the AVX-512 kernel above, over 4 ymm registers per iteration (32 lanes per
+// loop step), for the x86_64 tier without AVX-512.
+routine_ew_rust!(x86_64;
+    f32,
+    x86_64_fma_erf_f32_32n,
+    32,
+    8,
+    #[inline(never)]
+    fn run(buf: &mut [f32], _: ()) {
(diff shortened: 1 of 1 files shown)


czoli1976:
@kali can you give a try to this to see if it has any adverse effect on your x86 machine in the bench ?

czoli1976:
@TarikAlHadethi this is Vitamins for Clarus

TarikAlHadethi:
Just checked and yeah, 54 Erf nodes in rec and 52 in det. My machine's a 12700K so no AVX-512, meaning all of those were going through the scalar loop. That's basically every consumer Intel since Alder Lake, so not exactly niche.

OCR's the one slow thing in Clarus I'd kinda given up on, everything I tried just traded accuracy for speed. So getting some for free is great. Still on 0.20 so I don't have it yet, but that's a good reason to finally move.

Anyway, thanks for chasing it, appreciate it a lot.

czoli1976:
You are very welcome

PR #2787, opened 2026-09-07
core/nn: detect the erf GELU with the half factored out

czoli1976:
`detect_gelu_exact` only matched `(0.5 * x) * (1 + erf(x / sqrt(2)))`. A graph that multiplies by `x` first and scales at the end computes the same value but kept its five-node chain — that is the ordering PP-OCRv6 exports, so all 13 of its GELUs went unfused. Accept both orderings, anchored on the same `Erf`.

The in-file test now builds the chain both ways, over F32 and F16, and asserts the fused output is bit-exact against the unfused reference.

Scope, since this changes which graphs fuse: the rule is anchored on an `Erf` node, so a model without one cannot reach it. Across the 3011 ONNX models available here — the noise-suppression set (FastEnhancer, DFN3, DTLN, GTCRN, DPDFNet) and the test-asset cache including mobilenet v1/v2, inceptionv1q and en_tdnn_15M — only five contain an `Erf` at all, and four of those are single-op conformance cases.

Measured on PP-OCRv6 `tiny_det`, single-threaded x86-64: 13 `GeluExact` nodes, zero `Erf` left, -5.5% by profile.

🍍

### DIFF
--- a/core/src/ops/nn/gelu_exact.rs
+++ b/core/src/ops/nn/gelu_exact.rs
@@ -73,24 +73,31 @@ pub fn detect_gelu_exact(
     rule_if_some!(out = model.single_succ(one_plus_erf.id)?);
     rule_if_some!(out_op = out.op_as::<TypedBinOp>());
     rule_if!(out_op.0.is::<Mul>());
-    rule_if_some!(
-        half_x = out
-            .inputs
-            .iter()
-            .filter_map(|i| {
-                let n = &model.nodes()[i.node];
-                n.op_as::<TypedBinOp>()?.0.is::<Mul>().then_some(n)
-            })
-            .next()
-    );
-    rule_if!(model.matches_single_input_const(half_x, 0.5));
-    rule_if!(half_x.inputs.contains(&x));
+    let half_x = out
+        .inputs
+        .iter()
+        .filter_map(|i| {
+            let n = &model.nodes()[i.node];
+            n.op_as::<TypedBinOp>()?.0.is::<Mul>().then_some(n)
+        })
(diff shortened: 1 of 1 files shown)


czoli1976:
@kali can you give a try to this to see if any regression

kali:
i'm really fed up with these bimodal benches. non of these model use erf, right ?

czoli1976:
> i'm really fed up with these bimodal benches. non of these model use erf, right ?

right

czoli1976:
I don't think this PR can be responsible for those four rows.

`detect_gelu_exact` is the `declutter` hook on the `Erf` op itself
(`core/src/ops/math/mod.rs:761`), so it is only ever called with an `Erf` node —
a model without one never reaches it, for fusion or for decluttering cost.

I pulled all four flagged models from `tract-test-assets.tract.rs`. None contains
an `Erf`:

- `dummy-conmer-12M` — no `erf`/`gelu` anywhere in the archive, weights included
- `en_tdnn_8M_nnef` — `graph.nnef` is add/mul/matmul/sigmoid/tanh/conv/scan
  (a raw grep over the tar does hit "erf" twice, but both are inside `.dat`
  weight blobs, not the graph)
- the nemotron preprocessor, for both load rows — none

Each archive has a single `graph.nnef`, so there is no fragment file holding one
out of sight.

For what it is worth, #2771 in the same nightly cycle reports no speed
regressions and a different set of secondary rows, which suggests per-run
variance on those boards rather than a shifted baseline.

Happy to re-run if you would rather see it on a fresh nightly.

PR #2789, opened 2026-09-07
cli: glibc's allocator carries a lot of overhead on the graphs the be…

kali:
…nch suite runs, and a process can settle into a regime where a per-eval intermediate is mapped and unmapped on every pass, costing millions of minor faults. Use jemalloc as the CLI global allocator, wrapped in the same readings instrumentation, behind a jemalloc feature that bench-suite turns on.


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2796, opened 2026-09-08
fix(tensorflow): return errors instead of panicking on malformed TensorProto (CWE-248)

li-jin-quan:
I found a crash in the TensorFlow model loader.

A malformed `.pb` kills the process instead of returning an error. Three spots:

```rust
let dims: TVec<usize> =
    t.tensor_shape.as_ref().unwrap().dim.iter().map(|x| x.size as _).collect();
let dtype = DataType::try_from(t.dtype).unwrap();
_ => unimplemented!("missing type (for get_tensor_content) {:?}", dtype),
```

- `tensor_shape` missing -> `unwrap()` panics
- `dtype` set to a value prost does not know -> `unwrap()` panics
- a dtype tract does not support -> `unimplemented!()` aborts

The dim cast is a fourth one: `x.size as _` turns a negative dim into a huge
`usize`. A dim of `-1` becomes `usize::MAX`.

This matters for anything that loads a model it did not produce: a converter, a
model zoo, CI running on user uploads.

## You already fixed this on the ONNX side

`onnx/src/tensor.rs` had the exact same `unwrap()`. It was fixed in 1426b091 and
bd853a31, and the regression test there says:

> an ONNX TensorProto with a DataType value that prost doesn't recognise used to
> abort the process via `unwrap()`

The TensorFlow loader did not get the same treatment.

## Fix

- `tensor_shape` -> `ok_or_else`
- `dims` -> reuse `TryFrom<&TensorShapeProto> for TVec<usize>`, which already
  exists in this file and already rejects negative dims
- `unwrap()` -> `?`
- `unimplemented!()` -> `bail!()`
- added a regression test, same shape as the ONNX one

One detail: `-1` is a legal TF dim, it means "unknown". My test uses `-2` so it
does not depend on that. But tract cannot represent an unknown dim today, so
`-1` now gives a clean error instead of reaching the allocator with
`usize::MAX`.

Valid models load exactly as before.

Does this look right to you?

### DIFF
--- a/tensorflow/src/tensor.rs
+++ b/tensorflow/src/tensor.rs
@@ -83,11 +83,16 @@ fn tensor_from_repeated_field<T: Datum>(shape: &[usize], data: Vec<T>) -> TractR
 impl TryFrom<&TensorProto> for Tensor {
     type Error = TractError;
     fn try_from(t: &TensorProto) -> TractResult<Tensor> {
-        let dims: TVec<usize> =
-            t.tensor_shape.as_ref().unwrap().dim.iter().map(|x| x.size as _).collect();
+        let shape = t
+            .tensor_shape
+            .as_ref()
+            .ok_or_else(|| format_err!("TensorProto is missing its tensor_shape"))?;
+        // Go through the checked conversion instead of `as _`: it rejects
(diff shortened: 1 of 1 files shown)


kali:
Thanj you !

PR #2801, opened 2026-09-08
RFC: a WebGPU backend (tract-wgpu)

czoli1976:
tract reaches a GPU through Metal and CUDA, neither of which a browser can. This adds `tract-wgpu`, a third `tract-gpu` backend built on the `wgpu` crate: one set of WGSL kernels that run on WebGPU in a browser and on Vulkan/Metal/DX12 natively.

Opening as a draft because the code is the easy part of this question — the real one is whether you want a third GPU backend in-tree at all, and on what maintenance terms. I'd rather hear that before polishing anything.

## What it is

A new crate and nothing else. **It changes no existing file**: the diff is `wgpu/`, plus two lines of `Cargo.toml` and the lockfile. No edit to tract-core, tract-gpu, metal or cuda, so nothing that works today can regress.

It covers a full selfie-segmentation model on the GPU with no CPU fallback — convolution (dense, depthwise, transposed), pooling, matmul via `PrefixMatMul`, reduce, softmax, resize, cast, and the element-wise and binary sets, in f32 and f16. There is a `WgpuTransform` like `MetalTransform`, a coverage check that refuses to load a model it would silently split, and an optional hybrid mode that falls back to CPU per-op behind a feature flag.

## Three things WebGPU does not let a `tract-gpu` backend do

These are the parts worth your opinion, because they are contract questions rather than kernel ones.

**The memory pool's single-buffer arena is illegal.** WebGPU refuses a dispatch that binds one buffer both read-only and read-write, which is what an arena does whenever a kernel reads one tensor and writes another in the same allocation. The validation error is `Buffer with 'tract-wgpu-arena' usage ... is used as both a read-only storage and a writable storage`. The backend therefore gives each tensor its own buffer and addresses it as `(buffer, byte offset in a uniform)`.

**`DeviceBuffer::ptr()` has no meaning.** There is no GPU address to hand out. It returns the `wgpu::Buffer`'s handle identity, which is enough for cache keying and nothing else; kernels downcast to the concrete buffer type instead.

**`get_bytes_slice` cannot report a failure.** On a unified-memory backend it is a slice; here it is a device readback that can fail, and the trait returns `Vec<u8>`. Today the path is unreachable — it is only entered through the arena view, which this backend never installs — but it is a trap if the arena ever becomes reachable. On wasm it cannot succeed at all without JSPI.

None of these need a change from you for this PR to work; they are all handled inside the crate. I raise them because if `tract-gpu`'s contract were ever revisited, these are the three places a non-unified-memory backend pushes back.

## Numbers

MobileNet-style selfie segmentation, 144×256, on an M-series Mac.

| | tract-wgpu | reference |
|---|---|---|
| Chrome, GPU-resident | 2.70 ms/frame | onnxruntime-web WebGPU EP, 2.09 ms |
| Chrome, full frame incl. readback | 6.63 ms | onnxruntime-web, 5.40 ms |
| Chrome, background-replacement demo, model only | 1.5–1.7 ms | tract's own 1-thread wasm tier, ~9.5 ms |
| Native | ~3.1–3.7 ms/frame | tract's CPU path on the same box, ~6.1 ms |

**These replace an earlier table that had us ahead of onnxruntime-web on a full frame. That was wrong** — measured against a slower run of theirs, and on our side with a performance bug in the uniform path that I have since fixed. Measured properly in one sitting we are about 1.3× behind them in a browser, and the row that matters for the use case is still the third: inference on a video call must not eat the cores the encoder needs.

Where the gap is, in case it is useful to you: their kernels do **more** GPU work than ours (3.08 ms summed against our ~2.4 ms) and they still land lower, because they issue 122 dispatches to our 140 and recording is about **half** of a WebGPU frame — `record 1.28 ms + submit 0.32 + readback 3.12` on our side. So this backend's remaining problem in a browser is dispatch count and host-side recording, not arithmetic. Native does not show this at all: there the driver wait hides recording, and the same changes that help one target do nothing for the other.

## Testing

`cargo test -p tract-wgpu` is 26 tests against CPU reference, per kernel family. A model-level test compares the whole graph to the CPU plan at `Approximation::Approximate` and asserts full GPU coverage — that one lives in an example crate that is not part of this PR, because it embeds an ONNX file the repo does not carry. If you want the model test in-tree, tell me how you would rather a test model be provisioned and I will wire it that way.

Not covered: no CI runs this. It needs a GPU, and the browser numbers need a browser. I have run it on Apple hardware only — Vulkan and DX12 are what the `wgpu` crate gives us, not something I have verified.

## Not included

The browser demo and its benchmark page are a separate slice, blocked on the same test-model question. A background-replacement demo built on this exists but uses compositing shaders whose provenance I need to settle before offering them.

🍍

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -87,6 +87,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "android_system_properties"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ae221649c9976a6f6c56ae1facf410f3ddb33cc661c4b7b61020a912d4237fbc"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "anes"
 version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -196,6 +205,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "516b6b4f0e40d50dcda9365d53964ec74560ad4284da2e7fc97122cd83174516"
 dependencies = [
  "stable_deref_trait",
+]
+
+[[package]]
+name = "ash"
+version = "0.38.0+1.3.281"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0bb44936d800fea8f016d7f2311c6a4f97aebd5dc86f09906139ec848cf3a46f"
+dependencies = [
+ "libloading 0.8.9",
 ]
 
 [[package]]
@@ -382,6 +400,15 @@ dependencies = [
 
 [[package]]
 name = "bit-set"
+version = "0.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "09ec2f926cc3060f09db9ebc5b52823d85268d24bb917e472c0c4bea35780a7d"
+dependencies = [
+ "bit-vec 0.9.1",
+]
+
+[[package]]
+name = "bit-set"
 version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "56d87354e4229f54a44f7bf2435906a4656dba36026ab6eaca629a2c436a691c"
@@ -394,6 +421,12 @@ name = "bit-vec"
 version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5e764a1d40d510daf35e07be9eb06e75770908c27d411ee6c92109c9840eaaf7"
+
+[[package]]
+name = "bit-vec"
+version = "0.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b71798fca2c1fe1086445a7258a4bc81e6e49dcd24c8d0dd9a1e57395b603f51"
 
 [[package]]
 name = "bit-vec"
@@ -515,6 +548,20 @@ name = "bytemuck"
 version = "1.25.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c8efb64bd706a16a1bdde310ae86b351e4d21550d98d056f22f8a7f7a2183fec"
+dependencies = [
+ "bytemuck_derive",
+]
+
+[[package]]
+name = "bytemuck_derive"
+version = "1.12.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fc0e56a716f1e132ff6bf4bdac1c944a3fcdc1cae65f70a4a2a1ac3b401d2d1f"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 3.0.3",
+]
 
 [[package]]
 name = "byteorder"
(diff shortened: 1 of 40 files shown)


kali:
This is a great development, happy that the tract-gpu framework can handle a third platform with little issues.

I am tempted to say that the maintainance question is not as acute as it was in the pre-codebot era. They are pretty good at producing "good enough" implementations, so maintaining expectations across a third platform does not sound as crazy as it did just one year ago.

The three issues do not sound too bad either. The arena was an add-on, so making it noop on wgpu is ok. If the "read or write" contract on buffer is imperative, I am not sure it makes any sense to try and pursue an alternative implementation...

The question is more... do we try to get on par with cuda and metal, or keep it focused on audio/video signal processing ? Does it make sense to go for transformers ? Would nemotron make sense ?

czoli1976:
My focus so far has been DSP (Noise Suppression) but in my domain BG Blur and Replacement are important too. The latter need at least WebGPU to be performant , the former not good for GPU Acceleration due to latency.

So I am trying to complete the picture of what I am used to with Tract.

Took a lot of effort to take it close to ORT Web Runtime, still a bit slower on WebGPU (On CPU beat ORT in most of the cases)

So I would like to:

- Complete with Real Value Demo of BG Blur and Replacement with the Selfie Segmentation WebGPU Accelerated (almost done)
- Close the Gap with ORT, possibl beat it (my obsession)

For other types of models I am not really an expert but an ASR could be fine.

I have done some WebNN Experiments before which is the way to tap into HW Accelerated Inference via the Browser but that was before discovering TRACT, it could be another step (WebNN is pretty cutting edge and very new)

kali:
how do we slice this one ? it's "only" 9000 lines, that's less than the attempt at coremetal last spring :)

czoli1976:
1	the device — context, tensor, buffer pool, uniform ring, copy + cast	2,400
2	element-wise + binary, transform, coverage check, Runtime	950
3	matmul + einsum lowering	540
4	convolution family — dense, depthwise, transposed, pooling	1,250
5	reduce, softmax, resize	465
6	fusion — chains, epilogues, axis ops	1,020
7	browser seam — ingest/export	300

kali:
bring it on

czoli1976:
I have corrected the numbers table in the description, because the old one flattered this backend and you should not review it on those figures.

It claimed we beat onnxruntime-web on a full frame (5.04 ms against 6.41). That was measured against a slow run of theirs, and on our side with a bug in the uniform path that I have since fixed. Measured properly in one sitting, in Chrome on an M-series Mac: **we are about 1.3× behind them** — 2.70 ms against 2.09 GPU-resident, 6.63 against 5.40 with readback.

The interesting part is where the gap is. Their kernels do *more* GPU work than ours (3.08 ms summed against our ~2.4) and they still land lower, because they issue 122 dispatches to our 140 and recording is about half a WebGPU frame — ours splits `record 1.28ms + submit 0.32 + readback 3.12`. So what this backend needs in a browser is fewer dispatches and cheaper recording, not faster arithmetic. Native shows none of this: there the driver wait hides recording, and changes that help one target measure as nothing on the other.

None of that changes what the slices contain; it changes what I would work on next, and I would rather you had the real numbers in front of you.

kali:
understood. no change of plans as far as merging the sliced version.

kali:
@czoli1976 any thought about how we could manifest this in ci/bench ? daytime bandwidth is often tense, but i could still run a nightly bench on a rtx4060. not sure if we want a native or wasm run, vulkan or gl...

czoli1976:
The sliced PRs (#2805–#2811) and the follow-ups are in main, so this RFC is leftover. Closing.

On the nightly: a native Vulkan run of the selfie segmenter on the 4060 is the useful one — that measures GPU work. wasm/GL would mostly measure JS recording, which #2867 already went after. Happy to sketch a harness if you want it on the nightly.

kali:
sounds good, let's do this then.

PR #2805, opened 2026-09-08
wgpu 2/7: element-wise and binary ops, and the transform

czoli1976:
2/7 of #2801. Adds the element-wise and binary kernels, the `WgpuTransform` that translates a model onto them, and the coverage check. **Stacked on #2811 — its commit is included here, so read this one's top commit only.**

With this a model of element-wise ops runs end to end on the GPU.

The policy question worth your opinion is what to do with an op the backend has no kernel for. On Metal a fallback is a cheap trip through unified memory; on WebGPU it is a readback and an upload, which is far more expensive than the op it rescues. So the default refuses to load a model it would silently split, and names the ops that made it refuse; a `jspi` Cargo feature turns that into a per-op CPU fallback for the browsers that can suspend across a buffer mapping. Happy to change the default if you would rather it always limp than refuse.

Where the operands run with the output a thread handles four values at once — the kernel is bandwidth-bound, so a wider load is the whole win.

Ten tests, including a two-op model against the CPU plan and both coverage outcomes.

🍍

### DIFF
--- a/wgpu/src/tests.rs
+++ b/wgpu/src/tests.rs
@@ -1,11 +1,13 @@
 #![cfg(test)]
 
 use tract_core::internal::*;
+use tract_core::ops::math::mul;
+use tract_core::transform::ModelTransform;
 use tract_gpu::tensor::{DeviceTensor, IntoDevice};
 
 use crate::kernels::cast::wgpu_cast_dispatch;
 use crate::kernels::copy::wgpu_copy_nd_dispatch;
-use crate::with_wgpu_queue;
+use crate::{WgpuTransform, with_wgpu_queue};
 
 fn close(a: &Tensor, b: &Tensor) -> TractResult<()> {
     a.close_enough(b, Approximation::Close)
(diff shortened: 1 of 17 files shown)


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

wgpu/src/tests.rs
```diff
@@ -1,16 +1,37 @@
...
+        model,
+        &RunOptions { skip_order_opt_ram: true, ..RunOptions::default() },
+    )?;
+    let cpu_plan = SimplePlan::new(cpu_model)?;
+    let gpu_out = Arc::new(gpu_plan).run(tvec!(input.clone().into()))?.remove(0).into_tensor();
+    let cpu_out = Arc::new(cpu_plan).run(tvec!(input.into()))?.remove(0).into_tensor();
+    close(&gpu_out, &cpu_out)
+}
+
 #[test]
```
kali, line 16: is this the way to go ? VS test-rt -like ?

czoli1976: Well, cannot argue against raising the bar and going test-rt ...

czoli1976: doing it so will see another commit for it

czoli1976:
The gate allowed all sorts of dtypes there are no kernels able to handle (only got F32/F16) ....

kali (dismissed):

wgpu/src/tests.rs
kali: great :)

kali:
So... giving up on test-rt for now ? I should merge ?

czoli1976:
look again should be TTT https://www.youtube.com/watch?v=qLHLsbD9emc

kali (approved):

PR #2806, opened 2026-09-08
wgpu 3/7: matrix multiplication

czoli1976:
3/7 of #2801. Adds the matmul kernel, the `PrefixMatMul` op over it, and the EinSum lowering that feeds it. **Stacked on #2805 — read this one's top commit only.**

The lowering is not optional for vision models: a pointwise convolution reaches the backend as an EinSum, so without it a segmenter arrives as a graph of ops the backend has no kernel for and the coverage check from 2/7 refuses the model.

The kernel here is naive — one thread per output element, no blocking, no staging — and it carries an epilogue that 6/7 fills in, empty for now.

**Correction to what I first wrote here.** I said blocking had been measured and always lost by dividing the thread count. That was true of every variant I had tried at the time, and it is not true in general: a thread computing a 4×4 tile from `vec4` loads, with the k loop unrolled four deep, is **2.2× faster on this model's products** and correct against a CPU reference. The occupancy rule I had drawn was too broad — it should read that blocking loses when it cuts thread count *without* raising the work each thread does. Here the thread count drops 16× and the work per thread rises 16×, and the device still saturates.

It needs the weights transposed to `(k, n)` so four columns are one `vec4`, which is free because they are constant — the activations are not moved and there is no layout pass. I have it working (native: the products 1.95ms → 0.874ms, whole frame −12.4%) but it is deliberately **not in this PR**: it belongs on top of a reviewed base, and its value in a browser is still unresolved — WebGPU frames are dominated by dispatch recording, where it helps much less than it does natively. It will come as a follow-up with its own numbers.

Nine tests over transposes, batched prefixes and broadcast prefixes.

🍍

### DIFF
--- a/wgpu/src/kernels/matmul.rs
+++ b/wgpu/src/kernels/matmul.rs
@@ -1,0 +1,167 @@
+//! Naive prefix-broadcasting GEMM: one thread per output element.
+//!
+//! `PrefixMatMul` is what tract lowers `EinSum` to, and a MobileNet-style
+//! segmenter reaches it through every pointwise convolution, so the backend
+//! needs it even though the tiled kernels (tfjs `matmul_packed`, ORT
+//! `subgroup_matrix_gemm`) are the ones worth having later.
+
+use tract_core::internal::*;
+use tract_gpu::tensor::DeviceTensor;
+
+use crate::kernels::shaders::{
+    ChainStep, EntryPoint, LayoutKind, ModuleKey, ModuleKind, PipelineKey, ShaderDtype, keys_for,
+    matmul_module, pack_u32s, program_key, rpad8_dims, rpad8_strides,
+};
+use crate::utils::{element_offset, get_wgpu_buffer};
+use crate::with_wgpu_queue;
+
+/// Rank the 8 stride slots can address, two of them being the matmul axes.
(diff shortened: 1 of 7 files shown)


czoli1976:
Correcting something I wrote in the description, since it was wrong and you may have read it.

I said the naive kernel was measured rather than lazy, and that every blocked variant lost by dividing the thread count. That held for everything I had tried then, but not in general: a thread computing a 4×4 tile from `vec4` loads, with the k loop unrolled four deep, is **2.2× faster on this model's products** (1.95ms → 0.874ms) and matches a CPU reference. My occupancy rule was too broad — blocking loses when it cuts thread count *without* raising the work per thread; here both move by 16× and the device still saturates.

It wants the weights transposed to `(k, n)` so four columns are one `vec4`, which costs nothing because they are constant, and it leaves the activations alone — no layout pass. Working, but deliberately not in this PR: it belongs on a reviewed base, and its value in a browser is still unresolved. Follow-up with its own numbers.

czoli1976:
@kali ready

kali (approved):

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2807, opened 2026-09-08
wgpu 4/7: convolution, pooling and transposed convolution

czoli1976:
4/7 of #2801. Adds convolution, pooling and transposed convolution. **Stacked on #2806 — read this one's top commit only.**

A depthwise convolution gets a kernel of its own that stages its input tile in workgroup memory: with one channel per group the dense kernel reads every input value once per output and has nothing to reuse.

Both dense kernels run the channels innermost. Which input pixel a tap reads does not depend on the channel, and the transposed kernel needs two integer divisions to decide whether a tap lands on an input pixel at all — paying that per channel rather than per tap made it three times slower than it needed to be.

Conv, pool and transposed conv are covered by the shared conformance suite added in #2805 (`test-rt/test-wgpu`): the `suite-unit` conv/deconv proptests and the `suite-onnx` / `suite-pulse` cases, run across the three runtimes (`wgpu_translate`, `optimized_wgpu`, `optimized_wgpu_transpose`) — 4071 pass, 0 fail, nothing skipped. The transform declines what it has no kernel for, so anything uncovered runs on the CPU rather than failing the suite.

🍍

### DIFF
--- a/wgpu/src/kernels/conv.rs
+++ b/wgpu/src/kernels/conv.rs
@@ -1,0 +1,242 @@
+use tract_core::internal::*;
+use tract_core::ops::cnn::Conv;
+use tract_gpu::tensor::DeviceTensor;
+
+use crate::kernels::shaders::{
+    ChainStep, DW_WG, DepthwiseShape, EntryPoint, LayoutKind, ModuleKey, ModuleKind, PipelineKey,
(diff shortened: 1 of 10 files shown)


czoli1976:
@kali  Park this waiting for Nr 3 to get merged

czoli1976:
@kali  ready

kali (approved):

PR #2808, opened 2026-09-08
wgpu 5/7: reduction, softmax and resize

czoli1976:
5/7 of #2801. Adds reduction, softmax and resize — the rest of the op set a segmentation model needs. **Stacked on #2807 — read this one's top commit only.**

One thing to flag: `GpuReduce` carries `axes: TVec<usize>` but its `eval` only honours `axes[0]`, so a multi-axis reduce would quietly reduce one axis and return. This slice works around it in the transform by splitting a multi-axis reduce into a chain of single-axis ones, the way metal does.

**I originally offered to fix the shared op instead. Please ignore that offer — I tried it and it is a bad idea.** Collapsing a contiguous axis run into one dispatch is trivial (the kernel's `[outer, k, inner]` view merges them) and it is numerically correct, but it was **71% slower**, because the split is a parallel reduction in disguise: reducing W first gives `outer = C×H` threads, and merging both axes leaves `outer = C` — sixteen threads each summing 2304 elements. The one-dispatch version would need a workgroup-cooperative kernel with a tree reduction, which is a separate piece of work. The extra dispatch is the cheaper of the two, so the split earns its keep and the shared op's behaviour is fine as it is for this backend.

Two tests; the rest of this op set is covered by the model-level test that is still waiting on the test-model question from #2801.

🍍

### DIFF
--- a/wgpu/src/kernels/mod.rs
+++ b/wgpu/src/kernels/mod.rs
@@ -6,7 +6,10 @@ pub mod deconv;
 pub mod element_wise;
 pub mod matmul;
 pub mod pool;
+pub mod reduce;
+pub mod resize;
 pub mod shaders;
+pub mod softmax;
 
 /// Every pipeline this crate's kernels can need. They are built when the
 /// context is created so that no frame ever waits on a shader compile.
@@ -20,5 +23,8 @@ pub fn all_pipeline_keys(shader_f16: bool) -> Vec<shaders::PipelineKey> {
     keys.extend(element_wise::all_pipeline_keys(shader_f16));
     keys.extend(matmul::all_pipeline_keys(shader_f16));
     keys.extend(pool::all_pipeline_keys(shader_f16));
+    keys.extend(reduce::all_pipeline_keys(shader_f16));
+    keys.extend(resize::all_pipeline_keys(shader_f16));
+    keys.extend(softmax::all_pipeline_keys(shader_f16));
     keys
 }
--- a/wgpu/src/kernels/reduce.rs
+++ b/wgpu/src/kernels/reduce.rs
@@ -1,0 +1,82 @@
+use tract_core::internal::*;
+use tract_gpu::ops::reduce::Reducer;
+use tract_gpu::tensor::DeviceTensor;
+
+use crate::kernels::shaders::{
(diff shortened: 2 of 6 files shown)


czoli1976:
Correcting an offer in the description, since it would have sent you the wrong way.

I offered to make the shared `GpuReduce` honour all its axes instead of splitting them in the transform. I have since tried it: collapsing a contiguous axis run into one dispatch is trivial and numerically correct, and it was **71% slower**. The split is a parallel reduction in disguise — reducing W first gives `outer = C×H` threads, and merging both axes leaves `outer = C`, sixteen threads each summing 2304 elements.

So the extra dispatch is the cheaper of the two and this slice's workaround earns its keep. A single-dispatch version would need a workgroup-cooperative kernel with a tree reduction, which is separate work. Nothing needed from tract-gpu here after all.

czoli1976:
@kali wait

czoli1976:
ready ..... 
image

kali (approved):

czoli1976:
hmmmmmm

kali:
Looking at the reported A53 bug on the side, we can ignore it.

czoli1976:
CI failure on `linux (cortexa53-unknown-linux-musl)` is an unrelated flaky proptest, not from this PR.

Failing test: `ops::quant::scale::prop` in `tract-core` (quantized matmul, plain vs optimized path). This PR's diff is 6 files, all under `wgpu/` — no `core/` changes.

```
minimal failing input: a = 120, b = 101, scale = 692.5715
  plain:     1,1,I8 18
  optimized: 1,1,I8 17
```

A 1-LSB rounding disagreement. There's no pinned regression file for `ops/quant.txt`, so `prop` draws a fresh seed each run (26 successes, then this input). It failed only on the qemu-emulated cortex-a53 target — emulated aarch64 f32 rounding differs enough from the reference path to tip a value on a `.5` boundary. Every native target passed on the same run (apple darwin/ios, linux aarch64/armv7/riscv64/musl, wasm32, windows-arm64).

Could you re-run that job? It should pass on a new seed. If it reproduces, it's a `tract-core` quant rounding issue on that target, independent of this stack.

PR #2809, opened 2026-09-08
wgpu 6/7: fusion — chains, epilogues and axis ops

czoli1976:
6/7 of #2801. The graph-level work, and where the frame time actually goes. Slices 1–5 are merged; this is now a single commit on current main.

Three passes: a run of element-wise ops becomes one kernel; the ops trailing a GEMM or a convolution become that kernel's epilogue instead of another full pass over the result; and a shape-only axis op in front of an op that reads through strides becomes a change of strides rather than a copy.

The axis-op rules and `WgpuFusedAxisOp` follow metal's closely, with one divergence: a `DeviceSync` must stay a node of its own, because `rewire_syncs` and the coverage check both look for it by type, and fusing into it makes the model appear uncovered. Metal may well have the same latent hazard.

That closeness is worth naming: metal and cuda each already carry ~450 lines of `fuse_axis_op.rs` + `fused_axis_op.rs`, and this makes a third near-identical copy. It belongs in tract-gpu, parameterised by the backend's "reads through strides" op list. I did not do it here because it would touch cuda, which I have no way to run. Say the word and it becomes its own PR, before or after this one.

Three in-crate tests assert the fusions fire (one chain from a run of element-wise ops, a two-entry GEMM epilogue, a Move absorbed into `WgpuFusedAxisOp`); what the fused graphs compute is covered by the shared conformance suite from #2805.

🍍

### DIFF
--- a/wgpu/src/kernels/chain.rs
+++ b/wgpu/src/kernels/chain.rs
@@ -1,0 +1,87 @@
+//! Dispatch for a fused elementwise chain.
+
+use tract_core::internal::*;
+use tract_gpu::tensor::DeviceTensor;
+
+use crate::kernels::shaders::{
+    ChainOperand, ChainStep, EntryPoint, LayoutKind, ShaderDtype, broadcast_strides, chain_key,
+    chain_wgsl, pack_u32s, pad8_u32,
+};
+use crate::utils::{element_offset, get_wgpu_buffer};
+use crate::with_wgpu_queue;
+
+/// Extra operands a chain can read besides its head, bounded by what the
+/// 256-byte uniform slot holds.
+pub const MAX_EXTRA_INPUTS: usize = 3;
+
+pub fn wgpu_chain_dispatch(
+    steps: &[ChainStep],
+    inputs: &[&DeviceTensor],
+    output: &DeviceTensor,
+) -> TractResult<()> {
+    with_wgpu_queue(|q| {
+        ensure!(inputs.len() <= MAX_EXTRA_INPUTS + 1, "chain has too many inputs");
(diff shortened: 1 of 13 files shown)


czoli1976:
wait

czoli1976:
@kali  Ready

kali:
I think we merge this one, we see this epic through, but then we factorize this in gpu ? I'll find the time to handle what you can't on cuda.

kali (approved):

czoli1976:
> I think we merge this one, we see this epic through, but then we factorize this in gpu ? I'll find the time to handle what you can't on cuda.

You had 3 wishes; you have 1 more left.

kali:
> > I think we merge this one, we see this epic through, but then we factorize this in gpu ? I'll find the time to handle what you can't on cuda.
> 
> You had 3 wishes; you have 1 more left.

Well, I think your arithmetics need some proptest coverage.

czoli1976:
> > > I think we merge this one, we see this epic through, but then we factorize this in gpu ? I'll find the time to handle what you can't on cuda.
> > 
> > 
> > You had 3 wishes; you have 1 more left.
> 
> Well, I think your arithmetics need some proptest coverage.

noted, wait ...

czoli1976:
Ok will add the tests in a follow up then

kali:
that was a joke :) 3-1≠1

czoli1976:
> I think we merge this one, we see this epic through, but then we factorize this in gpu ? I'll find the time to handle what you can't on cuda.

Yes, these ~450 lines of duplicated code need to go at some point … I will give it a try

PR #2810, opened 2026-09-08
wgpu 7/7: take a frame in, hand a mask back

czoli1976:
7/7 of #2801. The two kernels a camera needs. **Stacked on #2809 — read this one's top commit only.**

RGBA8 pixels become an NCHW f32 tensor without a trip through the CPU, and a result is written into a storage texture the page already owns. Together they let a browser run a frame end to end without ever reading a tensor back, which is the case the whole backend exists for: on a video call, inference must not take the cores the encoder needs.

Drop this one if you would rather the browser plumbing stayed out of tree until there is a demo in tree to justify it — nothing in 1/7 through 6/7 depends on it.

One test: an RGBA8 buffer ingested and compared against the same conversion on the CPU.

🍍

### DIFF
--- a/wgpu/src/context.rs
+++ b/wgpu/src/context.rs
@@ -373,6 +373,7 @@ impl WgpuContext {
         buffers: &[&WgpuBuffer],
         uniform: &wgpu::Buffer,
     ) -> TractResult<wgpu::BindGroup> {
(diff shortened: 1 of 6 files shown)


czoli1976:
@kali  Rebased, this Rocket is ready to take off

kali (approved):

czoli1976:
image

PR #2811, opened 2026-09-08
wgpu 1/7: a WebGPU device for tract tensors

czoli1976:
First of seven slices of #2801, the WebGPU backend. This one is the device: a context that owns the `wgpu` device and caches its pipelines, layouts and bind groups, a tract tensor that lives in a `wgpu::Buffer`, and the copy and cast kernels. No model runs yet — that arrives in 2/7.

This is the slice worth your time, because everything in it is contact with the tract-gpu contract rather than WGSL:

- **The memory pool's arena cannot be used.** WebGPU refuses a dispatch that binds one buffer both read-only and read-write, which is what an arena does the moment a kernel reads one tensor and writes another in the same allocation: `Buffer with 'tract-wgpu-arena' usage ... is used as both a read-only storage and a writable storage`. So a tensor gets a buffer of its own from a pool that recycles by node and slot, and is addressed as `(buffer, byte offset in a uniform)`.
- **`DeviceBuffer::ptr()` has no meaning here.** There is no GPU address to hand out. It returns the buffer handle's identity, which is enough to key a cache and nothing else.
- **`get_bytes_slice` cannot report a failure.** On a unified-memory backend it is a slice; here it is a device readback that can fail, and the signature returns `Vec<u8>`. The path is unreachable today — only the arena view enters it, and this backend never installs one — and it is documented as such.

Uniform blocks are staged in one host buffer and uploaded once per flush rather than written per dispatch, which is worth about 20% of a frame on its own.

Three tests: an upload/download round trip, a strided copy against a CPU permute, and an f32→f16→f32 cast.

🍍

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -87,6 +87,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "android_system_properties"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ae221649c9976a6f6c56ae1facf410f3ddb33cc661c4b7b61020a912d4237fbc"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "anes"
 version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali (approved):

PR #2814, opened 2026-09-09
Add missing sign checks on loader dims casts (onnx/tflite)

li-jin-quan:
Both `onnx::tensor::load_tensor` and `tflite::tensors::flat_tensor_to_tract_fact` cast loader dims (`i64` / `i32`) straight to `usize` without a sign check. A crafted model with a negative dim turns into a huge `usize` and blows up the next allocation -- panic / host DoS.

Same pattern as the one already in `tensorflow/src/tensor.rs:44` -- bail on negative dims before the cast.

### Files

| file | change |
|---|---|
| `onnx/src/tensor.rs` | bail on `i < 0` in the dims map |
| `tflite/src/tensors.rs` | bail on `d < 0` in the shape map |

### Why safe

Negative dims are invalid in both formats. Mitigation sits at the loader boundary only -- no behavior change on well-formed models.

### DIFF
--- a/onnx/src/tensor.rs
+++ b/onnx/src/tensor.rs
@@ -160,7 +160,17 @@ pub fn load_tensor(
     let dt = DataType::try_from(t.data_type)
         .map_err(|e| format_err!("unknown ONNX TensorProto.DataType ({}): {e}", t.data_type))?
         .try_into()?;
-    let shape: Vec<usize> = t.dims.iter().map(|&i| i as usize).collect();
+    let shape = t
+        .dims
+        .iter()
+        .map(|&i| {
+            if i < 0 {
+                bail!("Negative dimension in ONNX TensorProto dims: {i}")
+            } else {
+                Ok(i as usize)
+            }
+        })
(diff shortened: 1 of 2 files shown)


kali:
I don't think we need the extra test. This is a pretty stable piece of code.

li-jin-quan:
Hey @kali — force-pushed a cleanup: the branch now sits on top of current main (the earlier extra commit was my own #2766 duplicated by mistake, removed). Same 24-line sign-check, no behavior change on valid models. 

CI is sitting at "action required" because the fork-PR workflow needs a maintainer click — could you approve the run when you have a sec? I'm around if anything else needs tweaking.

li-jin-quan:
Fixed a build error I introduced here: in tflite the shape iterator yields plain `i32` (it is a flatbuffers `Vector<'a, i32>`), so my extra `*` did not compile and clippy went red. My mistake, no test added as you said earlier. Force-pushed, same 24 lines, onnx side untouched. CI should be green now.

kali:
The metal fail is on us, the machine crashed.

kali:
Thank you!

PR #2815, opened 2026-09-09
nnef: every graph.nnef went through a template engine on the way in, …

kali:
…whether or not it held any template markers, and minijinja and its regex machinery cost 1.18MB of .text in a library that only loads models. Nothing in the tree templates a graph and the surface was never documented, so put it behind an off-by-default unstable-jinja feature, and name the module after the engine it has used since liquid was dropped.


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2820, opened 2026-09-09
(fix) wgpu: spread a large dispatch across two grid dimensions

czoli1976:
WebGPU caps a dispatch dimension at 65535 workgroups, which at 64 threads per group is 4194240 values. The kernels that walk a tensor linearly put every group on `x`, so any tensor past that failed validation outright — a model with 320x320 inputs hits it at the first convolution. The overflow now goes on `y`, and those kernels fold it back into their linear index.

`copy_past_one_grid_dimension` covers it; without the fix it fails with `wgpu error: Validation Error` rather than a wrong result.

🍍


kali (dismissed):

czoli1976:
@kali l'amour toujours

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali (approved):

PR #2821, opened 2026-09-09
(perf) wgpu 8/8: a blocked matmul over repacked weights

czoli1976:
Stacked on #2810 — its commits are included here, so read this one's top commit only.

A thread computes a 4x4 tile instead of a single value and reads both operands as `vec4`. That needs the second operand's `k` axis fastest, so it is transposed once on first use and kept.

**Measured on the selfie-segmentation model, natively (Metal): median 11.1% faster per frame, winning 15 of 16 interleaved A/B pairs** (3.213 ms vs 3.578 ms on a quiet machine). The one loss is a pair where both arms ran 3x slow under load. This is a *native* number — the browser is record-bound rather than GPU-bound on this model, and I have not measured a browser win here, so please don't read one into it.

The cache this needs is the interesting part. Weights are only worth transposing because they are read again every frame, which makes the packing worth keeping — but a cache keyed on a buffer is a trap when buffers are pooled and their ids recycled. So the packing lives on the queue that owns the frame rather than on the shared context (it holds a tensor from that thread's pool), the key names the buffer, its offset and the shape it was packed to, and `blocked_applies` takes the path only for a buffer the pool cannot recycle. That last condition is what makes "weights are constant" an invariant rather than a comment.

Conformance: `cargo test -p test-wgpu` — 4071 passed, 0 failed.

🍍

### DIFF
--- a/wgpu/src/context.rs
+++ b/wgpu/src/context.rs
@@ -441,7 +441,11 @@ impl WgpuContext {
     }
 
     pub fn wrap_storage(&self, inner: wgpu::Buffer) -> WgpuBuffer {
-        WgpuBuffer { inner, id: self.next_buffer_id() }
+        WgpuBuffer { inner, id: self.next_buffer_id(), pooled: false }
+    }
+
+    fn wrap_pooled_storage(&self, inner: wgpu::Buffer) -> WgpuBuffer {
+        WgpuBuffer { inner, id: self.next_buffer_id(), pooled: true }
     }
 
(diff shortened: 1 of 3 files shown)


kali (approved):

czoli1976:
@kali l'amour toujours

czoli1976:
@kali can you  re-run the failed Embedded job ?

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2823, opened 2026-09-10
GPU arena reuse and Metal buffer pool

JulienBalianSonos:
## Scope

This PR is narrowed to the GPU transient-memory arena and Metal buffer-pool work.

- B2: Metal exact-shape buffer pool for shared host/Metal buffer pairs, including command-buffer liveness so recycling waits until GPU work is complete.
- B3: session-lived arena storage reuse through `TurnState.shared`, without adding a new session resource channel.
- B4: default-on GPU arena behavior with `RunOptions::enable_gpu_memory_arena` as an explicit runtime override.
- Benchmark scaffolding for arena/pool comparisons on GPU backends.

## Runtime Behavior

- Metal prepares plans with the GPU arena enabled by default.
- CUDA keeps the existing hint-driven behavior unless `RunOptions::enable_gpu_memory_arena` explicitly overrides it.
- `None` for `enable_gpu_memory_arena` preserves each backend's default.
- `RunOptions::enable_gpu_memory_arena` is a per-prepare execution/memory policy, matching the existing role of `RunOptions`.
- The arena cache is stored in `TurnState.shared`, so it follows session lifetime without introducing a new runtime resource channel.
- Arena storage is reused only when the cached allocation is the sole owner; live arena views keep their backing allocation alive and force a fresh allocation for the next turn.
- Metal pooled buffers are keyed by datum type and exact shape, bounded by count and byte budget, and deferred while a command buffer is active.
- The Metal pool only recycles ordinary shared host/Metal buffers; exotic device tensors remain outside the pool.
- Sub-page Metal buffers use normal allocation rather than the pool.
- Memory sizing hints guide arena packing. Missing symbols get representative default values for packing order only, not correctness.
- The MLX conv layout rewrite only selects the MLX-only `OHWI` kernel layout for shapes the runtime MLX dispatch gate can accept; other convolutions keep the direct-kernel layout.

## Explicitly Out Of Scope

The previously bundled follow-ups were split to separate draft PRs:

- B5: #2855 keeps selected model outputs on device via an opt-in transform.
- B6: #2856 preserves simple device-backed `Tensor::slice` results as device views.

They are intentionally kept out of this PR because they change broader model-output and tensor-storage semantics.


czoli1976:
niiiiiice

api/rs/src/lib.rs
```diff
@@ -382,6 +364,24 @@ impl ModelInterface for Model {
...
+    /// ignores the declaration. Declaring an empty set clears a previous
+    /// declaration.
+    pub fn declare_device_resident_outputs(
+        &mut self,
+        outputs: impl IntoIterator<Item = usize>,
+    ) -> Result<()> {
+        tract_gpu::sync::declare_device_resident_outputs(&mut self.0, outputs)
+    }
+}
+
```
kali: This method with an iterator is not C-safe, we'll never manage to wrap it. I think this should just be a ModelTransform (just JSON for starters, no client side Config as we'll consider it experimental before making it explicitely public and commiting to the API).

api/rs/src/lib.rs
```diff
@@ -579,6 +562,54 @@ impl TensorInterface for Tensor {
...
+                    let host = dev.to_host()?;
+                    return Ok(Tensor(host.slice(axis, start, end)?.into_arc_tensor()));
+                }
+            };
+            return Ok(Tensor(sliced.into_tensor().into_arc_tensor()));
+        }
+        Ok(Tensor(self.0.slice(axis, start, end)?.into_arc_tensor()))
+    }
+}
+
```
kali: I'm very uncomfortable having this leave in api/rs. I think it should be moved to core, with the right hooks and/or trait so it can be overriden by DeviceTensor. Not sure how to do it.

kali:
mieux avec le storage ❤️

kali:
This feels a bit like a spike in the spike. Let's see how we can proceed... The global sentiment is, each one of this is a topic, so let's name them so we can discuss them, and maybe carve them out as a mergeable item.

Before any of it

Rebase onto main — the multi-output arena hunks already landed in 11c2dd431; they'll vanish and stop muddying the diff.
Drop the accidental peak_memory_terms doc deletion — rebase collateral.
Park commit_current — the in-flight command-buffer queue has no caller outside a test; it should travel with the PR that uses it.

B1 — lifetime precompute in eval_device_mem_req_for_nodes

Hoists the flush-list scan into a node-indexed Vec<Option<usize>> ahead of the main loop.

We can't see what it buys, so the default is to drop it. It's the same O(n²) — n nodes each scanning up to n flush buckets, the scan just moved — nothing reads end_by_node outside the loop immediately after it, and behavior is identical since build_flush_list puts each node in at most one bucket. If there is a reason (measured load-time cost of DeviceMemSchema::build on a big model), then it should come with that number and be the real fix: invert flush_lists once, or better, have build_flush_list return the node-indexed values_needed_until_step it already computes and currently throws away.

B2 — Metal exact-shape buffer pool

Recycles (host allocation, MTLBuffer) pairs, with recycling deferred until in-flight command buffers complete.

Good, and the soundest thing in the PR. The race fix is convincing. The open question is what it still catches once the arena is on — the big recurring allocation (KV cache) is state_owned, excluded from the arena, and changes shape every step, so it misses every lookup while filling the 512 MiB budget with dead buffers. Wants a pool-on/arena-on vs pool-off/arena-on A/B. Land first, independently.

B3 — session-lived arena storage cache

Keeps the arena's backing buffer across evaluations instead of reallocating it each run.

Right idea, wrong location. TurnState.shared already outlives turns — its own doc says so — so the new SessionShared core API isn't needed. And the one thing it adds, sharing across state clones, appears to make the cache never hit with two live states. Can we re-derive against turn.shared ? the core change should disappear.

B4 — arena on by default on Metal

Flips Metal from arena-iff-hints to arena-always, unlocked by defaulting missing hint symbols.

This is the actual payload, and the only slice with no structural objection. The soundness argument checks out — hints only drive packing order, never correctness. Two asks: evidence across real Metal models rather than two synthetic points before flipping a global default, and a decision on the fact that B2's pool and B3's cache are two caches for the same allocation. Land after B2/B3, as the smallest possible diff.

B5 — device-resident outputs

Lets declared outputs skip the ToHost sync and be fed back verbatim as next-step inputs.

A separate feature, not arena work, and blocked on one question. GpuDynKVCache already holds the cache on device across turns as op state, with state_owned keeping it out of the arena, and replace_kv_cache exists to fold an unfolded export. Before reviewing the mechanism we need to know why the cache is unfolded. Secondary: the input side short-circuits in eval against an ensureso the facts end up lying. Split out as its own PR with its own motivation section.

B6 — zero-copy slicing of device tensors

Makes Tensor::slice return a metadata view instead of a copy when the storage is device-backed.

The refactor to TensorStorage::slice is the right idiom — we're not there yet. It builds gappy-strided views that check_strides_validity exists to reject, then patches as_bytes with a gather so host read go through as_bytes.
This one needs the widest discussion because the reach is now 67 call sites.

---

Possible sequencing, B1 out, B2 → B3 → B4 as the arena story in that order.

B5 out entirely as separate PRs with their own motivation — neither is arena work, and B5's premise may dissolve on contact with GpuDynKVCache.

B6, let me think about it.

JulienBalianSonos:
thanks for the guidance. I will make a dent in it and come back to you once something more polished emerge (I need to clarify how it impact the sequence of PR to land fast MoE - if everything is still optimized with your proposal - and if prior benched speed remain unchanged).

JulienBalianSonos:
## Follow-up on B2, B3, and B4

### B2 — Metal exact-shape buffer pool

Measured on an M4 Pro with the arena enabled in both runs. Workload: 8 layers × 3 active experts, shape `1×32×2048`, 60 samples.

| Configuration | Median latency | Mean latency |
| --- | ---: | ---: |
| Arena + pool | 2.669 ms | 2.787 ms |
| Arena, pool disabled | 2.758 ms | 2.859 ms |

The pool reduces median latency by 3.2% in this workload. This supports it as bounded, completion-safe allocation reuse, but it is not the primary throughput claim.

### B3 — session-lived arena storage

Reworked to use the existing `TurnState.shared` resource store. The new core-level `SessionShared` API is removed.

The backing storage is reused across sequential turns of one state. State clones do not share the cache, avoiding the contention and cache-miss behavior discussed in the review.

### B4 — Metal arena enabled by default

Real-model validation on an M4 Pro with the Qwen 3 8B f16 NNEF export and the Metal runtime, without causal-LLM sizing hints:

| Policy | PP512 | TG128 |
| --- | ---: | ---: |
| Previous arena-if-hinted policy | 389.3 tok/s | 13.0 tok/s |
| Default-on arena | 452.9 tok/s | 14.1 tok/s |
| Improvement | 16.3% | 8.5% |

The tracked MoE-shaped end-to-end benchmark also improves with the arena:

| Configuration | Median latency |
| --- | ---: |
| Arena disabled | 3.328 ms |
| Arena enabled | 2.904 ms |
| Improvement | 12.7% |

### Long-context follow-up

The long-context benefit was measured (in original moe pr) as more significant: larger prompt contexts create larger transient allocations, (compute and attention bandwidth dominate in current tract after some point but other optims were made to remedy it that should land after in the MoE PR breakdown PR sequence ).

PR #2829, opened 2026-09-10
(perf) core/nn: integer nearest upsample in one pass

TarikAlHadethi:
First slice of #2771: the nearest upsample lowering, on its own. It replaces the Reshape/Tile/MultiBroadcastTo/Reshape chain with a single `NearestUpsample` op, so the broadcast intermediate is never materialised.

One file, `core/src/ops/nn/resize.rs`. None of the kernels from #2771 are here. This is the lowering only, it does not depend on the rest of that PR and the rest of that PR does not depend on it.

### Numbers

i7-12700K, Windows, `RAYON_NUM_THREADS=1`, `-O`. Baseline is this branch's merge-base, built and measured on the same box tonight.

Five interleaved rounds, alternating the two binaries with no build between them. Worth saying why: running all the baseline first and all the branch second moved the *same* branch binary by 25 ms, so that ordering measures the machine as much as the patch.

| | base | this PR |
|---|---|---|
| tiny_det 1x3x640x640 | 337.5 ms | **310.0 ms** |

1.09x. Base runs span 336.4 to 339.0, PR runs 304.2 to 314.5, no overlap.

### tiny_det, ops that moved

| op | base | this PR | delta |
|---|---|---|---|
| MultiBroadcastTo | 43.906 | 16.647 | **-27.3** |
| Reduce&lt;Sum&gt; | 4.460 | 1.436 | -3.0 |
| MoveAxis | 0.509 | - | gone |
| NearestUpsample | - | 2.933 | new |

Everything else is inside a 1.3 ms band.

### Structural

- 6 `Resize.N.tile` MultiBroadcastTo nodes become 6 `Resize.N.nearest` NearestUpsample nodes
- The one surviving MultiBroadcastTo is `ConvTranspose.0.broadcast_bias`, as before
- 3 `Conv.*.restore_out_*` MoveAxis nodes go entirely

The node counts drop past the ops that changed: IntoShape 80 to 69, OptAddUnicast 30 to 26, OptMulByScalar 31 to 27, OptMaxByScalar 15 to 11, OptAddByScalar 11 to 7. Not inserting the unit axes in the first place lets the optimiser fold about 27 nodes it could not fold around the reshape pair.

### Two changes from the version on #2771

The doc comment described what the op replaces; it states the contract now. The generic fallback opened with `.to_vec()` on the whole input, which is unnecessary once the `PlainView` from `try_as_plain` is given a name, so that copy is gone.

### Numerics

The op itself is exact. On a model containing nothing but the nearest `Resize`, this branch and the merge-base are **bit identical**, and both are bit identical to an independent `numpy.repeat` reference.

End to end on tiny_det the outputs are not bit identical: **max abs 2.9e-6** on a sigmoid map in 0 to 0.999, mean abs 1.0e-7. None of that comes from `NearestUpsample`, which only copies. It comes from the folding the change unblocks, since reassociating a chain moves the last few bits. For scale, the figure quoted on #2771 for the whole series was 2.4e-5.

### One thing to know before it is benchmarked elsewhere

`expand_row_f32` has a NEON store-interleave for scale 2 and falls to scalar on x86, so the number above is the scalar path. An AVX2 row expand is a leaf kernel, and per your rule of thumb that belongs in `routines` rather than here, so I have left it for a later slice instead of widening this one.

🍍

### DIFF
--- a/core/src/ops/nn/resize.rs
+++ b/core/src/ops/nn/resize.rs
@@ -1,5 +1,4 @@
 use crate::internal::*;
-use crate::ops::array::Tile;
 
 /// Maps an output coordinate back to the input axis. The ONNX coordinate
 /// transformation modes that have a well-defined inverse without an input ROI.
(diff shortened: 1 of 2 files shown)


core/src/ops/nn/resize.rs
```diff
@@ -489,71 +642,13 @@ pub fn lower_nearest_integer_upsample(
...
-        )?[0];
-    }
-
-    wire = patch.wire_node(
-        format!("{}.tile", node.name),
-        Tile { multipliers: tile_multipliers },
+    let wire = patch.tap_model(model, node.inputs[0])?;
+    let wire = patch.wire_node(
+        format!("{}.nearest", node.name),
+        NearestUpsample { scales: int_scales.iter().cloned().collect() },
```
kali: I think this is a one-liner with TypedModelPath::replace_single_op

kali (dismissed):

TarikAlHadethi:
Pushed as `a98e201`, amended into the one commit. The force-push dismissed your approval, so this needs another look.

The lowering is `TypedModelPatch::replace_single_op` now. One side effect worth knowing: the node keeps the name of the `Resize` it replaces rather than getting a `.nearest` suffix.

The red `test-nnef-cycle` was real and mine. `NearestUpsample` had no NNEF dumper, so `test_resize_upsample_scales_nearest` failed with "No serializer found". It now dumps as the stdlib `nearest_upsample` when its first two scales are 1, which loads back as a `Resize` that declutters to `NearestUpsample` again. Anything else goes out as `tract_core_resize`, since `nearest_upsample`'s factor only covers the axes from 2 on.

Locally `test-nnef-cycle` is 2472 passed and 0 failed, up from 2471 and 1, and CI is green now. Both paths round trip bit exact and land back on `NearestUpsample` after reload. The fallback needed its own check, scales `[1, 2, 1, 2]`, because nothing in the suite reaches it.

On the nemotron load row: the bot's two runs here put it at +27.8% and then +11.1%, and the only changes between them are a node name and a serializer, neither of which runs on load. A nemotron preprocessor load row also flagged +24.4% on #2822, which only added a runner. Locally tiny_det loads and optimises in 252 ms against 271 ms on main, interleaved. So I read it as noise rather than a load cost.

kali:
Don't worry too much about benches, unless i point something out. They are a bit noisy. Working on it.

kali (approved):

PR #2830, opened 2026-09-10
(perf) core/cnn: vectorise NCHW MaxPool 2x2

TarikAlHadethi:
Second slice of #2771: NCHW MaxPool 2x2. The kernel lives in `core/src/ops/cnn/maxpool.rs` and is purely additive: only 2x2 f32 NCHW pools with unit dilation take the new path, and every other pool runs the existing `visit_output` code untouched. The test is a `max_pool` proptest in test-rt/suite-unit.

I took this one second on purpose, because it is the smallest diff that puts the core/linalg seam in front of us. So I measured that question rather than argue it.

### The number, and it is not where I expected

i7-12700K, Windows, `RAYON_NUM_THREADS=1`, `-O`. Three binaries off the same merge-base, built and measured on the same box, profiled in interleaved rounds. `OptMaxPool` on tiny_det 1x3x640x640:

| build | OptMaxPool | vs base |
|---|---|---|
| merge-base | 8.82 ms | |
| loop reorder only, 2x2 f32 specialisation removed | 10.46 ms | **+1.6, slower** |
| this PR | **2.55 ms** | **-6.3, 3.5x** |

Three rounds each, spreads of 0.06 ms or less: base 8.694/8.918/8.855, loop-only 10.383/10.565/10.418, this PR 2.525/2.581/2.557. End to end tiny_det moves about -6 ms, which matches the op.

**The NCHW loop reorder on its own is a regression.** All of the win comes from the leaf kernel, so the reorder now only runs under it. An earlier push also sent every other NCHW 2D pool through the reordered loop, and the bench caught it on inceptionv1q: inception v1's pool shapes nearly doubled (u8 7.65 to 14.69 ms). That loop is gone, those pools are back at main's speed (7.88 ms against 7.65), and tiny_det keeps its win (10.44 to 3.07 ms in the same rounds).

### Where the kernel belongs

That makes this change a target-dependent leaf kernel and nothing else: one kernel wins outright when it applies, with no shape to weigh, no pool and no tier ladder. You settled it: it stays in tract-core for now as temporary platform-dependent code, recorded on #2836 with the other kernels to move to `routines`.

### Something to fix when it moves

The AVX block calls `_mm256_*` behind a runtime `is_x86_feature_detected!` but the enclosing function carries no `#[target_feature(enable = "avx")]`, unlike every other x86 kernel in linalg (`x86_64/by_scalar.rs:17` and the rest). The intrinsics will not inline like that, so 2.55 ms is probably not the floor. It is noted on #2836, so it gets fixed on the way to `routines`.

### Correctness

Base and this PR are **bit identical** on tiny_det, every one of 409,600 outputs, same fixed input. Max is exact, so there is no reassociation to hide behind: a difference here would be a defect, not rounding.

The test is a proptest in test-rt/suite-unit, so it runs on every runtime: all four layouts, 1D to 3D, Valid, SameUpper, SameLower and explicit pads, strides and dilations. Half the cases are 2x2 NCHW with rows up to 17, because drawn uniformly that shape almost never comes up and the AVX body needs rows longer than 8. Dropping one tap from the AVX max makes it fail, so it does reach the vector body. The five shapes from the original test are fixed cases. NNEF cannot express SameLower, so nnef-cycle and f16 filter it out, and tflite keeps what its pool options can express.

### One divergence I found and closed

The 2x2 border loop seeded its running max with `f32::NEG_INFINITY`, where the fold it replaces uses `T::min_value()`. A window with no in-bounds tap would have written `-inf` where the existing code writes `f32::MIN`. Nothing in tiny_det produces one, and the first test could not catch it because its own reference seeded the same way. Both are `f32::MIN` now, so the border matches the fold exactly.

### Still divergent, and I do not think it can be fixed cheaply

`_mm256_max_ps` and `vmaxq_f32` take the second operand when either is NaN, while the scalar fold's `<` drops NaN entirely. So a NaN input can come out differently on the SIMD path. Guarding it would cost the kernel its point. Flagging rather than fixing, since a NaN reaching a maxpool means something upstream already went wrong.

### Correction to my last comment

I wrote `DepthwiseMF32`. It is `DepthwiseWF32`.

🍍

### DIFF
--- a/core/src/ops/cnn/maxpool.rs
+++ b/core/src/ops/cnn/maxpool.rs
@@ -147,6 +147,13 @@ impl OptMaxPool {
         geo: &ConcretePoolGeometry,
     ) -> TractResult<TVec<TValue>> {
         let input_dt = input.datum_type();
+
+        if self.with_index_outputs.is_none()
+            && let Some(values) = self.try_nchw_2x2_f32::<T>(input, geo)?
+        {
+            return Ok(tvec!(values.into_tvalue()));
+        }
+
         let input_plain = input.try_as_plain()?;
         let input: ArrayViewD<T> = input_plain.to_array_view()?;
         let input_ptr = input.as_ptr();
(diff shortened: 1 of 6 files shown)


core/src/ops/cnn/maxpool.rs
```diff
@@ -248,4 +467,94 @@ mod tests {
         let opt = optimized.into_runnable().unwrap().run(input).unwrap();
         assert_eq!(*opt[0], *plain[0]);
     }
+
+    fn nchw_maxpool(
```
kali: Could this be put in the test-rt/test-unit and turn into a proptest ? it would benefit all runtimes.

kali:
Ok to merge "as is" (as in "temporary platform dependent in tract-core"). I hope you keep a close record of the "need-to-move-to-linalg" routines we leave on the side of the road.

kali (dismissed):

TarikAlHadethi:
Sorry, that push dismissed the approval you'd just given, I hadn't seen it land. It carries the proptest you asked for: `max_pool` in test-rt/suite-unit, and the core test is gone.

It covers all four layouts, 1D to 3D, Valid, SameUpper, SameLower and explicit pads, strides and dilations. Drawn uniformly, a 2x2 NCHW pool comes up about once in a thousand cases and the AVX body needs rows longer than 8, so half the cases are that shape with rows up to 17. The five shapes from the old test are fixed cases. Dropping one tap from the AVX max makes the proptest fail on raw, decluttered and optimized, so it does reach the vector body. It passes here on unit-core, nnef-cycle, f16 and wgpu; Metal and tflite I can't run on this machine.

Two filters, both for things a serialiser can't express. NNEF bails on SameLower, which also hits f16 since nnef_f16 goes through it. tflite only takes 2D, Valid or SameUpper, and a batch axis, and its pool options drop a dilation silently rather than refusing it, so dilated cases are filtered there too.

It also never builds a window that's all padding, because the runtimes disagree there: CPU and wgpu give f32::MIN, the Metal kernel gives -inf (its comment says the CPU does too).

For the record: I've opened an issue listing the kernels these slices leave in core and what each one needs to become a routine. It's linked above.

kali:
@TarikAlHadethi ok, this time, i'm pretty sure the bench is signal not noise. Are you in a position to repro ?

TarikAlHadethi:
Reproduced, and it's mine. inceptionv1q's pools run on quantized tensors, so none of them is the 2x2 f32 case, and this PR sent every NCHW 2D pool without an index output through the new (n, c, y, x) loop. That loop is slower than visit_output for them. My own tiny_det numbers had already shown the loop alone losing to visit_output, and I still left it catching every other shape.

The quantized model isn't on this machine, so I rebuilt its workload: a model holding inception v1's nine max pool shapes, in u8 and in f32. OptMaxPool total, one thread on the i7-12700K, median of 5 interleaved runs:

| | u8 | f32 |
|---|---|---|
| main | 7.65 ms | 8.88 ms |
| PR before this push | 14.69 ms | 14.71 ms |
| this push | 7.88 ms | 8.76 ms |

The fix narrows the fast path to 2x2 f32 with unit dilation and deletes the generic loop, so every other pool runs visit_output exactly as on main. That is what the commit message already claimed; now the code does it. tiny_det keeps the win (OptMaxPool 10.44 ms on main, 3.07 ms now), and the proptest still covers the other shapes through visit_output.

kali:
For future reference (this is totally undiscoverable, it's on me.) "evaltime" is the one that matters.

```
% cargo run -p tract-cli --features bench-suite  --release -- bench-suite --manifest ./.travis/benches.toml --filter inceptionv1q ; cat metrics
    Finished `release` profile [optimized] target(s) in 0.15s
     Running `target/release/tract bench-suite --manifest ./.travis/benches.toml --filter inceptionv1q`
Warning: could not lock GPU clocks (need privilege?)
  inceptionv1q pass
wrote 8 metrics to metrics
net.inceptionv1q.active_at_before_optimize.pass 7780418
net.inceptionv1q.active_at_model_ready.pass 8965297
net.inceptionv1q.evaltime.pass 0.05752289
net.inceptionv1q.rsz_at_before_optimize.pass 31694848
net.inceptionv1q.rsz_at_model_ready.pass 42160128
net.inceptionv1q.time_to_before_optimize.pass 0.066
net.inceptionv1q.time_to_model_ready.pass 0.213
bundle.bench_runtime 5
```

kali (approved):

PR #2831, opened 2026-09-10
(perf) core/cnn: unpack 2x2 stride-2 NCHW deconv along W

TarikAlHadethi:
Third slice of #2771: 2x2 stride-2 NCHW deconv. One file, `core/src/ops/cnn/deconv/deconv_sum.rs`, +205/-2, and everything that does not match the guard falls through to the generic loop untouched.

### Numbers

i7-12700K, Windows, `RAYON_NUM_THREADS=1`, `-O`. Baseline is this branch's merge-base built on the same box tonight, profiled in interleaved rounds. `DeconvSum` on tiny_det 1x3x640x640:

| build | DeconvSum |
|---|---|
| merge-base | 5.71 ms |
| this PR | **1.54 ms** |

**3.7x.** Three rounds each: base 5.770/5.622/5.742, this PR 1.552/1.562/1.510.

### Why it is this shape and not a general one

The generic scatter walks the output channel innermost, and on NCHW that stride is `H*W`, so consecutive writes land a plane apart and each one spends a cache line on a single float. The guard picks the one case with no overlap to reason about: 2x2, stride 2, unit dilation, no leading pad, contiguous W, output exactly twice the input. There each output pixel takes exactly one tap, so the four gemm planes interleave straight into even and odd rows and columns with nothing to accumulate.

### Correctness

**Bit identical** on tiny_det, all 409,600 outputs, same fixed input. That is a real check here rather than a lucky one: the guard admits only the non-overlapping case, so there is no summation whose order could have changed. If this had come back merely close, the guard would have been wrong.

The test lives in the suite-unit deconv proptest, so it runs on every runtime: the strategy draws this shape half the time, since drawn uniformly it comes up about once in fifty thousand cases, and two fixed cases keep the original geometry with distinct integer taps, so a tap read from the wrong plane changes the answer. Breaking one tap offset fails both.

### On the seam

Same reading as the maxpool slice: a target-dependent leaf with no shape to weigh. Per your call there it stays in tract-core for now, recorded on #2836 with the rest.

🍍

### DIFF
--- a/core/src/ops/cnn/deconv/deconv_sum.rs
+++ b/core/src/ops/cnn/deconv/deconv_sum.rs
@@ -76,14 +76,23 @@ impl DeconvSum {
         if !self.pool_spec.data_format.has_n() {
             tensor.insert_axis(0)?;
         }
-        eval(
+        if !try_fast_nchw_2x2_s2_f32(
             self,
             &input_shape,
             &output_shape,
             &spatial_output_details,
             &n_o_hkwk_hw,
             &mut tensor,
-        )?;
+        )? {
+            eval(
+                self,
+                &input_shape,
+                &output_shape,
+                &spatial_output_details,
+                &n_o_hkwk_hw,
+                &mut tensor,
+            )?;
+        }
         if !self.pool_spec.data_format.has_n() {
             tensor.remove_axis(0)?;
         }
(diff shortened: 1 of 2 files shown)


TarikAlHadethi:
Moved the test into the suite-unit deconv proptest, same treatment as the maxpool PR: the strategy draws the 2x2 stride-2 NCHW shape half the time, since drawn uniformly it comes up about once in fifty thousand cases, and the original geometry stays as two fixed cases with distinct integer taps, so a tap read from the wrong plane changes the answer. Breaking one tap offset fails both the fixed cases and the proptest. Passes here on unit-core, nnef-cycle, f16 and wgpu.

Also dropped an unused parameter and an empty cfg block from the kernel on the way; the hot code is unchanged, so the numbers in the description stand. The kernel is on #2836 with the others.

kali (approved):

PR #2843, opened 2026-09-12
Update stale snipsco/tract URLs in crates metadata and example READMEs

### LINKED ISSUE #1934: snipsco repository URL is still everywhere
I was confused by https://crates.io/crates/tract for a while, until I realized that it really is this same project. The source tree has the old URL in many many places, and the crates.io metadata has it too.

ajianaz:
The crates' `repository` metadata and four example READMEs still pointed at `snipsco/tract`, which now only redirects to `sonos/tract`; update all 26 occurrences (22 Cargo.toml files, 4 example READMEs, one line each).

Closes #1934.

🍍


kali (approved):

kali:
Thank you

PR #2846, opened 2026-09-13
(fix) wgpu: tanh returns NaN for a large argument

czoli1976:
WGSL's built-in `tanh()`, as naga lowers it for this backend, overflows to NaN once its argument's magnitude gets large enough that the underlying `exp()` overflows. `f32::tanh` (CPU), `metal::precise::tanh` and CUDA's native `tanh` all saturate correctly instead — this op is the only one of the four hand-written rather than delegated to a vendor intrinsic or the language std.

Reachable directly (`tanh` of a large tensor), and reachable through ordinary-looking values once a GEMM epilogue or even a short chain of amplifying ops (`square, square`) feeds it something large. Fix is a clamp to ±15 before the call — `tanh` already rounds to `1.0` in f32 well inside that range, so this changes nothing observable except that it stops returning NaN.

Follow-up to #2809, where I told kali I'd send proptest coverage for the fusion arithmetic. Three proptests here check what a fused element-wise chain, a GEMM epilogue, and a Move-then-chain actually compute against the CPU reference, rather than only asserting the rewrite fired (as the existing structural assertions do). The GEMM-epilogue one is what found this bug — any activation drawn against a matmul-scale accumulator reaches the overflow a small standalone tensor never would.

~~A third proptest for the GEMM-epilogue path found a second, harder issue — reproducible only through a sequence of prior proptest cases in the same process, not in isolation~~ — chased that down: it wasn't a second bug. It was this same tanh overflow, sampled inconsistently because proptest draws a fresh random seed each run and the failure depends on whether an activation gets drawn *and* whether the accumulator happens to cross the overflow threshold for that draw — not on any cross-call state. Confirmed with ~22,000 cases across several sweeps: reliably fails within the first ~5 draws without the fix (tanh is drawn about 1-in-8 of the time and overflows readily at matmul scale), zero failures across all of them with it.

Conformance: `cargo test -p test-wgpu` — 4095 passed, 0 failed.

🍍


kali (approved):

kali:
i think the mac is disturbed right now, i'm using it to fix the asan checks. no reason to worry.

PR #2850, opened 2026-09-14
core: QuantizeLinear rounded ties away from zero instead of to even

yuefdev:
ONNX QuantizeLinear rounds `x / y_scale` to nearest with ties to even, but `quantize_linear_f32_u8` / `quantize_linear_f32_i8` used `f32::round`, which rounds ties away from zero. #2672 fixed the same thing for DynamicQuantizeLinear. This switches the two helpers to `round_ties_to_even`. They are only used by `QuantizeLinearU8` / `QuantizeLinearI8`, which only the ONNX importer creates.

With scale 0.5 and zero point 0, inputs `[0.25, 0.75, 1.25, 1.75]` give `[1, 2, 3, 4]` on main and `[0, 2, 2, 4]` with this change, which matches onnxruntime and `onnx.reference`. The new `onnx/test_cases/quantizelinear_ties_to_even` case checks positive and negative ties for both u8 and i8 outputs. It sets `IGNORE=nnef` like the other quantized cases, because `QuantizeLinearU8` has no NNEF serializer.

`cargo test` on tract-core, test-unit-core, test-onnx-core, test-nnef-cycle and test-f16 still passes, and so do the q* cases in `onnx/test_cases`.

🍍


kali (approved):

kali:
Thank you!

yuefdev:
> Teşekkürler!

🫡

PR #2855, opened 2026-09-14
B5: device-resident outputs

JulienBalianSonos:
## Scope

Adds an opt-in device-resident I/O declaration for application-owned recurrent state.

- Selected model outputs can skip automatic GPU-to-host sync.
- Matching model inputs get an explicit boundary op that accepts either an initial host tensor or a previously returned device tensor.
- Ordinary host-to-device sync remains strict; pass-through is only available on explicitly declared device-resident inputs.

That lets a declared output be fed back verbatim as a next-step input without paying a host round-trip. The feature is independent of the GPU arena and buffer-pool work.

## Motivation

`DynKeyValueCache` already covers the common stateful decode path: it folds explicit `Source + Concat + model output` cache wiring into op state, the GPU transform lowers it to `GpuDynKVCache`, and its `state_owned` output fact keeps the cache out of ordinary arena reuse.

The separate use case is an application-owned external state loop:

1. The model returns logits plus trailing state outputs.
2. The application keeps those state tensors as the session state.
3. The application feeds those same tensors back on the next step.

A downstream LLM inference runtime may deliberately use that mode even after tract has recognized and folded exported past/present cache I/O into `DynKeyValueCache`: it can unfold the cache again so the application owns those tensors outside tract op state.

That supports application-level workflows such as:

- freezing and restoring sessions into a prefix-cache tree,
- truncating caches for partial prefix reuse,
- rolling back rejected speculative-decoding drafts,
- supporting fixed-size hybrid-attention states that are not growing KV caches but still need output-to-input feedback.

Those workflows are awkward or unsupported with opaque tract op state unless the op-state API exposes equivalent clone/truncate/freeze/restore semantics.

On GPU backends, the external-state loop has a bad default cost: every cache output is synchronized back to host at the model boundary, then the next step immediately syncs it back to device. For LLMs this can mean dozens of full-cache readbacks per token. Device-resident I/O preserves the external-state semantics while avoiding that unnecessary output-boundary transfer.

## Behavior

- `gpu_device_resident_io` declares device-resident input and output indices.
- `gpu_device_resident_outputs` remains available for output-only compatibility.
- Declared outputs skip automatic `ToHost` insertion.
- Declared inputs use `ToDeviceOrPass`, which sends host tensors to the device and forwards already-device tensors unchanged.
- Non-declared inputs still use strict `ToDevice`; a device tensor does not silently pass through an ordinary host-to-device boundary.

This makes the feature an explicit model I/O contract for application-owned recurrent state, not a replacement for `GpuDynKVCache`.

## Open Design Questions

- Is `gpu_device_resident_io` the right public transform/API shape, or should output-only compatibility remain the only named transform?
- Should output/input selection remain index-based, or should the transform resolve and store outlets?
- Should the transform contract require applying it immediately before backend transform?
- How much CUDA/WGPU coverage is expected beyond the shared sync path and Metal smoke tests?


kali:
I think this barely need to be stacked on 2823 at all, but stands on its own. Can you rebase against main ?

PR #2857, opened 2026-09-14
(perf) wgpu: let a GEMM or convolution epilogue grow past its bias

czoli1976:
The GEMM and convolution epilogue rule fired once, and only on a direct successor: once the bias was folded in, the activation behind it stayed a launch of its own, and a reshape between a GEMM's 2-D result and the convolution reading it kept even the first step out. Absorption now appends to an existing epilogue (operand indices renumbered past the ones already there) and steps over `Reshape`/`Add`/`Rm` axis ops — the ones that leave every value in place — re-wiring them after the fused op so its output keeps its own shape. `Move` is deliberately excluded.

On the selfie-segmentation landscape model this removes every standalone hardswish and relu launch: 113 → 96 dispatches per frame, throughput 1.63–1.68 → 1.56–1.58 ms in Chrome on an Apple GPU (4 steady-state rounds, all below all baseline rounds), output unchanged. `test-wgpu` conformance: 4104 passed, 0 failed.

🍍


czoli1976:
@kali l'amour toujours

kali (dismissed):

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali:
i've messed up main, these onnx stuff should not be here. fixing it.

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali (approved):

PR #2858, opened 2026-09-14
(perf) wgpu: read a 3x3 depthwise convolution's taps without staging them

czoli1976:
Every depthwise convolution staged its workgroup's input halo and filter in workgroup memory (the tfjs port). For a 3×3 there is too little reuse across taps for that staging and its barrier to repay themselves, and the overlap between neighbouring outputs is served from cache — so a 3×3 now reads its taps directly, from a workgroup half as wide across, and drops the workgroup arrays. Wider kernels keep the tile.

The gate is measured, not guessed (tiled vs direct, µs, two shape families, Apple GPU):

| kernel | 32ch 36×64 | 96ch 9×16 | |
|---|---|---|---|
| 3×3 | 1.22× | 1.17× | direct wins |
| 5×5 | 1.07× | 0.76× | mixed |
| 7×7 | 0.97× | 0.93× | tile wins |
| 9×9 | 0.94× | 0.95× | |
| 11×11 | 0.89× | 0.82× | |

On the selfie-segmentation landscape model (6 of its 11 depthwise layers are 3×3): 1.15–1.33× per 3×3 layer, 331.7 → 285.1 µs of depthwise time per frame across all 11, and 1.49 → 1.45 ms end-to-end throughput in Chrome (7/8 interleaved pairs). Output matches the CPU reference. `test-wgpu` conformance: 4104 passed, 0 failed.

🍍


czoli1976:
@kali l'amour toujours

kali (approved):

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2859, opened 2026-09-14
(perf) wgpu: sum a trailing run of axes in one cooperative launch

czoli1976:
A global pooling reached the device as one launch per reduced axis, since the shared reduce op handles a single axis with a serial loop per output — and merging the axes into that kernel was measured 71% slower, because a small outer extent starves the device. A sum over every axis from some point to the last is now kept whole by `split_multi_axis_reduce` and run as `WgpuSumRun`: one 256-thread workgroup per output value, threads striding over the reduced values and meeting in a tree over workgroup memory. Taken only when at least 64 values feed each output.

On the selfie-segmentation landscape model (ten squeeze-excitation poolings): 91 → 81 dispatches per frame, throughput 1.51–1.53 → 1.49–1.52 ms in Chrome on an Apple GPU (interleaved A/B, 3/4 pairs). The tree adds in a different order from the CPU, so the new test compares under `Approximate`. `test-wgpu` conformance: 4104 passed, 0 failed.

🍍


czoli1976:
@kali l'amour toujours

kali (approved):

PR #2862, opened 2026-09-14
(perf) wgpu: fuse a squeeze-excitation gate into one kernel, and make it fast

czoli1976:
A MobileNetV3 squeeze-excitation gate reached the device as six launches over a few dozen values: the mean's 1/HW multiply, two single-row products with a bias and relu between them, an axis move, and a bias-plus-sigmoid chain. Three commits fold that into one `WgpuGemvPair` kernel, then make the kernel fast:

1. **fuse a squeeze-excitation gate into one kernel** — `Gemm(m=1) → [reshapes] → bias/relu → [reshapes] → Gemm(m=1)` becomes one op, one workgroup, the hidden vector held in workgroup memory across a barrier. Registered in `can_fuse_move` so the `Move` in front of it keeps folding into strides instead of materialising a copy.
2. **fold the gate's normalisation and trailing activation into the gemv pair** — the op gains an input scale read on load (the mean's 1/HW), trailing elementwise steps run per output, and an explicit output shape; a rule folds the `Mul` before and the chain after, stepping over axis ops on tensors with a single non-unit axis in both directions. 140 → 113 dispatches per frame on the selfie-segmentation landscape model, throughput 1.79–1.89 → 1.63–1.68 ms in Chrome on an Apple GPU (4 steady rounds each, all below all).
3. **unroll both of the gemv pair's reductions** — ten steps per iteration with a scalar tail, the second into ten accumulators. Only a few dozen of the workgroup's threads have a row at all, so the win is instruction-level parallelism inside them: 1.85× across the nine gate shapes (2.35× at 128 wide, 1.13× at 16), 1.66 → 1.60 ms end to end, 8/8 interleaved pairs. Output byte-identical.

`test-wgpu` conformance on the stacked result: 4107 passed, 0 failed.

🍍


czoli1976:
@kali l'amour toujours

kali (dismissed):

PR #2874, opened 2026-09-15
The facade parsed ONNX unconditionally, so a consumer loading only NN…

kali:
…EF still built and linked tract-onnx and its operator register, and reached tract-core through it. Put the parser, the Onnx loader and the InferenceModel and InferenceFact types behind a default-on onnx feature, and take tract-core directly. enable_onnx is untouched: the ONNX operators an NNEF graph carries live in tract-onnx-opl.


kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2875, opened 2026-09-15
(perf) wgpu: resize the two trailing axes in one launch, plans on the device

czoli1976:
A `GpuResize` over height and width ran one launch per axis with an intermediate between them, and re-uploaded its index and weight plans on every frame: on the selfie segmenter that was 12 of the 14 buffers created per frame, and 6 of its 81 launches.

`WgpuResize2d` takes over when the plan covers exactly the two trailing axes. The plans become device constants wired as inputs, and each output gathers its `window_h x window_w` taps straight from the input. Bilinear resizes get a dedicated kernel with the two taps unrolled over an 8x8x2 workgroup and a three-dimensional grid; anything else (nearest, cubic, other windows) goes through the generic fused gather.

The bilinear kernel came out of an evolutionary search scored **in Chrome** rather than under naga: the same generic WGSL is 37 µs under naga/Metal and 97 µs under Tint for the largest upsample, and what Tint punishes is the one-dimensional `%`/`/` index mapping. Under Tint, the segmenter's three upsamples (16c 36x64→72x128, 16c 18x32→36x64, 24c 9x16→18x32):

| | µs / frame |
|---|---|
| two launches per axis (before) | 172 |
| fused generic gather | 141 |
| fused bilinear kernel | **21** |

LiteRT.js does the same three in 16–20 µs. In the whole-frame GPU replay of the segmenter the resize category went from 105–233 µs to 26.

Tests: a model-level test runs linear, nearest and cubic resizes over the fused path against the CPU; `cargo test -p tract-wgpu`, `cargo test -p test-wgpu` (4128 passed) green.

🍍


czoli1976:
@kali l'amour toujours

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

kali (approved):

PR #2876, opened 2026-09-15
(perf) wgpu: a per-pixel kernel for dense 3x3 convolutions

czoli1976:
The generic convolution runs one thread per output element, so the 16 channels of a stem pixel each walked the same 27 inputs. On the selfie segmenter the stem is a single launch that cost 115 µs, four times what LiteRT spends on the same layer, and the largest single item in its frame.

The per-pixel kernel applies to a single-group 3x3 at dilation 1 in channels-first layout, and gives one thread every output channel of its pixel: the window is read once per four channels, the taps are unrolled, and the four accumulators are named rather than indexed — Tint keeps an indexed private array in memory, and that alone costs a fifth under Chrome (naga is indifferent). It stays general over batch, input channels, stride and padding.

The kernel came out of an evolutionary search over tract's own conv contract (bindings, `Params`, epilogue), 22 rounds / 365 variants, then re-timed under Tint before shipping:

| stem 3→16, 144x256→72x128, s2, bias+hardswish | native (naga) | Chrome (Tint) |
|---|---|---|
| generic `conv2d` | 115 µs | 110–125 µs |
| per-pixel, indexed accumulators | 38 | 46–55 |
| per-pixel, named accumulators (shipped) | 40 | **37–47** |

Tests: a model-level test covers the stem shape, a batch of two, ci=7 at stride 1, and an odd unpadded size against the CPU; `cargo test -p tract-wgpu`, `cargo test -p test-wgpu` (4128 passed) green.

🍍


czoli1976:
@kali l'amour toujours

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

PR #2877, opened 2026-09-16
Jetson/4060/5090: allocate dynamic shared memory the way ThunderKittens does

czoli1976:
The CUDA kernels pick dynamic shared memory apart with pointer arithmetic on `extern __shared__` arrays (`char data_mmv[]`, `half smem[]`, `float data_soft_max_f32[]`). That is easy to get wrong once more buffers show up, and it does not force 16-byte alignment.

This copies ThunderKittens' `shared_allocator` (`include/common/util.cuh`): one `alignment_dummy` with `alignas(16)`, then typed `allocate<T, …>()` bumps. Used in `flash_attn.cu` (Q/K overlap + V) and `mm_mv.cu` (`buf_iw`). `nn.cu` only switches the softmax scratch to `alignment_dummy`; it still slices `buf_iw` / `vals` by hand.

`mm_mv_q.cu` is left on `extern __shared__ int data_mul_mat_q[]`. NVRTC miscompiled `al.allocate<int,…>` in that translation unit and zeroed `matmul_q40`. The three-commit history on this branch is that experiment and the revert.

ThunderKittens also runs a runtime `align_ptr()` on every allocate. Under NVRTC that cannot fold (`extern __shared__` is a runtime address) and cost 4–18% prefill on SM80/SM89, so it is not here. The dummy is 16-byte aligned and every allocate size in this crate is a multiple of 16.

No new ISA. Same binary path on Jetson, 4060, 5090, and H100.

| GPU | What changes |
|-----|----------------|
| any CUDA device | shared layout via the allocator; kernels otherwise unchanged |

```
cargo test -q -p tract-gpu -p tract-cuda -p test-cuda
```

🍍


czoli1976:
@kali can you ask a Robot to look at it?

czoli1976:
@kali can you share logs of https://github.com/sonos/tract/actions/runs/35079276598/job/104747258702?pr=2877 ?

czoli1976:
@kali Cuda CI Job ?  Rolled back to c++17

czoli1976:
let's try again, CUDA tests only if you can to get a quick feedback

czoli1976:
@kali

czoli1976:
Noise?

czoli1976:
will try to neuter these potential regressions

czoli1976:
@kali  une autre fois, especially to verify the extra potential overhead was shaved away

kali:
⚠️⚠️⚠️ Just rebased! ⚠️⚠️⚠️

czoli1976:
WDYT, @kali ? Keep? Rework ? Drop ?

czoli1976:
Also are these within noise floor or real ?

kali:
It seems pretty neutral performance-wise. Does it improve the code ? protects us from misalignement down the road ? unlocks some of the stuff that I have not looked at yet ?

kali:
also, i should probably add an llm to this bench...

kali:
ha ! LLMs are run, but not showing up, they're in the dead zone.

czoli1976:
Perf-neutral as expected, the code win is honestly small here but real in the TMA/WGMMA follow-up (mbarriers + multi-buffer tiles), alignment protection is base-only until size-rounding is added: 

`ptr += ((sizeof(at) + 15) & ~15) / sizeof(int);`

Unlocks the WGMMA follow-up  whcih have to admit is H100/H200+, while TMA should benefit 5090.

So honestly, not a biggie; I started doing this series under the wrong assumption (hallucination of the model) that more would benefit Jetson and 4090 and 5090.

So we can safely retire the entire series of just give it a try with TMA and the 5090 to see if it pays off, otherwise, unless you have ambitions to reach to Data Centre Variants in the foreseeable future.

As I can;t really test this stuff myself I kinda regret the entire series, proved to be more painful than I anticipated :-P

kali:
which PR is most likely to move the needle on my rtx 50900 ? let's give it a chance.

czoli1976:
> which PR is most likely to move the needle on my rtx 50900 ? let's give it a chance.

looking into it and see if I can pooint you to one or two for a quick try

czoli1976:
Honest answer: **#2885** is the only one that can move a number here, and only the Q4_0 **prefill** rows (Llama-3.2-1B/3B, 3.1-8B, OpenELM). Expect low single digits at p50/p99 prompt lengths: it hides the second Y tile's L2 latency behind `vec_dot`, the weight tile load stays synchronous. Decode is untouched (`mul_mat_vec_q`). While the LLM rows sit in the dead zone it will look flat.

**#2893** will not move this bench. TMA engages at head dim 64 only, so Llama-3.2-1B and OpenELM attention; 3B/8B and the ASR encoders are head dim 128 and never reach it. At ≤150 tokens of KV the kernel is latency-bound and TMA replaces 2 `cp.async` + swizzle math per thread per tile. Its value is proving TMA works under NVRTC on SM120, so a follow-up can stream the Q4_0 weight tiles with it — that is where the 5090 spends its time.

Both got rebased today (they carried duplicates of your commits) and both had bugs that reading found, so `/ci blackwell` on each comes first:
- #2893: `mbarrier.init` count was 0 (PTX wants ≥1) and nothing fenced the init; the TMA path was chosen by `__CUDA_ARCH__` alone, so the d=80/128 kernels ran TMA through the zeroed maps; the shared array is now 1024-byte aligned as `SWIZZLE_128B` requires; `sm_120a` is gone (plain sm_90+ PTX). `TRACT_CUDA_NO_TMA=1` compiles and launches without TMA — same binary A/B, and if the startup SIGSEGV from the last run persists it splits TMA from the rest of the PR.
- #2885: the predicated copy zero-filled past the Y tile into `tile_x`, and the host under-sized shared memory for the doubled tile. It had never run on a GPU.

No NVRTC on my side, so the kernel changes are read-verified only; the Rust side is checked.

#2877 is perf-neutral by construction. Retire it, or keep only `alignment_dummy` + the `buf_iw` bound — your call.

PR #2882, opened 2026-09-16
Lazy host storage

kali:


kali:
@JulienBalianSonos alternative option I am considering for B5 (and a bit of B6). stay-on-device is the default, API manifests the tensor lazily when their content is accessed only...

JulienBalianSonos:
I will  give this a try downstream and report back with how it behaves in practice.

The design seems cleaner from a tract API perspective and safer around facts: tensors remain ordinary tensors at the boundary, and the host/device decision is pushed into storage materialization rather than exposed as an explicit feedback API.
 
The main tradeoff I see is that readbacks become a bit subtler to reason about: any path that inspects bytes, serializes, logs, compares, or otherwise asks for host storage may trigger synchronization lazily. Arena views also do not seem representable directly as long-lived lazy host storage, so they still need either an owned device tensor or a copy at the boundary. Observability/debugging may also be a bit more involved, as the actual sync point becomes implicit and demand-driven.

That said, for the downstream use case where outputs are mostly fed back verbatim as later inputs, this still seems like the better direction.

kali:
@czoli1976 I'm going to use this one as a control sample

PR #2880, opened 2026-09-16
H100: Hopper WGMMA GEMM — parked until hardware is available

czoli1976:
Draft. Please skip. Not for the 5090 / 4060 / Jetson.

`wgmma.mma_async` is Hopper SM90. RTX 5090 is SM120 and `ptxas` rejects that instruction; 4060 / Jetson stay on cuBLAS already. This branch is #2877 plus a kernel those GPUs will never dispatch.

The live PRs for the machines we have are #2877 (allocator) and #2885 (Q4_0 `cp.async`). Reopen this after an H100 has checked the accumulator store.

🍍

PR #2884, opened 2026-09-16
H100: parked — TK L06 is TMA, not cp.async on WGMMA

czoli1976:
Draft. Please skip.

ThunderKittens educational level 06 is TMA + double-buffer + mbarrier on Hopper. This branch is `cp.async` on the old WGMMA kernel, and its text treated the 5090 as SM90 (it is SM120 and has no `wgmma.mma_async`).

Live WGMMA is #2880. A real L06 would stack TMA double-buffering on that, after an H100 has checked it.

🍍


czoli1976:
Converted to draft.

ThunderKittens educational **level 06 is TMA + double-buffering**, not `cp.async` on top of homemade WGMMA. The WGMMA PTX this sat on was the SS-descriptor-wrong, fence-after-MMA, SM90+ (RTX 5090) kernel; that path is rewritten in #2880 as Hopper-only.

A real L06 follow-up is TMA double-buffer + mbarrier on top of the new #2880 kernel, after an H100 has validated that GEMM.

PR #2885, opened 2026-09-16
Jetson/4060/5090: load the next Q4_0 Y tile while vec_dot runs

czoli1976:
`mul_mat_q_process_tile` loads a Y tile, `__syncthreads`, `vec_dot`, repeat. The second tile never starts until the first `vec_dot` is done.

On SM80+ (`__CUDA_ARCH__ >= 800`) this issues both Y tiles as `cp.async.ca` 16-byte copies into two shared buffers, as two groups, then:

- `wait_group<1>` — Y0 is done, Y1 may still be in flight
- `vec_dot(X, Y0)`
- `wait_group<0>` — Y1 is done (`wait_group<0>` is wait-all)
- `vec_dot(X, Y1)`

`cp.async.wait_group N` waits until at most N groups are still pending. Using `wait_group<0>` after Y0 would drain Y1 as well and collapse the overlap. Turing (SM75) keeps the original one-int-per-thread loads into a single Y buffer.

The host side doubles the Y-tile shared allocation when `props.major >= 8` so the two buffers fit.

Independent of #2877. This is the PR a 5090, 4060, or Jetson actually exercises.

| GPU | Path |
|-----|------|
| 5090 (SM120), 4060 (SM89), Jetson (SM87), H100 (SM90) | cp.async, two Y tiles |
| Turing SM75 | original direct loads |

```
cargo test -p tract-cuda mmm_ggml_prop_q4
cargo test -p test-cuda matmul_q40
```

🍍


kali:
not sure we have much signal here either :/

czoli1976:
LLM Rows?

czoli1976:
well, let's just close it @kali

czoli1976:
@kali sorry for wasting your time :P

kali:
it's alright :) it can't work all the time.

PR #2886, opened 2026-09-17
H100: Hopper flash TMA — parked (5090 TMA is a separate PR)

czoli1976:
Draft. Please skip. The TMA path here is compiled only for `__CUDA_ARCH__` in `[900, 1000)` and launched only when `major == 9` and head dim is 64.

A 5090 has TMA, but this PR does not enable it (NVRTC is `sm_120`, `use_tma` is false). 4060 / Jetson keep `cp.async.cg`. Passing flash tests on those GPUs only proves the fallback.

If we want TMA on the 5090, that is a new PR: compile the 2-D helpers for SM120 and turn `use_tma` on after the 128 B swizzle is checked against the existing MMA addressing. Do not mix that into this Hopper-only branch.

Conflicts with #2877 on `flash_attn.cu`.

🍍


czoli1976:
Drafted for the same reason. 5090 TMA would be a follow-up that compiles for SM120, not this Hopper-only path.

PR #2889, opened 2026-09-17
H100: parked — TMA WGMMA GEMM, wait for #2880

czoli1976:
Draft. Please skip.

Still the old WGMMA, plus `PR_UPDATE_SUMMARY.md` and `linalg/SDE_VM_SETUP.md`, which do not belong in a kernel PR. It also called the 5090 SM90.

After #2880 is correct on an H100, TMA double-buffering (TK level 06) can be a follow-up without those files.

🍍


czoli1976:
Converted to draft.

This PR still carries the old WGMMA, plus `PR_UPDATE_SUMMARY.md` and `linalg/SDE_VM_SETUP.md` which do not belong here. TMA-for-B also copied tiles without the transpose the original B load did.

Once #2880 is validated on an H100, a TMA double-buffer follow-up (TK level 06) can be stacked on it without those files.

PR #2890, opened 2026-09-17
H100: parked — not TK L07

czoli1976:
Draft. Please skip.

ThunderKittens educational level 07 is a producer warpgroup (TMA, fewer registers) and a consumer warpgroup (WGMMA, more registers) with `full` / `empty` semaphores. This branch is 256 threads where `tid==0` issues TMA and 127 load threads sit idle.

The 5090 is SM120, so this kernel would not run there either. Wait for #2880 on Hopper.

🍍


czoli1976:
Converted to draft.

ThunderKittens educational **level 07** is a producer warpgroup (TMA + `decrease_registers`) and a consumer warpgroup (WGMMA + `increase_registers`) with `full`/`empty` semaphores. 256 threads where `tid==0` issues TMA and 127 load threads sit idle is not that pattern.

Parked until #2880 is correct on Hopper; then L07 can be a real producer/consumer split.

PR #2893, opened 2026-09-17
5090: TMA for flash-attention K/V tiles

czoli1976:
**RTX 5090 (SM120) only.** Jetson (SM87) and the 4060 (SM89) have no TMA; they keep today's `cp.async.cg`. Hopper SM90 TMA is the drafted #2886, not this PR.

Flash attention already copies K and V into shared memory with predicated `cp.async.cg` and a software swizzle. The 5090 can issue that copy as one bulk-tensor load with an mbarrier (`cp.async.bulk.tensor.2d`). The MMA is still `mma.sync` — this is not WGMMA.

Compiled only for `__CUDA_ARCH__` in `[1200, 1300)`. Launched only when `props.major == 12` and the head dim is 64 (128-byte rows, so `CU_TENSOR_MAP_SWIZZLE_128B` can match the existing MMA swizzle). Other head dims, and every other GPU, stay on `cp.async.cg`. Partial KV tails always use predicated `cp.async.cg` (TMA box size is a full `BLOCK_KV` tile).

Tensor maps are 2-D, `boxDim` in elements, `globalDim[0]` is the fastest axis, descriptors passed by value as `__grid_constant__`.

`TRACT_CUDA_NO_TMA=1` compiles the flash library without TMA and launches on `cp.async.cg`, so both paths can be compared on one build.

Independent of #2877 / #2885 / #2880. Will conflict with #2877 on `flash_attn.cu` if both land.

| GPU | Path | Exercises TMA? |
|-----|------|----------------|
| RTX 5090, head dim 64 | TMA | **yes — this PR** |
| RTX 5090, other head dims | `cp.async.cg` | no |
| 4060 / Jetson | `cp.async.cg` | no (no TMA hardware) |
| H100 | `cp.async.cg` here | no (see #2886) |

```
cargo test -p tract-cuda test_nernst_fattn
```
On a 4060 or Jetson this suite only proves the fallback. Needs a 5090 with head dim 64 to hit TMA.

🍍


czoli1976:
The `/ci blackwell` failure was not a kernel failure. The job checked out this PR (branched from an older `main`) and ran `.github/blackwell-guard.sh`, which only exists after `d019238d2`. Cargo test never started.

Rebased onto current `main` and pushed:
- `cuTensorMapEncodeTiled` now fails the launch if encode returns non-success (was a silent zero map)
- `test_nernst_fattn` has d=64 cases with `seq_len=64` and `kv_len` a multiple of 32, so `fullq_` can actually take TMA on SM120. Two of the old d=64 cases had `kv_len=1` and never hit TMA.

Needs another `/ci blackwell` on `0e3f09fd0`.

kali:
@czoli1976 i'm on it. let me handle this :)

czoli1976:
> @czoli1976 i'm on it. let me handle this :)

Roger that

kali:
https://github.com/sonos/tract/actions/runs/35323426438/job/105530928007 

I think we have a couple of failures, but the workflow seems to work now.

No promise on when the jobs will run, this machine is one of my workhorse, so I'll have to manually schedule time for one-shot CI jobs when I can spare it.

kali:
@czoli1976 pr is yours now. tell me if there is another one we should look at before this one.

czoli1976:
The live log (relu / add / reshape / pow / … all FAILED, ~1800/6801) is not a TMA-flash numerical bug. FlashAttn was first in the eager NVRTC list and the TMA PTX lived in `common.cuh`, which is injected into every library. On `sm_120` a ptxas reject of `cp.async.bulk.tensor` fails context init and takes the whole CUDA suite down with it.

Pushed `bcf34a30e`: TMA helpers only in `flash_attn.cu`, FlashAttn compiles last and is skipped if NVRTC fails, FlashAttn targets `sm_120a` on SM120. Other libraries stay on portable `sm_120`.

Needs another `/ci blackwell` when the box is free. If this was the cause, ONNX/unit should go green and only flash/SDPA d=64 would remain to judge TMA. If everything is still red, SM120 CUDA on this box has a pre-existing problem unrelated to this PR.

czoli1976:
@kali for the order and which box, not everything needs the 5090.

**Look at these, in this order**

1. **#2877** (allocator) — any CUDA GPU. Jetson or 4060 is enough; the 5090 is not required. Check Q4_0 prefill vs `main` is ~flat (an earlier `align_ptr` in this series was +4–18% on those boxes). Merge this first: #2893 conflicts with it on `flash_attn.cu`.

2. **#2885** (Q4_0 `cp.async` Y tiles) — SM80+, so Jetson / 4060 / 5090 all run the new path. Jetson or 4060 is enough. This is the only open PR that can move llama Q4_0 numbers on the machines we have.

3. **#2893** (this one, flash TMA) — **5090 only**. Jetson (SM87) and the 4060 (SM89) have no TMA; they keep `cp.async.cg`. A green flash test on those two only proves the fallback. Rebase onto #2877 after that lands.

**Do not look at** the H100 drafts (#2880 WGMMA, #2886 Hopper TMA, #2884 / #2889 / #2890). `wgmma.mma_async` is SM90; the 5090 is SM120 and `ptxas` rejects it. Those wait for an H100.

| PR | Jetson | 4060 | 5090 |
|----|--------|------|------|
| #2877 | yes | yes | yes (not required) |
| #2885 | yes | yes | yes (not required) |
| #2893 | fallback only | fallback only | **this box** |

kali:
cancelleed, restarted...

czoli1976:
The isolation worked: ONNX/unit ran, FlashAttn NVRTC failed on a real compile error (not TMA ISA):

```
error: expression must have a constant value
    (uint32_t)max(BLOCK_Q, 3 * BLOCK_KV) * PADDED_DIM * sizeof(half);
note: cannot call non-constexpr function "max(int, int)"
```

Pushed a fix (`k_tensor_elems` as a ternary of template ints, unused mbar smem addrs removed). Remaining FAILED tests were all `unit::sdpa::*`, which is FlashAttn skipped.

Needs another `/ci blackwell` when the box is free.

czoli1976:
> cancelleed, restarted...

some progress .....

czoli1976:
@kali [run 35331407896](https://github.com/sonos/tract/actions/runs/35331407896) finished on `bcf34a30e` (isolation, before the constexpr fix):

`4860 passed; 24 failed; 1917 ignored` in ~21 min.

The 24 are all `unit::sdpa::*` (three harnesses). FlashAttn cubin was skipped at context init because NVRTC rejected `max(BLOCK_Q, 3 * BLOCK_KV)` in a constexpr. Everything else on the 5090 is green.

The constexpr fix is already on the PR as `805bfdf7a`. That run cannot pick it up. Another `/ci blackwell` when the box is free, please — that SHA is the one that can actually compile FlashAttn / TMA.

PR #2897, opened 2026-09-18
tensor: slicing a device-backed tensor went through the host, so an a…

kali:
…pplication keeping recurrent state on the device paid a download and an upload to truncate it. Give TensorStorage a slice hook and let DeviceTensor answer it with a device-to-device copy through the copy_nd kernels the backends already carry, reached from the new Tensor::slice_in_storage and api Tensor::sliced. The result owns its buffer, so it neither aliases the arena it may have been read from nor pins it, and any axis works: the copy is strided on the source where a view would have had to be dense.


kali:
@JulienBalianSonos This is an alternative idea to B6 by-view slicing. It implements a device-to-device slices instead, so it's not as "free" as the views from a speed perspective. The point is, the blast radius is significantly less important than slicing views: no buffer aliasing (with complicated ownership and deferred release of resource), no assumpution on the consumer ability to use a buffer with non-trivial strides and/or start offset. The cost is, of course, this makes a copy.

Reasoning says its the right thing on cuda, because in a split memory context the d-to-d path will be maybe 100x faster than a d-to-h, copy, h-to-d. It's murkier on metal.

Can you give it a shot still ? See if it gets close to the performance you were able to reach with the views on your use case ?

kali:
@JulienBalianSonos Mmm, i forgot the key LazyStorage on that one...

kali:
Here we go. According to claude: "On Metal this buys a memcpy's worth of bandwidth (GPU copy vs single-core host copy) — order 0.4 ms/step at a 32 MB state, nothing at small ones, since MetalTensor::to_host is an Arc::clone and there was never a real readback to remove. CUDA is where it's worth two orders of magnitude."

JulienBalianSonos:
That’s exactly the implem I’m worried about: even if the device-to-device copy is fast, it could make memory usage explode for long-context LLMs, making it difficult to compete with ggml.
Still, we’ll see: I’ll give it a try on Monday.

JulienBalianSonos:
ok I tested the original downstream and one with this variant slice. I understand the motivation for returning independently owned storage: it avoids aliasing, arena-lifetime concerns, ...

But, for a downstream LLM application on memory-constrained hardware, I will not use this copying path (esp. on long-context KV truncation). Removing even a handful of tokens copies almost the entire retained cache across the affected layers. This adds allocation and bandwidth costs, and cached snapshots can keep the original buffers alive alongside the
copies.

This isn't a concern about correctness, nor the hit of 2 to 3 sec on big model it would take.

I see value in this as a copying fallback, but it wouldn't replace the non-copying path I need downstream.

Longer term, a paged or persistent cache with shared immutable prefixes and independently owned continuations might fit this workload better. That would be a separate design discussion, though.

kali:
Ok, thanks for giving it a chance, and yeah, no big surprise, we were expecting this at least on metal (on cuda, copies are one order of magnitude faster dtod, so it may have been enough).

I think I will still merge this, it feels harmless enough and might be useful even if it does help you.

And I'll keep exploring options for either some workable form of aliasing, or a more adhoc solution dedicated to kv-cache operations.

PR #2901, opened 2026-09-21
(perf) wgpu: fold the elementwise ops after a transposed convolution into it

czoli1976:
A transposed convolution's bias and activation ran as a separate pass over its output. On the selfie segmenter that is the sigmoid over the whole 144x256 mask — one of the 81 launches, the last one before readback, and a full-size elementwise pass.

`WgpuDeconv` now carries an epilogue the way `WgpuConv` does: the kernel takes extra operands after the weights (the same `Params` tail, chain layout and `epilogue_body` as the convolutions), and `fuse_deconv_epilogue` absorbs the chain that follows through the same `absorb` the GEMM and convolution rules use.

On the segmenter (Chrome, same-page replay) the deconv + bias + sigmoid pair goes from two launches at 56 µs to one at 26 µs.

Tests: a model-level test asserts the fold fires on a bias + sigmoid and that the result matches the CPU; `cargo test -p tract-wgpu`, `cargo test -p test-wgpu` (4128 passed), clippy clean.

Rebased on current main: the deconv rule passes its existing epilogue to `absorb` the way the convolution rule does now, so a transposed convolution's epilogue can also grow past its bias; main's rewritten deconv kernel took the epilogue unchanged.

🍍


czoli1976:
I understand  some of the the stuff in latest and upcoming wgpu PRs may need to move to the GPU backend

kali:
ok. i'm not opposed to prototyping / evaluating on one single backend, we're already doing it everywhere. only issue is keeping track of opportunities. at least looking at operators and transforms list may help mechanize it down the road.

kali (approved):

PR #2909, opened 2026-09-21
(perf) core/cnn: accumulate generic NCHW deconvolution along width

Hosi121:
The generic NCHW deconvolution sum jumps between channel planes for each input position. This change accumulates input rows for each kernel tap, so reads are contiguous and writes stay within one output row.

The path uses safe slices for contiguous NCHW/CHW buffers with input width >= 16. It preserves the addition order for each output, adds no heap buffer, and leaves the 2x2 stride-2 path from #2831 unchanged. F16, F32, and F64 use the same loop structure. Small rows and other layouts use the existing loop.

On a Core Ultra 7 255H, WSL2, Rust 1.97.1, release build, one thread:

| Workload | Base | Change | Time reduction |
|---|---:|---:|---:|
| DCGAN generator, batch 1 | 3.304 ms | 2.549 ms | 22.9% |
| DCGAN generator, batch 4 | 14.591 ms | 11.752 ms | 19.5% |
| ConvTranspose, CI=4, CO=32, 64x64, K=3x3, stride=2 | 1.438 ms | 1.076 ms | 25.2% |
| ConvTranspose, CI=4, CO=32, 64x65, K=3x3, stride=1, dilation=2 | 0.983 ms | 0.664 ms | 32.5% |

The generator test includes all five layers, activation functions, and buffer allocation. It uses fixed synthetic weights, not a trained checkpoint. Six alternating process pairs give 21.7-23.2% reduction for batch 1 and 17.4-21.3% for batch 4. Output hashes match. The measured base is `98df95c8`; the CPU crates used here have no changes up to the PR base, `92a30df4`.

[Source, commands, all samples, controls, and check logs](https://github.com/Hosi121/tract/tree/701e1b0f06111f721290ede477706dc89fbb12d4) are available. Small fallback cases with width 8/9 measured 1.4%/1.0% slower (0.034/0.027 microseconds). Unchanged paths also vary, so I do not claim a gain for every shape. ARM performance is not measured.

Validation:

- 99 deconvolution tests pass with `PROPTEST_CASES=512`.
- 5,760 independent reference cases match bit for bit across types, layouts, groups, strides, dilation, and padding.
- Added suite cases cover widths 15/16/17, rectangular kernels, CHW/NCHW, groups, and strides 1/2/3.
- Formatting and diff checks pass. The CI command `cargo clippy --all-targets` completes with existing warnings outside the changed code. Full workspace lint stops at an Apple framework dependency on Linux.


kali (approved):

Hosi121:
Thanks for the benchmark results. I used the PR's x86 CLI to check the optimized CPU graphs. The speaker_id and Nemotron preprocessor graphs have no `DeconvSum` nodes. TRUNet has six `DeconvSum` nodes, but their input width is 1. The new path needs a width of at least 16.

The nightly reference is `617126f2`. This PR has parent `92a30df4`, so the comparison includes 17 other commits. These include a change to `PulsedSameAxisConcat`, which both models with speed alerts use. I have not identified the cause of the alerts.

Could you compare `92a30df4` and `bb833d97` on the same Jetson and Orange Pi, with the same build settings? That would isolate this PR. I cannot verify those device results on my x86 machine.

kali:
Don't worry about these red benches, they're not yours.

kali:
Thanks you for the contribution !

PR #2911, opened 2026-09-22
transformers: add routed mixture-of-experts FFN operator

JulienBalianSonos:
## Motivation

Routed mixture-of-experts feed-forward blocks select a small set of experts for each token and combine their outputs using router weights. `tract_moe_ffn` represents that computation as a single operator, with CPU execution and an NNEF primitive for model exports.

## Operator

- Router projection and top-k expert selection, followed by weighted accumulation of expert outputs.
- Four gating modes: softmax over selected experts, softmax over all experts followed by selection, sigmoid, and raw router scores.
- Per-expert subplans for feed-forward computation, including gated activations, optional router and expert biases, and the clamped SwiGLU variant used by GPT-OSS.
- Canonical and linear-layer expert weight layouts, with floating-point and block-quantized weights.
- A block-quantized routed-matmul path for grouped expert computation, alongside the subplan implementation.
- NNEF import and export of `MoeFfn`, preserving optional inputs, gating, expert layout, activation parameters, and Q40 weight storage.

The operator returns the combined hidden states; router logits are internal to expert selection rather than a separate output.

NNEF export operates on the portable graph before CPU codegen. Packed `OptMoeFfn` execution plans are not serializable and report an explicit error; the runtime does not retain an additional copy of the expert weights for export.

## Scope

This is the baseline operator implementation, not the performance-optimization part of the broader MoE proposal. Its scope is operator semantics, CPU execution, and NNEF integration; it is not presented as a fully tuned MoE inference implementation. It includes the block-quantized routed-matmul path described above, but further performance work is separate.

It does not introduce Metal/CUDA lowering, GPU memory-management changes, or application-level inference behavior. Performance optimizations, including GPU acceleration, will follow in a subsequent PR built on this operator.

Related broader proposal: https://github.com/sonos/tract/pull/2084.


kali:
Let's try to move this forward. It's big, but pretty self-contained, low risk to other areas of tract.

Claude has some notes:

```
. OptMoeFfn equality lies — moe_ffn.rs:2373-2395: Hash/PartialEq cover only k/gate/num_experts/d_model/d_hidden/q40_is_some and deliberately skip expert_plans, router_plan, wg_bias. PushSplitDown runs inside the codegen optimizer (core/src/optim/mod.rs:132), after the pass that creates OptMoeFfn, and merges two successors of one outlet when Node::same_as (same inputs + dyn_eq). OptMoeFfn's only input is x, so two MoE blocks reading the same outlet with matching shapes collapse into one and silently use one block's weights. I haven't built a repro — a plain decoder stack never shares that outlet — but the equality contract is wrong regardless.
2. MoeFfn::output_facts (:1655) validates nothing: no input count, no ranks, no k <= E, no D agreement between x/wg/w1. RouteTopK and RoutedMatMul both check. A malformed export types fine and dies in eval.
3. Top-k tie-break (:427, :603, :1542, :2511, :2779): sort_unstable_by(total_cmp) descending — arbitrary order on equal logits, where torch.topk takes the lowest index. Rare, and exactly the kind of single-token divergence nobody reproduces. select_nth_unstable would also drop the per-token E log E.
```

And I would add one of mine: for once, can we add a module-level comment that summarize which variants of MOE patterns are supported and which models it unlocks ? Rationale is, these operators are not derived from something which has a documenting surface (like ONNX operators and NNEF ones) while being pretty complex.

I think it calls for:
* documentation of the nnef primitives it introduces (with or w/o how a couple of relevant models are combining them maybe ?)
* inventory of ops (are they "decluttered" ? "optimized" ?).

Also, try to split the huge file, maybe ?

I will also push back on the unsafe bits. "Why they are actually sound" is not good enough. I need to know why we can't express it with safe code, what they buy us.

JulienBalianSonos:
Ready for another look. The operator semantics and CPU execution are separated, and ClampedSwiGlu now lives in core with standalone NNEF round-trip coverage. GPU optimizations remain for the follow-up PR.

transformers/src/ops/moe_ffn/activation.rs
```diff
@@ -0,0 +1,12 @@
+use crate::ops::{gelu_approximate::gelu_approximate, silu::silu};
+use tract_nnef::internal::*;
+
+pub(super) fn activation_op(name: &str, has_w3: bool) -> Option<Box<dyn TypedOp>> {
+    match name {
+        "silu" => Some(Box::new(silu())),
+        "swiglu" if has_w3 => Some(Box::new(silu())),
```
kali: is this a typo ?

transformers/src/ops/moe_ffn/cpu.rs
```diff
@@ -0,0 +1,951 @@
...
+    build_block_quant_routed_matmul, pack_prepared_routed_matmul_rhs, run_prepared_routed_matmul,
+    run_prepared_routed_matmul_accumulate_one, run_prepared_routed_matmul_many,
+    run_prepared_routed_matmul_many_same_rhs,
+};
+use std::{
+    fmt,
+    hash::{Hash, Hasher},
+    sync::Arc,
+    time::{Duration, Instant},
+};
```
kali: This style of input is supposed to be banned from tract. I'm a bit surprised it's not managed upstram or in the CI. I'll figure it out.

transformers/src/ops/moe_ffn/cpu.rs
```diff
@@ -0,0 +1,951 @@
...
+/// Only the router keeps a long-lived state. Expert plans are spawned per eval
+/// so they can run on the rayon pool; they are stateless matmuls, so there is
+/// nothing to carry between calls anyway.
+struct OptMoeFfnState {
+    op: OptMoeFfn,
+    router_state: TypedSimpleState,
+    experts: CpuExpertState,
+}
+
+impl fmt::Debug for OptMoeFfnState {
```
kali: is this different from the auto derive ?

transformers/src/ops/moe_ffn/activation.rs
JulienBalianSonos: thats wished because SwiGLU is SiLU followed by multiplication with a second projection:

```
  SwiGLU(x) = SiLU(x · W1) ⊙ (x · W3)
```
That helper selects only the activation for the first projection. The surrounding expert code computes `x · W3` and performs the multiplication.

  So:

  - silu() supplies the activation.
  - has_w3 ensures the second projection exists.
  - Together, they implement SwiGLU. (decomposition that allow to avoid duplicate of silu optims)

transformers/src/ops/moe_ffn/cpu.rs
```diff
@@ -0,0 +1,951 @@
...
+
+#[derive(Clone)]
+/// Only the router keeps a long-lived state. Expert plans are spawned per eval
+/// so they can run on the rayon pool; they are stateless matmuls, so there is
+/// nothing to carry between calls anyway.
+struct OptMoeFfnState {
+    op: OptMoeFfn,
+    router_state: TypedSimpleState,
+    experts: CpuExpertState,
+}
```
kali: Is this really stateful ? What the stateful bit ? Unless i missed something, everything is scoped with the Runnable, and I suspect this is what costs us the unsafe Send and Sync down the type tree.

core/src/ops/nn/clamped_swiglu.rs
```diff
@@ -0,0 +1,80 @@
...
+        ensure!(inputs[0].shape() == inputs[1].shape(), "ClampedSwiGlu gate/up shape mismatch");
+        ensure!(
+            inputs.iter().all(|t| matches!(t.datum_type(), DatumType::F16 | DatumType::F32)),
+            "ClampedSwiGlu inputs must be f16 or f32"
+        );
+        let mut output = Tensor::zero_dt(f32::datum_type(), inputs[0].shape())?;
+        if output.len() == 0 {
+            return Ok(tvec![output.into_tvalue()]);
+        }
+        let gate = inputs[0].cast_to::<f32>()?.into_owned();
```
kali: i think the args_2!(...) would avoid a copy here.

transformers/src/ops/moe_ffn/lowering.rs
```diff
@@ -0,0 +1,339 @@
...
+use super::{
+    ExpertLayout, MoeFfn, OptMoeFfn, RouteTopK, RoutedCombine, RoutedInputMode, RoutedMatMul,
+    block_quant_group_tensor,
+};
+use std::sync::Arc;
+use tract_nnef::internal::*;
+use tract_nnef::tract_core::ops::{konst::Const, math::mul};
+use tract_nnef::tract_core::tract_linalg::block_quant::BlockQuantStorage;
+
+impl TypedOp for MoeFfn {
```
kali: i'm not sure i would have split this :) per-op felt more natural. in the other filen, we have Q40 in the middle of the "regular" mode, should we split away that bit instead ?

transformers/src/ops/moe_ffn/routing.rs
```diff
@@ -0,0 +1,756 @@
...
+    }
+    op_as_typed_op!();
+}
+
+impl EvalOp for RouteTopK {
+    op_out_of_plan!();
+
+    fn eval(&self, _ctx: &EvalContext, inputs: TVec<TValue>) -> TractResult<TVec<TValue>> {
+        let x_t = inputs[0].cast_to::<f32>()?.into_owned();
+        let wg_t = inputs[1].cast_to::<f32>()?.into_owned();
```
kali: args_2(...) to skip a copy

transformers/src/ops/moe_ffn/routing.rs
```diff
@@ -0,0 +1,756 @@
...
+        ensure!(inputs[3].rank() == 1, "RoutedMatMul route_expert_ids must be rank 1");
+        let route_count = inputs[2].shape.to_tvec()[0].clone();
+        let out_dim = inputs[1].shape.to_tvec()[2].clone();
+        Ok(tvec![f32::datum_type().fact(&[route_count, out_dim])])
+    }
+
+    as_op!();
+}
+
+impl RoutedQ40MatMul {
```
kali: let's try to keep implementation together

transformers/src/ops/moe_ffn/tests.rs
```diff
@@ -0,0 +1,1069 @@
...
+    let result = plan.spawn()?.run(tvec![x_data.clone().into_tvalue()])?;
+    let ref_result = SimplePlan::new(ref_model)?.spawn()?.run(tvec![x_data.into_tvalue()])?;
+    result[0].close_enough(&ref_result[0], Approximation::Approximate)?;
+
+    Ok(())
+}
+
+#[test]
+fn test_q40_linear_expert_layout_matches_dequantized_reference() -> TractResult<()> {
+    let mut model = TypedModel::default();
```
kali: I'm not sure about test that implement a model inline with no variations. This could be a harness/nnef test exampl, right ? We are just describing a model in a more verbose sentence.

Or else, we go all the way and make these proptest in test-rt (and then we can test all runtimes).

transformers/src/ops/routed_matmul.rs
```diff
@@ -0,0 +1,626 @@
...
+                }
+            }
+            _ => RoutedRowOffsets::Explicit(offsets),
+        }
+    }
+}
+
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub(super) struct RoutedInputRows {
+    pub(super) base: *const f32,
```
kali: Can we use a shallow copy of the input + offset here ? avoid the pointer altogether?

transformers/src/ops/moe_ffn/cpu.rs
kali: ok, it's still not part of rustfmt (semi controversial since it does more than editing whitespace). The fix is in rust-analyser config, but that's not something I can enforce it for contributors. Putting it AGENTS.md, Maybe it will help. 

```imports.granularity.group = "module"```

transformers/src/ops/moe_ffn/cpu.rs
JulienBalianSonos: The retained state is reusable router/Q40 execution workspace, not model history. Evaluation now borrows expert plans from the operator. The custom unsafe Send/Sync implementations are gone.

transformers/src/ops/routed_matmul.rs
JulienBalianSonos: Routed operands now retain an Arc<Tensor> plus checked element offsets. Cloning shares the input storage; packing uses safe slice reads instead of raw source pointers.

